PR #14, opened 2019-11-30
Toolchain upgrade

brunoborges:
# Description

This PR includes the following changes:

 - Maven Module project

Now the project is a multi-module Maven project, with `examples` and `client` as two different projects. The parent has dependency management (versions).

 - Client project generates gRPC client and Protobuf classes

Added Maven plugins to generate these classes during build. May want to consider moving these classes out of src/main/java and open space for more user friendly Java Client API.

 - Moved to Java 11

Originally the code was set for Java 7. The ecosystem has moved on to 8 at large, and modern applications are being built with Java 11, the latest LTS release. Considering Dapr is a new project, it should be up to speed with modern Java.

 - Updated versions of dependencies and Maven plugins

 - Added a rewrite of the helloworld Node.js example written in Java (needs testing)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


brunoborges:
// @brendandburns fyi

client/src/main/java/io/dapr/DaprProtos.java
```diff
@@ -20,12 +20,10 @@ public static void registerAllExtensions(
 
     /**
      * <code>.google.protobuf.Any data = 1;</code>
-     * @return Whether the data field is set.
```
artursouza: Why did the auto-generated code remove the @return documentation?

pom.xml
```diff
@@ -1,14 +1,59 @@
...
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <grpc.version>1.25.0</grpc.version>
+    <protobuf.version>3.11.0</protobuf.version>
+    <protoc.version>3.10.0</protoc.version>
+    <os-maven-plugin.version>1.6.2</os-maven-plugin.version>
+    <maven-dependency-plugin.version>3.1.1</maven-dependency-plugin.version>
+    <maven-antrun-plugin.version>1.8</maven-antrun-plugin.version>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
```
artursouza: I understand that Java 11 is the latest. On the other hand, this is a SDK to be consumded by existing applications as well. If we move to Java 11, would it break compatibility with previous version? If so, then we should move down to Java 10, 9 or maybe even 8 to make sure we have a decent coverage for existing apps. I searched and found one report from 2018 and this one for 2019: https://www.jetbrains.com/lp/devecosystem-2019/java/

It seems that Java 8 is still the most popular SE by a wide margin.

pom.xml
```diff
@@ -1,14 +1,59 @@
...
 
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <grpc.version>1.25.0</grpc.version>
+    <protobuf.version>3.11.0</protobuf.version>
+    <protoc.version>3.10.0</protoc.version>
+    <os-maven-plugin.version>1.6.2</os-maven-plugin.version>
+    <maven-dependency-plugin.version>3.1.1</maven-dependency-plugin.version>
+    <maven-antrun-plugin.version>1.8</maven-antrun-plugin.version>
+    <maven.compiler.source>11</maven.compiler.source>
```
artursouza: Can we have code in 11 but target SE 8, for example?

client/src/main/java/io/dapr/DaprProtos.java
brunoborges: No clue. Need to dig into the Java generator. May be able to look late this week.

pom.xml
brunoborges: DAPR is a brand new project, and applications written for it should be targeting Java 11. Developers can reuse existing Java libraries and frameworks written for older versions of Java, for sure, but I see no point in having this SDK supporting Java 8.

artursouza: I agree that new Java projects should be in Java 11, mainly because it is the latest LTS version. My question here is, what should be the minimum version to allow existing Java apps to use Dapr? - not 100%, of course. It seems that prior to 11, the previous LTS version for Java is 8. I would like to hear from customers that want to use Dapr if Java 11 is OK for them now. We should eventually migrate to Java 11, my concern is if doing it now will hurt adoption.

https://www.oracle.com/technetwork/java/java-se-support-roadmap.html

brunoborges: We may need to schedule a call. Hit me on brborges @ ms for later this week.

client/pom.xml
```diff
@@ -0,0 +1,130 @@
+<project
```
artursouza: I see that you have split it into two projects -> two jars. We are currently discussing the structure of the project and how packaging will be. You probably want to be part of that discussion too. We will be forking this repo soon for a major work on the Java-SDK.

brunoborges: Happy to connect with the team, @artursouza ! Where are discussions being held?

artursouza: We just moved to a public issue: https://github.com/dapr/java-sdk/issues/16

pom.xml
artursouza: I created an issue to discuss this with the broader community: https://github.com/dapr/java-sdk/issues/17

artursouza:
I merged your repo into mine and I am making the changes to JDK8 plus doing a small change in the packaging structure: still separating samples from the sdk but we will have a single sdk jar for now. Later one, we might split into client, actors, pubsub, etc.

artursouza:
I took this change, made some changes and created this PR instead: https://github.com/dapr/java-sdk/pull/18

Please, review if you have the time.

Issue #28, opened 2019-12-09
[java-sdk] Dapr HTTP Client for Actors


artursouza:
Create a Dapr client to make http calls directly (without using auto-gen client).
Implement all APIs needed to support Actors.

This overlaps with #22. While issue #22 will create a complete version of the Dapr client for Java, this one will unblock Actors work first, making HTTP calls directly.

PR #39, opened 2019-12-11
Basic Dapr HTTP Client for Actors.

artursouza:
# Description

Adding Dapr HTTP Client Async for actors.

## Issue reference

Please reference the issue this PR will close: #28 

## Checklist

* [X ] Code compiles correctly
* [ ] Created/updated tests


sdk/src/main/java/io/dapr/actors/DaprAsyncClient.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Saves Actor state in Dapr. This is temporary until the Dapr runtime implements the Batch state update.
+     * @param actorType Type of actor.
+     * @param actorId Actor Identifier.
+     * @param keyName State name.
+     * @param data State to be saved.
+     * @return Asynchronous void result.
+     */
+    Mono<Void> saveState(String actorType, String actorId, String keyName, String data);
```
amanbha: We can remove this saveState, it will not be used as Actors will always save transactionally, Thsi will be rmeoved from dotnet-sdk as well.

sdk/src/main/java/io/dapr/actors/DaprHttpAsyncClient.java
```diff
@@ -0,0 +1,186 @@
...
+        Request request = new Request.Builder()
+                .url(new URL(this.baseUrl + urlString))
+                .method(method, body)
+                .addHeader(Constants.HEADER_DAPR_REQUEST_ID, requestId)
+                .build();
+
+        Response response = this.httpClient.newCall(request).execute();
+        if (!response.isSuccessful())
+        {
+            throw new IOException(String.format("Dapr's Actor API %s failed with return code: %d.", urlString, response.code()));
```
amanbha: In the failure case, Dapr returns a json with error code and an error messgae, the code can try to parse and throw a more specific DaprEcception wimilar to https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Actors/DaprHttpInteractor.cs#L446

sdk/src/main/java/io/dapr/actors/DaprHttpAsyncClient.java
```diff
@@ -0,0 +1,186 @@
...
+    public Mono<Void> removeState(String actorType, String actorId, String keyName) {
+        String url = String.format(Constants.ACTOR_STATE_KEY_RELATIVE_URL_FORMAT, actorType, actorId, keyName);
+        return invokeAPIVoid("DELETE", url, null);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Mono<Void> saveStateTransactionally(String actorType, String actorId, String data) {
```
amanbha: We can remove this saveState, it will not be used as Actors will always save transactionally,

sdk/src/main/java/io/dapr/actors/Constants.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors;
+
+public final class Constants {
```
amanbha: This doesn't need to be public. Only classes used by user code should be made public.

sdk/src/main/java/io/dapr/actors/DaprAsyncClient.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package io.dapr.actors;
+
+import reactor.core.publisher.Mono;
+
+/**
+ * Interface for interacting with Dapr runtime.
+ */
+public interface DaprAsyncClient {
```
amanbha: Doesn't need to be public, this will used by other public Apis in SDK. eg. ActorProxy on client side & ActorRuntime classes

sdk/src/main/java/io/dapr/actors/DaprClientBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package io.dapr.actors;
+
+import okhttp3.OkHttpClient;
+
+/**
+ * Builds an instance of DaprAsyncClient or DaprClient.
+ */
+public class DaprClientBuilder {
```
amanbha: Doesn't need to be public, this will used by other public Apis in SDK. eg. ActorProxy on client side & ActorRuntime class.

sdk/src/main/java/io/dapr/actors/DaprAsyncClient.java
artursouza: Done.

sdk/src/main/java/io/dapr/actors/DaprHttpAsyncClient.java
artursouza: I did parse the error code. I am doing a simpler error handling for now.

We will enhance once more use cases are implemented. I found the error handling in the .Net SDK a little bit complicated. Plus, the parsing from String to Enum seems to be overkill since the SDK will probably never keep up with the changes on the server side.

sdk/src/main/java/io/dapr/actors/DaprHttpAsyncClient.java
artursouza: Done.

sdk/src/main/java/io/dapr/actors/Constants.java
artursouza: For now, made it visible to package. If it needs to be accessed by another Java package (not jar) it will probably need to be made public again.

sdk/src/main/java/io/dapr/actors/DaprAsyncClient.java
artursouza: For now, made it visible to package. If it needs to be accessed by another Java package (not jar) it will probably need to be made public again.

sdk/src/main/java/io/dapr/actors/DaprClientBuilder.java
artursouza: For now, made it visible to package. If it needs to be accessed by another Java package (not jar) it will probably need to be made public again.

sdk/pom.xml
```diff
@@ -91,11 +106,11 @@
               <outputTargets>
                 <outputTarget>
                   <type>java</type>
-                  <outputDirectory>${project.build.sourceDirectory}</outputDirectory>
+                  <outputDirectory>${protobuf.output.directory}</outputDirectory>
                 </outputTarget>
                 <outputTarget>
                   <type>grpc-java</type>
-                  <outputDirectory>${project.build.sourceDirectory}</outputDirectory>
+                  <outputDirectory>${protobuf.output.directory}</outputDirectory>
```
artursouza: I am moving auto-generated code to the auto-generated folder and out of source control. This way, we only need to version the proto files.

sdk/src/main/java/io/dapr/actors/DaprHttpAsyncClient.java
amanbha: Sure, ErrorCode can remain as string instead of enum. but in general SDK should keep up with the runtime

sdk/src/main/java/io/dapr/actors/DaprHttpAsyncClient.java
```diff
@@ -0,0 +1,205 @@
...
+        String requestId = UUID.randomUUID().toString();
+        RequestBody body = json != null ? RequestBody.create(MEDIA_TYPE_APPLICATION_JSON, json) : REQUEST_BODY_EMPTY_JSON;
+
+        Request request = new Request.Builder()
+                .url(new URL(this.baseUrl + urlString))
+                .method(method, body)
+                .addHeader(Constants.HEADER_DAPR_REQUEST_ID, requestId)
+                .build();
+
+        Response response = this.httpClient.newCall(request).execute();
```
LMWF: Is this a blocking call?

artursouza: Yes. I have not found a way to make it async from the http client itself. We might end up using RestClient from Microsoft's jar. I would love to fix it but prefer to keep moving on this until we find the solution here.

LMWF: That's fine, we can revisit this line later.

amanbha: Could you please add a TODO comment here that this needs to be async before release?

artursouza: Adding in my next PR.

Issue #29, opened 2019-12-09
[java-sdk] Java Actor runtime


artursouza:
Creating the Java Actor runtime classes, similar to the one in the Dotnet sdk. 

Implementing the API endpoints for the SDK. The business logic should be (preferably) identical to the one in the Dotnet SDK. 

Following are some of the few classes in C# SDK; 

ActorRuntime: This is the actor runtime in SDK and contains information about registered types. It contains a mapping of ActorType and ActorManager. Actor types are registered with ActorRuntime. 

ActorManager: It manages actors instances for an actor type, activates, deactivates them and make method calls to them. 

ActorStateManager: Abstracts state management for Actors, provides a cache for Actor states, so that states which have already been worked on are not fetched again from Dapr for an active actor. 

Actor: Base class for Actors. (In Java it can be called DaprActor) 

DaprStateProvide: ActorStateManager uses it to interact with Dapr for state management. 

RouterBuilderExtensions: This is for registering routes with aspnetcore, it registers the required routes (urls) on which Dapr runtime will call into user service code.

artursouza:
Unit tests to close on this issue: https://github.com/dapr/java-sdk/issues/85 https://github.com/dapr/java-sdk/issues/86

PR #41, opened 2019-12-12
ActorTypeInformation with parsing logic + some stubs.

artursouza:
# Description

ActorTypeInformation with parsing logic + some stubs.
Added TODO to fix async call in http client.

## Issue reference

Please reference the issue this PR will close: #29 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/runtime/AbstractActor.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/AbstractActor.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors.runtime;
+
+/**
+ * TODO
+ */
+public abstract class AbstractActor {
+}
--- a/sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
@@ -1,0 +1,181 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2019 Swen Schisler <swen.schisler@fourtytwosoft.io>.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.dapr.actors.runtime;
+
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Contains the information about the class implementing an actor.
+ *
+ * @author Swen Schisler <swen.schisler@fourtytwosoft.io>
+ * @author Artur Souza <artur.barbalho@microsoft.com>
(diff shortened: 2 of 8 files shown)


sdk/src/test/java/io/dapr/actors/runtime/ActorTypeInformationTest.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        ActorTypeInformation info = ActorTypeInformation.parse(A.class);
+        Assert.assertNotNull(info);
+        Assert.assertEquals("A", info.getName());
+        Assert.assertEquals(A.class, info.getImplementationClass());
+        Assert.assertFalse(info.isAbstractClass());
+        Assert.assertFalse(info.isRemindable());
+        Assert.assertEquals(1, info.getInterfaces().size());
+        Assert.assertTrue(info.getInterfaces().contains(MyActor.class));
```
artursouza: @amanbha In this example, should it return only MyActor (like it is now) or also Actor. Is the expected result is a shallow list of interfaces, then it is correct. Otherwise, we should do a deep search. This is the same behavior I identified in the .Net SDK as well.

sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    /**
+     * Creates an {@link #ActorTypeInformation} from actorType.
+     *
+     * @param actorClass The class implementing the actor to create ActorTypeInformation for.
+     * @return {@link #ActorTypeInformation} created from actorType.
+     */
+    public static ActorTypeInformation parse(Class actorClass) {
```
LMWF: Agree this is better than "Get()".  This looks like a factory.  What do you think about construct/build/create instead?

sdk/src/main/java/io/dapr/actors/runtime/Remindable.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors.runtime;
+
+/**
+ * TODO
+ */
+public interface Remindable {
```
amanbha: It should have a RecieveReminder method, is it TODO?

sdk/src/main/java/io/dapr/actors/runtime/AbstractActor.java
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors.runtime;
+
+/**
+ * TODO
+ */
+public abstract class AbstractActor {
```
amanbha: Is this the Actor base class? Name doesn't seem obvious to me.

sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
artursouza: I have considered the 4 options on the table. Let me know which one you feel stronger about.

Why parse?
It can throw an exception, so we have a tryParse as well that can return null. This is why parse came to me naturally: we see parse() and tryParse() everywhere. Also, the logic to construct this object is very heavy, so parse represents that effort IMO.

Why not construct?
I am not used this the "construct" term in Java code. It is just my experience.

Why not build?
It gives the idea that it will never fail and always return a valid result.

Why maybe create?
I have seen the term create before for things like this. I am fine with this option as well.

sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
```diff
@@ -0,0 +1,181 @@
+/*
+ * The MIT License
```
amanbha: Update copyright

sdk/src/main/java/io/dapr/actors/runtime/AbstractActor.java
artursouza: I renamed it to the Java way, looking at the SpringBoot code as example.

sdk/src/main/java/io/dapr/actors/runtime/Remindable.java
artursouza: TODO because we are not implementing this yet. It is just to make the code compile and run for the code we cover in this PR. More to come.

sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
artursouza: Renamed to create()

sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
artursouza: I followed the copyright from the author of the file. Is it OK to change it as per the license terms?

sdk/src/main/java/io/dapr/actors/runtime/AbstractActor.java
artursouza: This is mainly because IActor becomes Actor interface in Java, so name conflicts.

sdk/src/main/java/io/dapr/actors/runtime/ActorTypeInformation.java
amanbha: It should be our license like in any other file,

artursouza: This file is significantly different, updating license.

PR #44, opened 2019-12-12
Add ActorRuntime

LMWF:
# Description

Adds the ActorRuntime class.  ActorManager is not implemented yet, so the calls to it are commented out.  There are two TODOs inside that should be addressed later:
- should it be singleton
- the logging mechanism

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/src/main/java/io/dapr/actors/DaprAsyncClient.java
```diff
@@ -10,7 +10,7 @@
 /**
  * Interface for interacting with Dapr runtime.
  */
-interface DaprAsyncClient {
+public interface DaprAsyncClient {
```
amanbha: Let's not make these public, it's made public because of adding a runtime package,  if needed put classes in same package OR better if this can be split into calls from proxy to Dapr and from runtime to Dapr. Then they can stay in their individual packages.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,164 @@
...
+    private ActorRuntime() throws IllegalStateException{
+        if (instance != null) {
+            throw new IllegalStateException("ActorRuntime should only be constructed once");
+        }
+
+        this.actorManagers = new HashMap<String, ActorManager>();
+        daprAsyncClient = new DaprClientBuilder().buildAsyncClient();
+    }
+
+    /** TODO: maybe this should not be a singleton
```
amanbha: There will only be one ActorRuntime in user service.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public Set<String> getRegisteredActorTypes() {
+        return this.actorManagers.keySet();
+    }
+
+    /**
+     * Registers an actor with the runtime.
+     * @param clazz The type of actor.
+     */
+    void RegisterActor(Class clazz) {
```
amanbha: Can it be RegisterActor<T>?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private ActorManager GetActorManager(String actorTypeName) throws IllegalStateException
+    {
+        ActorManager actorManager = this.actorManagers.get(actorTypeName);
+
+        if (actorManager == null)
+        {
+            String errorMsg = String.format("Actor type %s is not registered with Actor runtime.", actorTypeName);
+
+            // TODO - figure out logging.  For now just print it straight out
```
amanbha: For logging, add an interface and use it, printing to console can be one implement ation of it.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
LMWF: Removed comment.

sdk/src/main/java/io/dapr/actors/DaprAsyncClient.java
LMWF: Split into 2 clients: app -> dapr and proxy -> dapr -> app.  The former is in the runtime dir and the other is in the dir above it.

sdk/src/main/java/io/dapr/actors/ActorProxyToAppAsyncClient.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.dapr.actors;
+
+import reactor.core.publisher.Mono;
+
+/**
+ * Interface to invoke actor methods.
+ */
+interface ActorProxyToAppAsyncClient {
```
amanbha: This can move to io.dapr.actors.client package. Fine for now, it can be moved when ActorProxy class is added.

sdk/src/main/java/io/dapr/actors/ActorProxyToAppHttpAsyncClient.java
```diff
@@ -0,0 +1,40 @@
...
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.UUID;
+
+/**
+ * Http client to call actors methods.
+ */
+class ActorProxyToAppHttpAsyncClient extends DaprClientBase implements ActorProxyToAppAsyncClient {
```
amanbha: This can move to io.dapr.actors.client package. Fine for now, it can be moved when ActorProxy is added.

sdk/src/main/java/io/dapr/actors/ActorProxyToAppAsyncClient.java
artursouza: ActorProxyAsyncClient

sdk/src/main/java/io/dapr/actors/ActorProxyToAppHttpAsyncClient.java
artursouza: ActorProxyHttpAsyncClient

sdk/src/main/java/io/dapr/actors/ActorProxyToAppHttpAsyncClient.java
```diff
@@ -0,0 +1,40 @@
...
+        super(port, httpClient);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Mono<String> invokeActorMethod(String actorType, String actorId, String methodName, String jsonPayload) {
+        String url = String.format(Constants.ACTOR_METHOD_RELATIVE_URL_FORMAT, actorType, actorId, methodName);
+        return invokeAPI("PUT", url, jsonPayload);
```
artursouza: super.invokeAPI (style).

sdk/src/main/java/io/dapr/actors/Constants.java
```diff
@@ -8,7 +8,7 @@
 /**
  * Useful constants for the Dapr's Actor SDK.
  */
-final class Constants {
+public final class Constants {
```
artursouza: I wonder if we should get rid of this class for Java and keep the constants in the used files, unless we have shared ones. @amanbha  ?

sdk/src/main/java/io/dapr/actors/DaprClientBuilder.java
```diff
@@ -21,10 +22,10 @@
      * Builds an async client.
```
amanbha: Name of client doenst match the Return type of BuildAsyncClient method, the method is returning ActorProxyToAppHttpAsyncClient This class can be package private also as it will be used by ActorProxyonly, (both will be in io.dapr.actors.clientpackage)

sdk/src/main/java/io/dapr/actors/DaprClientBase.java
```diff
@@ -1,206 +1,128 @@
...
+
+import okhttp3.*;
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.UUID;
+
+// base class of hierarchy
+public class DaprClientBase {
```
artursouza: AbstractDaprAsyncClient

I looked at the SpringBoot code to decide on naming for these things: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/AbstractCommand.java

BTW, can you also put in the fix from Swen?

sdk/src/main/java/io/dapr/actors/DaprClientBase.java
```diff
@@ -1,206 +1,128 @@
...
+        String requestId = UUID.randomUUID().toString();
+        RequestBody body = json != null ? RequestBody.create(MEDIA_TYPE_APPLICATION_JSON, json) : REQUEST_BODY_EMPTY_JSON;
+
+        Request request = new Request.Builder()
+                .url(new URL(this.baseUrl + urlString))
+                .method(method, body)
+                .addHeader(Constants.HEADER_DAPR_REQUEST_ID, requestId)
+                .build();
+
+        // TODO: make this call async as well.
```
artursouza: We have the code snippet for this, can you put it in?

sdk/src/main/java/io/dapr/actors/DaprClientBase.java
```diff
@@ -1,206 +1,128 @@
...
+     * Shared object representing an empty request body in JSON.
+     */
+    private static final RequestBody REQUEST_BODY_EMPTY_JSON = RequestBody.create(MEDIA_TYPE_APPLICATION_JSON, "");
+
+    /**
+     * JSON Object Mapper.
+     */
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private final String baseUrl;
```
artursouza: comment, I forgot probably.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Set;
+import java.util.function.Function;
+
+import io.dapr.actors.*;
+
+/**
+ * Contains methods to register actor types. Registering the types allows the runtime to create instances of the actor.
+ */
+public class ActorRuntime {
+    private static volatile ActorRuntime instance;
```
artursouza: comments everywhere.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+    private static AppToDaprAsyncClient appToDaprHttpAsyncClient;
+    private final String TraceType = "ActorRuntime";
```
artursouza: static

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -57,7 +57,7 @@ public static ActorRuntime getInstance() {
      * Registers an actor with the runtime.
      * @param clazz The type of actor.
      */
-    void RegisterActor(Class clazz) {
+    <TActor extends AbstractActor> void RegisterActor(Class<TActor> clazz) {
```
amanbha: this method would be public

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        return  instance;
+    }
+
+    /**
+     *
+     * @return Actor type names registered with the runtime.
+     */
+    public Set<String> getRegisteredActorTypes() {
```
artursouza: Return a readonly collection. It is the lowest common denominator we need to adhere to. Set is too specific, the caller just needs to know that it is a collection and no order should be assumed.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+        return this.actorManagers.keySet();
+    }
+
+    /**
+     * Registers an actor with the runtime.
+     * @param clazz The type of actor.
+     */
+    <TActor extends AbstractActor> void RegisterActor(Class<TActor> clazz) {
```
artursouza: I usually just see people using T, unless you have more than one, single letter is enough.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+        RegisterActor(clazz, null);
+    }
+
+    /**
+     * Registers an actor with the runtime.
+     * @param clazz The type of actor.
+     * @param actorServiceFactory An optional delegate to create actor service. This can be used for dependency injection into actors.
+     */
+    <TActor extends AbstractActor> void RegisterActor(Class<TActor> clazz, Function<ActorTypeInformation, ActorService> actorServiceFactory)
```
artursouza: T

sdk/src/main/java/io/dapr/actors/runtime/DaprClientBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+
+package io.dapr.actors.runtime;
+
+import okhttp3.OkHttpClient;
+import io.dapr.actors.*;
+
+/**
+ * Builds an instance of DaprAsyncClient or DaprClient.
+ */
+public class DaprClientBuilder {
```
amanbha: The class can be package private as its used only by actorRuntime. 
Update the class name, there are 2 classes with name DaprClientBuilder and it was confusing.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+        {
+            actorService = actorServiceFactory.apply(actorTypeInfo);
+        }
+        else
+        {
+            actorService = new ActorService(actorTypeInfo);
+        }
+
+        // Create ActorManagers, override existing entry if registered again.
+        this.actorManagers.put(actorTypeInfo.getName(), new ActorManager(actorService));
```
artursouza: What if there is a collision?

Maybe we should synchronize this method too and to it atomically to avoid race condition.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+
+        this.actorManagers = new HashMap<String, ActorManager>();
+        appToDaprHttpAsyncClient = new DaprClientBuilder().buildAsyncClient();
+    }
+
+    /**
+     * Returns an ActorRuntime object.
+     * @return An ActorRuntime object.
+     */
+    public static ActorRuntime getInstance() {
```
artursouza: Good standard way to do singleton. Since this is public, it makes sense. Let's try to minimize calls to this method in our code and get ActorRuntime via dependency injection as much as possible. LGTM here.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -0,0 +1,163 @@
...
+     */
+    static void FireTimer(String actorTypeName, String actorId, String timerName)
+    {
+        // uncomment when ActorManager implemented
+        // return instance.GetActorManager(actorTypeName).FireTimerAsync(new ActorId(actorId), timerName);
+    }
+
+    private ActorManager GetActorManager(String actorTypeName) throws IllegalStateException
+    {
+        ActorManager actorManager = this.actorManagers.get(actorTypeName);
```
artursouza: We probably need to synchronize this, not?

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprAsyncClient.java
```diff
@@ -1,89 +1,80 @@
...
+    Mono<String> getState(String actorType, String actorId, String keyName);
+
+    /**
+     * Removes Actor state in Dapr. This is temporary until the Dapr runtime implements the Batch state update.
+     * @param actorType Type of actor.
+     * @param actorId Actor Identifier.
+     * @param keyName State name.
+     * @return Asynchronous void result.
+     */
+    Mono<Void> removeState(String actorType, String actorId, String keyName);
```
amanbha: This method can be removed as aCtors will always updates states using the saveStateTransactionally method.

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprAsyncClient.java
```diff
@@ -1,89 +1,80 @@
...
+
+package io.dapr.actors.runtime;
+
+import reactor.core.publisher.Mono;
+
+
+/**
+ * Interface for interacting from the actor app to the Dapr runtime.
+ */
+interface AppToDaprAsyncClient {
```
artursouza: Some alternatives:
ActorRuntimeDaprAsyncClient
ActorDaprAsyncClient
RuntimeDaprAsyncClient
DaprAsyncClient (since it is already in the actors.runtime package)

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprHttpAsyncClient.java
```diff
@@ -0,0 +1,92 @@
...
+import io.dapr.actors.DaprClientBase;
+import okhttp3.*;
+import reactor.core.publisher.Mono;
+
+
+/**
+ * Http client to call Dapr's API for actors.
+ */
+//public class DaprHttpAsyncClient implements DaprAsyncClient {
+class AppToDaprHttpAsyncClient extends DaprClientBase implements AppToDaprAsyncClient {
```
artursouza: Same as above, just with the Http in the name you choose.

sdk/src/main/java/io/dapr/actors/runtime/DaprClientBuilder.java
artursouza: Why public since this is only for the runtime?

sdk/src/main/java/io/dapr/actors/runtime/DaprClientBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+    public DaprClientBuilder withPort(int port) {
+        this.port = port;
+        return this;
+    }
+
+    /**
+     * Tries to get a valid port from environment variable or returns default.
+     * @return Port defined in env variable or default.
+     */
+    private static int GetEnvPortOrDefault() {
```
artursouza: A lot of duplicate code in the builders. Can we have an Abstract class here too?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
amanbha: Insertion into dictionary and fetching from this method will not happen concurrenctly. The dictionary will be populated at startup before Dapr can make calls into it.

sdk/src/main/java/io/dapr/actors/DaprClientBase.java
LMWF: Renamed.  But which fix are you referring to?

sdk/src/main/java/io/dapr/actors/runtime/DaprClientBuilder.java
LMWF: Removed public.

sdk/src/main/java/io/dapr/actors/DaprClientBuilder.java
LMWF: Make package private and moved.  I think the name of `buildAsyncClient()` method is ok though.  The caller will have an instance of this class to call this and that will show the type.

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprHttpAsyncClient.java
LMWF: Unless you feel strongly I'll keep the current name.  It can always be renamed.

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprAsyncClient.java
LMWF: Unless you feel strongly I'll keep the current name.  It can always be renamed.

sdk/src/main/java/io/dapr/actors/Constants.java
artursouza: ?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: Comments?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: ?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: ?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: ?

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: From my understanding, there is a Dapr endpoint upstream that will be invoking this down the stream. How can we guarantee there will be only one http call being processed?

For example, multiple clients might be activating the same or different actors, then these calls will all come here concurrently unless you lock at the level above.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
LMWF: Missed this.

sdk/src/main/java/io/dapr/actors/DaprClientBase.java
LMWF: I think that can go separately.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
amanbha: Multiple threads reading from it don't need synchronization for this dictionary. Each of the call will get their actor manager and work with it.
Calls for the same actor id cannot come here concurrently, actor runtime in Dapr guarantees that.
Items in this dictionary are added at startup only.

sdk/src/main/java/io/dapr/actors/Constants.java
amanbha: Yes, they can b moved to respective files.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: I still not sure about this not getting concurrent calls. I will not block this PR. I will revisit this code once we have the SDK working end-to-end.

PR #47, opened 2019-12-14
Add ActorMethodInfoMap

LMWF:
# Description

Adds ActorMethodInfoMap + unit test.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/runtime/ActorMethodInfoMap.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/ActorMethodInfoMap.java
@@ -1,0 +1,32 @@
+package io.dapr.actors.runtime;
+
+import java.lang.reflect.Method;
+import java.util.HashMap;
+
+class ActorMethodInfoMap {
+    private final HashMap<String, Method> methods;
+
+    public <T> ActorMethodInfoMap(Iterable<Class<T>> interfaceTypes)
+    {
+        this.methods = new HashMap<String, Method>();
+
+        // Find methods which are defined in Actor interface.
+        for (Class<T> actorInterface : interfaceTypes)
+        {
+            for (Method methodInfo : actorInterface.getMethods())
+            {
+                this.methods.put(methodInfo.getName(), methodInfo);
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/dapr/actors/runtime/ActorMethodInfoMap.java
```diff
@@ -0,0 +1,32 @@
+package io.dapr.actors.runtime;
+
+import java.lang.reflect.Method;
+import java.util.HashMap;
+
+class ActorMethodInfoMap {
```
artursouza: comments, please.

sdk/src/test/java/io/dapr/actors/runtime/ActorMethodInfoMapTest.java
```diff
@@ -0,0 +1,54 @@
...
+        try {
+            Method m1 = m.LookupActorMethodInfo("getData");
+            Assert.assertEquals("getData", m1.getName());
+            Class c = m1.getReturnType();
+            Assert.assertEquals(c.getClass(), String.class.getClass());
+            Parameter[] p = m1.getParameters();
+            Assert.assertEquals(p[0].getType().getClass(), String.class.getClass());
+        }
+        catch(Exception e) {
+            Assert.fail("Exception not expected.");
```
artursouza: No need to try catch it. Let the exception propagate, the test will fail as well with less code.

sdk/src/main/java/io/dapr/actors/runtime/ActorMethodInfoMap.java
LMWF: Done

sdk/src/test/java/io/dapr/actors/runtime/ActorMethodInfoMapTest.java
LMWF: Unfortunately that doesn't compile.  The only alternative I can think of is to mark `normalUsage()` itself with a `throws`, but that looks weird too.  I'm going to leave it, but if you know of a better way we can change it.

Issue #19, opened 2019-12-06
[java-sdk] Adds checkstyle and bug finder


artursouza:
* Adds a checkstyle to Java SDK and make build fail due to style violations.
* Adds [SpotBugs](https://spotbugs.github.io/) and make build fail due to violations.
* Fixes existing violations.
* Generated code should be excluded from static analysis.

yaron2:
Can we close this?

LMWF:
Checkstyle is done, static analysis is not complete (spot bugs).

PR #48, opened 2019-12-15
Improved async call to okhttp. 

ji11er:
# Description

Former solution just wrapped a blocking call, now we using asynch call and collect the results via callbacks. Mono's are created based on the result. This make more efficient use of the resources since nothing is blocking now. 

Reformatted according new checkstyle

## Issue reference
#19 
#46 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/AbstractClientBuilder.java
+++ b/sdk/src/main/java/io/dapr/actors/AbstractClientBuilder.java
@@ -2,48 +2,46 @@
  * Copyright (c) Microsoft Corporation.
  * Licensed under the MIT License.
  */
-
 package io.dapr.actors;
-
-import okhttp3.OkHttpClient;
-import io.dapr.actors.*;
 
 /**
  * Base class for client builders
  */
 public abstract class AbstractClientBuilder {
 
-    /**
-     * Default port for Dapr after checking environment variable.
-     */
-    private int port = AbstractClientBuilder.GetEnvPortOrDefault();
+  /**
+   * Default port for Dapr after checking environment variable.
+   */
+  private int port = AbstractClientBuilder.GetEnvPortOrDefault();
 
-    /**
-     * Overrides the port.
-     * @param port New port.
-     * @return This instance.
-     */
-    public AbstractClientBuilder withPort(int port) {
-        this.port = port;
-        return this;
+  /**
+   * Overrides the port.
+   *
+   * @param port New port.
+   * @return This instance.
+   */
+  public AbstractClientBuilder withPort(int port) {
+    this.port = port;
+    return this;
+  }
+
+  /**
+   * Tries to get a valid port from environment variable or returns default.
+   *
+   * @return Port defined in env variable or default.
+   */
+  protected static int GetEnvPortOrDefault() {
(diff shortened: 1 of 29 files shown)


artursouza:
The Java SDK is WIP, so we are not merging into master yet. I changed the base of this PR to the right branch. One we are done with this milestone, we should be able to use master directly again.

sdk/src/main/java/io/dapr/actors/AbstractClientBuilder.java
```diff
@@ -2,48 +2,46 @@
...
-     * @param port New port.
-     * @return This instance.
-     */
-    public AbstractClientBuilder withPort(int port) {
-        this.port = port;
-        return this;
+  /**
+   * Default port for Dapr after checking environment variable.
+   */
+  private int port = AbstractClientBuilder.GetEnvPortOrDefault();
```
artursouza: Any particular reason to move from 4 spaces to 2 spaces? I am good either way but sticking to 4 spaces would reduce the diff :)

sdk/src/main/java/io/dapr/actors/AbstractDaprClient.java
```diff
@@ -1,119 +1,173 @@
...
+  }
+
+  public interface DaprHttpCallback {
 
     /**
-     * Invokes an API asynchronously that returns a text payload.
-     * @param method HTTP method.
-     * @param urlString url as String.
-     * @param json JSON payload or null.
-     * @return Asynchronous text
```
artursouza: Why is the return comment removed?

sdk/src/main/java/io/dapr/actors/AbstractDaprClient.java
```diff
@@ -1,119 +1,173 @@
...
+    String requestId = UUID.randomUUID().toString();
+    RequestBody body = json != null ? RequestBody.create(MEDIA_TYPE_APPLICATION_JSON, json) : REQUEST_BODY_EMPTY_JSON;
+
+    Request request = new Request.Builder()
+        .url(new URL(this.baseUrl + urlString))
+        .method(method, body)
+        .addHeader(Constants.HEADER_DAPR_REQUEST_ID, requestId)
+        .build();
+
+    this.httpClient.newCall(request).enqueue(new Callback() {
```
artursouza: Thanks for that.

sdk/src/main/java/io/dapr/actors/AbstractClientBuilder.java
ji11er: I'd like to preserve horizontal space. If you can agree, I'd prefer. Unfortunately, I had two commits one with small changes and one more with the formatting only. But I created a patch to follow the rules for pull requests and patched my forked copy of the repo. Therefore the commit is huge.

sdk/src/main/java/io/dapr/actors/AbstractDaprClient.java
ji11er: Never mind

ji11er:
I promise less headache with upcoming commits.

ji11er:
I hit that button accidentially before the review was finished

PR #51, opened 2019-12-17
Add ActorTimer and related

LMWF:
# Description

Adds the ActorTimer class and related.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/runtime/ActorTimerImpl.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/ActorTimerImpl.java
@@ -1,0 +1,53 @@
+package io.dapr.actors.runtime;
+
+import org.json.JSONObject;
+
+import java.time.Duration;
+import java.util.function.Function;
+
+class ActorTimerImpl implements ActorTimer {
+
+  private final AbstractActor owner;
+  private String name;
+  private Function<Object, Void> asyncCallback;
+  private Object state;
+  private Duration dueTime;
+  private Duration period;
+
+  public ActorTimerImpl(AbstractActor owner, String timerName, Function<Object, Void> asyncCallback, Object state, Duration dueTime, Duration period) {
+    this.owner = owner;
+    this.name = timerName;
+    this.asyncCallback = asyncCallback;
+    this.state = state;
+    this.dueTime = dueTime;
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/io/dapr/actors/runtime/ActorTimerImpl.java
```diff
@@ -0,0 +1,53 @@
+package io.dapr.actors.runtime;
+
+import org.json.JSONObject;
+
+import java.time.Duration;
+import java.util.function.Function;
+
+class ActorTimerImpl implements ActorTimer {
+
+  private final AbstractActor owner;
```
artursouza: Javadoc comments.

sdk/src/main/java/io/dapr/actors/runtime/ActorTimerImpl.java
```diff
@@ -0,0 +1,53 @@
...
+  public Duration getPeriod() {
+    return this.period;
+  }
+
+  public Object getState() {
+    return this.state;
+  }
+
+  String serialize()
+  {
```
artursouza: Use ObjectMapper instead. Ideally this should not be in this layer. This should be just a Java Bean.

LMWF: Is there a way to use ObjectMapper to write out specific values without annotating the class and giving it the whole object?  The .net implementation of this method, for example, specifically writes out individual fields instead of using some library that may need to reflect over the whole class to determine what to write out.

amanbha: This is doing custom serialization of the properties, either you will have to create extra objects or plug in custom serialization. Writing the json directly is the fastest way to serialize when you know the types of properties.

artursouza: We discussed this over the phone. This is in the code path for IO, so reflection will not be a big different. I agree to this approach mainly because it avoids new classes. I end up creating new classes on my end to handle this. Later on, we should see if it is possible to remove the serialization logic out from this class.

PR #53, opened 2019-12-18
ActorService + DaprStateProvider + some more.

artursouza:
# Description

ActorService + DaprStateProvider + some more.

## Issue reference

Please reference the issue this PR will close: #29 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
@@ -8,7 +8,6 @@ import io.dapr.actors.*;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.function.Function;
 
 /**
  * Contains methods to register actor types. Registering the types allows the
@@ -88,18 +87,16 @@ public class ActorRuntime {
    * Registers an actor with the runtime.
    *
    * @param clazz The type of actor.
-   * @param actorServiceFactory An optional delegate to create actor service.
+   * @param actorFactory An optional factory to create actors.
    * This can be used for dependency injection into actors.
    */
-  public <T extends AbstractActor> void RegisterActor(Class<T> clazz, Function<ActorTypeInformation, ActorService> actorServiceFactory) {
+  public <T extends AbstractActor> void RegisterActor(Class<T> clazz, ActorFactory actorFactory) {
     ActorTypeInformation actorTypeInfo = ActorTypeInformation.create(clazz);
 
-    ActorService actorService;
-    if (actorServiceFactory != null) {
-      actorService = actorServiceFactory.apply(actorTypeInfo);
-    } else {
-      actorService = new ActorService(actorTypeInfo);
-    }
+    ActorFactory actualActorFactory = actorFactory != null ? actorFactory : new DefaultActorFactory<T>(actorTypeInfo);
+      // TODO: Refactor into a Builder class.
+      DaprStateAsyncProvider stateProvider = new DaprStateAsyncProvider(this.appToDaprAsyncClient, new ActorStateProviderSerializer());
+      ActorService actorService = new ActorServiceImpl(actorTypeInfo, stateProvider, actualActorFactory);
 
     // Create ActorManagers, override existing entry if registered again.
     synchronized (this.actorManagers) {
--- a/sdk/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors.runtime;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * State Provider to interact with Dapr runtime to handle state.
+ */
+class DaprStateAsyncProvider {
+
+  /**
+   * Shared Json serializer/deserializer as per Jackson's documentation, used only for this class.
+   */
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  private final AppToDaprAsyncClient daprAsyncClient;
+
+  private final ActorStateProviderSerializer serializer;
+
+  DaprStateAsyncProvider(AppToDaprAsyncClient daprAsyncClient, ActorStateProviderSerializer serializer) {
+    this.daprAsyncClient = daprAsyncClient;
+    this.serializer = serializer;
+  }
+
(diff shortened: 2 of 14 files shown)


sdk/src/main/java/io/dapr/actors/runtime/DefaultActorFactory.java
```diff
@@ -0,0 +1,49 @@
...
+  public T createActor(ActorService actorService, ActorId actorId) {
+    try {
+      if (this.actorTypeInformation == null) {
+        return null;
+      }
+
+      Constructor<T> constructor = this.actorTypeInformation.getImplementationClass().getConstructor(ActorService.class, ActorId.class);
+      return (T)constructor.newInstance(actorService, actorId);
+    } catch (ReflectiveOperationException e) {
+      e.printStackTrace();
```
LMWF: It's probably a good idea to log to a common class so all of it happens in one place.  ActorTrace is a stub for this right now.

sdk/src/main/java/io/dapr/actors/runtime/ActorStatefulService.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package io.dapr.actors.runtime;
+
+import io.dapr.actors.ActorId;
+
+/**
+ * Implementation of the Actor Service that contains a state provider.
+ */
+class ActorStatefulService implements ActorService {
```
LMWF: I don't know if we want to name it this.  "stateful" is a bit of a strong word, and the existence of the state provider doesn't mean they actually have to store state.  What about ActorServiceImpl?

sdk/src/main/java/io/dapr/actors/runtime/ActorStatefulService.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Customizable factor for Actors.
```
LMWF: "factory"

sdk/src/test/java/io/dapr/actors/runtime/DefaultActorFactoryTest.java
```diff
@@ -0,0 +1,85 @@
...
+    ActorId actorId;
+
+    public MyActor(ActorService actorService, ActorId actorId) {
+      this.actorService = actorService;
+      this.actorId = actorId;
+    }
+  }
+
+  /**
+   * A non-compliant implementation of Actor to be used in the tests below.
```
LMWF: "non-compliant" is usually written as a single word.

sdk/src/main/java/io/dapr/actors/runtime/ActorStatefulService.java
LMWF: This class should be visible to the user - it should be public.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -88,18 +87,16 @@ public static ActorRuntime getInstance() {
    * Registers an actor with the runtime.
    *
    * @param clazz The type of actor.
-   * @param actorServiceFactory An optional delegate to create actor service.
+   * @param actorFactory An optional factory to create actors.
    * This can be used for dependency injection into actors.
    */
-  public <T extends AbstractActor> void RegisterActor(Class<T> clazz, Function<ActorTypeInformation, ActorService> actorServiceFactory) {
+  public <T extends AbstractActor> void RegisterActor(Class<T> clazz, ActorFactory actorFactory) {
```
amanbha: Why can't it be just a lambda?

sdk/src/main/java/io/dapr/actors/runtime/ActorStatefulService.java
amanbha: Agree in name change.

sdk/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
```diff
@@ -0,0 +1,128 @@
...
+
+    return result.map(s -> {
+      if (s == null) {
+        return (T)null;
+      }
+
+      try {
+        return this.serializer.deserialize(s, clazz);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
amanbha: Throw more meaningful custom DaprException with error code.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
amanbha: Can it take an ActorService factory or builder which takes actor factory.
It would allow users to override settings at ActorService layer. It's fine for now, but keep that extensibility open for changes.

sdk/src/main/java/io/dapr/actors/runtime/DefaultActorFactory.java
artursouza: I agree. I would like to get the log and tracing part done as a separate work and covering all the codebase.

sdk/src/main/java/io/dapr/actors/runtime/ActorStatefulService.java
artursouza: Will do. I had it names ActorServiceImpl but I am not sure about the Impl convention yet.

sdk/src/main/java/io/dapr/actors/runtime/DefaultActorFactory.java
LMWF: Ok.

sdk/src/main/java/io/dapr/actors/runtime/ActorStatefulService.java
artursouza: The idea is to minimize exposure to customers. It is possible to have the same flexibility via a configuration class.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: We discussed to take a configuration class instead.

sdk/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
artursouza: We need to refactor all exception cases. It will be handled as a separate change.

sdk/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: https://stackoverflow.com/questions/18400210/java-8-where-is-trifunction-and-kin-in-java-util-function-or-what-is-the-alt

artursouza: Adding @FunctionalInterface like in here:
https://github.com/spring-projects/spring-boot/blob/bf294a762a1235b942e8d6316032d131f5e94e2b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/ClusterFactory.java

sdk/src/test/java/io/dapr/actors/runtime/DefaultActorFactoryTest.java
artursouza: As per the dictionary in IntelliJ, "non-compliant" is correct and "noncompliant" is wrong. My browser also complains about "noncompliant".

PR #54, opened 2019-12-19
Add ReminderInfo

LMWF:
# Description

Adds ReminderInfo + unit test

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/runtime/ReminderInfo.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/ReminderInfo.java
@@ -63,10 +63,8 @@ class ReminderInfo
   static ReminderInfo deserialize(byte[] stream) throws IOException {
     try {
       ObjectMapper om = new ObjectMapper();
-
-      final ObjectMapper mapper = new ObjectMapper();
-      final MapType type = mapper.getTypeFactory().constructMapType(Map.class, String.class, Object.class);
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/dapr/actors/runtime/ReminderInfo.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      return om.writeValueAsString(objectNode);
+    } catch (IOException e) {
+      throw e;
+    }
+  }
+
+  static ReminderInfo deserialize(byte[] stream) throws IOException {
+    try {
```
amanbha: Can we change this desrialization to be more performant by directly iterating over json and also not creating these extra objects like Map etc.
https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#compatible-not-so-easy-use-the-streaming-api

It's more code but more performant, do this both for serialization and deserialization.

artursouza: The serialize and deserialize should get using the same type, not String vs byte. Let me fix this since Leon is OOO. Keep the PR open so I remember.

artursouza: I actually need this class now, so I am merging this.

PR #67, opened 2019-12-31
Add UnitTest to ActorProxyImpl and JaCoCo (coverage tool) implementation

JuanJose-Herrera:
… from an actor method is not wrapped in "Data" object

Add Unit test for the ActorProxy class

# Description

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -122,24 +122,24 @@
             </goals>
             <configuration>
               <rules>
-                <!--<rule>
+                <rule>
                   <element>CLASS</element>
                   <limits>
                    <limit>
                       <counter>LINE</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>0.80</minimum>
+                      <minimum>0.00</minimum>
                     </limit>
                     <limit>
                       <counter>BRANCH</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>0.80</minimum>
+                      <minimum>0.00</minimum>
                     </limit>
                   </limits>
                   <excludes>
                     <exclude>io.dapr.utils.Constants</exclude>
                   </excludes>
-                </rule>-->
+                </rule>
               </rules>
             </configuration>
           </execution>
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -53,7 +53,9 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
-
+  <properties>
+    <skipITs>true</skipITs>
+  </properties>
   <build>
     <plugins>
       <plugin>
@@ -88,6 +90,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
         <version>2.22.2</version>
+        <configuration>
+          <skipITs>${skipITs}</skipITs>
+        </configuration>
         <executions>
           <execution>
             <goals>
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -97,6 +97,55 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <version>0.8.4</version>
+        <executions>
+          <execution>
+            <id>default-prepare-agent</id>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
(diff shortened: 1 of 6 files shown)


sdk/pom.xml
```diff
@@ -97,6 +97,55 @@
...
+            </goals>
+            <configuration>
+              <rules>
+                <!--<rule>
+                  <element>CLASS</element>
+                  <limits>
+                   <limit>
+                      <counter>LINE</counter>
+                      <value>COVEREDRATIO</value>
+                      <minimum>0.80</minimum>
```
artursouza: Change it to a small number to make it pass. We must change it once we are code complete.

sdk/src/test/java/io/dapr/actors/client/ActorProxyImplTest.java
```diff
@@ -4,18 +4,290 @@
...
-        final ActorProxyHttpAsyncClient actorProxyAsyncClient = (ActorProxyHttpAsyncClient)new ActorProxyClientBuilder().buildAsyncClient();
-        final ActorProxyImpl actorProxy= new ActorProxyImpl(
-          "myActorType",
-          new ActorId("100"),
-          new ActorStateSerializer(),
-          actorProxyAsyncClient);
-        Assert.assertEquals(actorProxy.getActorId().toString(),"100");
-        Assert.assertEquals(actorProxy.getActorType(),"myActorType");
+        final ActorProxyHttpAsyncClient actorProxyAsyncClient = mock(ActorProxyHttpAsyncClient.class);
+        final ActorStateSerializer serializer = mock(ActorStateSerializer.class);
```
artursouza: Please, use a real serializer. It is OK since it does not do IO and it must uncover more problems with this implementation.

sdk/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
```diff
@@ -125,8 +144,8 @@ public String getActorType() {
      * @param <T>      Expected response type.
      * @return Response object, null or RuntimeException.
      */
-    private <T> T unwrap(final String response, Class<T> clazz) {
-        return this.serializer.unwrapMethodResponse(response, clazz);
+    private <T> T unwrap(final String response, Class<T> clazz) throws IOException {
+        return this.serializer.deserialize(response, clazz);
```
artursouza: I wonder if this can actually work. There is a reason we use the unwrapMethodResponse - I used the example to manually validate end-to-end and that drove this change. I see that you are mocking serializer, so it might be why it does not fail in the unit tests. See my comment on using a real Serializer instance.

artursouza:
Please, change description to refer to issues being addressed by this change - even if just partially.

sdk/src/test/java/io/dapr/actors/client/ActorProxyImplTest.java
JuanJose-Herrera: Agree, but this is just to test the constructor, that is that I  inject the mock, which is just for this test.

sdk/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
JuanJose-Herrera: The mock serializer is just for the constructor, the reason that I change this, is that doing some testing using the examples for .net-SDK, the response to call an actor method is not wrapped in a data node in the JSON response, that the unwrapMethodResponse expect inside the serializer object, please let me know if I am wrong in the assumption that the response does not contains  the "data" wrapper, or if I should change the method  unwrapMethodResponse in order to be used in both cases (with data wrapper and without data wrapper)

artursouza: Can you test this using the example for Actors? It will tell us if this will work or not.

sdk/src/test/java/io/dapr/actors/client/ActorProxyImplTest.java
artursouza: OK.

sdk/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
JuanJose-Herrera: I already test this with the Actor example, you were right the response is a base64 under data, I already change the unit test, and the code to expect this kind of answers from Dapr. Thanks for your help.

Issue #22, opened 2019-12-06
[java-sdk] Dapr's base Client


artursouza:
* Define interface for the Dapr's Client - wrapper on top of the auto-generated code.
* Implement Adapter for GRPC.
* Implement Adapter for HTTP.

Detail below:

Interface DaprClient { // customer will use this interface.
// methods defined as part of this work.
// actor APIs re defined here too.
}

Class GrpcDapClient implements DaprClient { // customers don’t talk to this one directly
  DaprGrpc grpcClient;
 // implements methods from interface
// throws exceptions for all actor APIs since actors are http only for now.
}

Class HttpDaprClient implements DaprClient { // customers don’t talk to this directly.
  DaprHttpClient httpClient;
 // implements methods from interface
// implements actors APIs as well
}

artursouza:
We have auto-generated client in Grpc and one manually generated for HTTP. This base client can sit on top of those now.

PR #68, opened 2020-01-01
Refactoring HTTP Client and creating Client HTTP Adapter

AndresRoblesMX:
# Description

Refactoring HTTP Client to follow same pattern as the GRPC Client. 
Creating DaprClientHttpAdapter to be exposed to the users.
Refactor all uses of the http Client to use the adapter instead following composition rather than inheritance.
Dealing with conflicts

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/22

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


yaron2:
This PR looks great. I'll start reviewing soon.

sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -20,7 +20,7 @@
      * @param <T>   The type of event to be publishded.
      * @return a Mono plan of type Void
      */
-    <T> Mono<Void> publishEvent(T event);
+    <T> Mono<Void> publishEvent(ClientRequest<T> event);
```
artursouza: This seems to make the interface specific to http

AndresRoblesMX: I did not found anything specific to GPRC that is needed beyond the actual value to be sent over, unlike for http who needs more data, do you have any suggestion on how to handle that?

yaron2:
We might want to refactor a little more.

Since users are not expected to use the HttpDaprClient or GrpcDaprClient directly but instead talk to the Adapter (which really acts like a facade to the Dapr API), the questions remains on why do we need wrappers on top of the autogenerated gRPC libraries and native HTTP.

All of the Dapr functionality can be accessed today in gRPC except for actors.
The DaprClientAdapter is then free to use gRPC directly for everything other than actors, where it'll call Dapr using HTTP.

Once [this](https://github.com/dapr/dapr/issues/927) issue is complete, SDKs will be able to use gRPC only and not HTTP.

In addition, the DaprClientAdapter will need to host either a web-server or gRPC server for Dapr to communicate with it (for example, getting published events, binding events, etc.).
If we settle on gRPC, we not only gain considerable performance improvements (imperative to Java), but also we solve the "which web framework do we use" conundrum.

Given the above, combined with simplicity and maintainability, would it not be easier to refactor to this?

AndresRoblesMX:
Even when GRPC will provide a better performance dapr will still provide
both ways to comunicate to the API, so we need to provide both ways through
a common interface for simplicity, that's why it was designed this way,
settling for GRPC for everything unless is not provided will limit the Java
SDK in comparison with the other languages SDKs which is not ideal.
The users of the SDK should be provided with both supported ways and let
them decide which one is better for them.

yaron2:
I disagree.

The whole purpose of the SDK is to be prescriptive and simplify the life for the developer.

A developer consumes an API, not a communication protocol.
As a developer using an SDK, you want to have the ability to do X: you don't care if X is done over HTTP or gRPC under the covers.

Also, when you say "other language SDKs", remember that all of them are gRPC only except for the dotnet SDK which is headed that direction too.

artursouza:
@yaron2 and @AndresRoblesMX 

The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to *one* underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.

The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.

Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.

sdk/src/main/java/io/dapr/actors/client/ActorProxyBuilder.java
```diff
@@ -39,6 +40,21 @@ public ActorProxyBuilder withPort(int port) {
...
+        this.clientBuilder.withHost(host);
+        return this;
+    }
+
+    public ActorProxyBuilder withThreadPoolSize(int threadPoolSize) {
+        this.clientBuilder.withThreadPoolSize(threadPoolSize);
+        return this;
+    }
+
+    public ActorProxyBuilder withOkHttpClientBuilder(OkHttpClient.Builder okHttpClientBuilder) {
```
artursouza: Let's not expose OKHttp to the layer above.

sdk/src/main/java/io/dapr/actors/client/ActorProxyHttpAsyncClient.java
```diff
@@ -4,24 +4,33 @@
...
+
+    private DaprClient daprClient;
 
     /**
      * Creates a new instance of {@link ActorProxyHttpAsyncClient}.
      *
      * @param port       Port for calling Dapr. (e.g. 3500)
      * @param httpClient RestClient used for all API calls in this new instance.
      */
-    ActorProxyHttpAsyncClient(int port, OkHttpClient httpClient) {
```
artursouza: The idea is to share the same httpClient for all instances. With this change, each instance will have its own client, which is expensive.

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprClientBuilder.java
```diff
@@ -18,8 +17,9 @@
      * @return Builds an async client.
      */
     public AppToDaprAsyncClient buildAsyncClient() {
```
artursouza: This class needs to go. Any place where this class is being used, should just be DaprClientBuilder.

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprHttpAsyncClient.java
```diff
@@ -4,205 +4,111 @@
...
 import reactor.core.publisher.Mono;
 
-import java.util.HashMap;
-import java.util.Map;
-
 /**
  * Http client to call Dapr's API for actors.
  */
 //public class DaprHttpAsyncClient implements DaprAsyncClient {
-class AppToDaprHttpAsyncClient extends AbstractDaprHttpClient implements AppToDaprAsyncClient {
```
artursouza: Move all implementation to AbstractDaprHttpClient - there is no need for this to be separate.

yaron2:
> @yaron2 and @AndresRoblesMX
> 
> The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> 
> The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> 
> Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.

We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.

So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.

It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.

artursouza:
> > @yaron2 and @AndresRoblesMX
> > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> 
> We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> 
> So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> 
> It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.

Once we decide to have a single implementation of the common interface, yes we can simplify.

yaron2:
> > > @yaron2 and @AndresRoblesMX
> > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > 
> > 
> > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> 
> Once we decide to have a single implementation of the common interface, yes we can simplify.

Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.

yaron2:
The reason why I press on this now is because changes like these are best done sooner rather than later. We've had our fair share of technical debt in the past..

artursouza:
> > > > @yaron2 and @AndresRoblesMX
> > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > 
> > > 
> > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > 
> > 
> > Once we decide to have a single implementation of the common interface, yes we can simplify.
> 
> Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.

I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.

My point is: calling out this change now is a big deal and is not the intent for this PR.

artursouza:
> > > > > @yaron2 and @AndresRoblesMX
> > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > 
> > > > 
> > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > 
> > > 
> > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > 
> > 
> > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> 
> I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> 
> My point is: calling out this change now is a big deal and is not the intent for this PR.

BTW, there is a more fundamental tech debt that I want to address here instead. I will post my detailed request soon - I am editing it now.

yaron2:
> > > > > @yaron2 and @AndresRoblesMX
> > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > 
> > > > 
> > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > 
> > > 
> > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > 
> > 
> > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> 
> I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> 
> My point is: calling out this change now is a big deal and is not the intent for this PR.

Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.

artursouza:
> > > > > > @yaron2 and @AndresRoblesMX
> > > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > > 
> > > > > 
> > > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > > 
> > > > 
> > > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > > 
> > > 
> > > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> > 
> > 
> > I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> > My point is: calling out this change now is a big deal and is not the intent for this PR.
> 
> Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.

My point is the opposite: keeping both adapters now and moving to the facade later is easier than if we decide the other way around.

yaron2:
> > > > > > > @yaron2 and @AndresRoblesMX
> > > > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > > > 
> > > > > > 
> > > > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > > > 
> > > > > 
> > > > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > > > 
> > > > 
> > > > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> > > 
> > > 
> > > I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> > > My point is: calling out this change now is a big deal and is not the intent for this PR.
> > 
> > 
> > Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.
> 
> My point is the opposite: keeping both adapters now and moving to the facade later is easier than if we decide the other way around.

First, can you explain why its easier?
And second, why keep both adapters now if its temporary? what goal does that serve?

artursouza:
> > > > > > > > @yaron2 and @AndresRoblesMX
> > > > > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > > > > 
> > > > > > > 
> > > > > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > > > > 
> > > > > > 
> > > > > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > > > > 
> > > > > 
> > > > > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> > > > 
> > > > 
> > > > I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> > > > My point is: calling out this change now is a big deal and is not the intent for this PR.
> > > 
> > > 
> > > Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.
> > 
> > 
> > My point is the opposite: keeping both adapters now and moving to the facade later is easier than if we decide the other way around.
> 
> First, can you explain why its easier?
We have the code for HTTP and GRPC with corresponding builders separate already. If we decide to combine those into one is easier, IMO. On the other hand, if we have a facade that partially implements both now and then we decide to split into HTTP and GRPC client we will need a bigger refactoring (IMO). Again, the code as-is is still not doing what I envisioned: the client is split into actors vs non-actor clients and we want to fix that now.

> And second, why keep both adapters now if its temporary? what goal does that serve?
It is temporary if we assume the decision has been made - AFAIK we will discuss this in the design review next week.

yaron2:
> > > > > > > > > @yaron2 and @AndresRoblesMX
> > > > > > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > > > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > > > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > > > > > 
> > > > > > > 
> > > > > > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > > > > > 
> > > > > > 
> > > > > > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> > > > > 
> > > > > 
> > > > > I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> > > > > My point is: calling out this change now is a big deal and is not the intent for this PR.
> > > > 
> > > > 
> > > > Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.
> > > 
> > > 
> > > My point is the opposite: keeping both adapters now and moving to the facade later is easier than if we decide the other way around.
> > 
> > 
> > First, can you explain why its easier?
> > We have the code for HTTP and GRPC with corresponding builders separate already. If we decide to combine those into one is easier, IMO. On the other hand, if we have a facade that partially implements both now and then we decide to split into HTTP and GRPC client we will need a bigger refactoring (IMO). Again, the code as-is is still not doing what I envisioned: the client is split into actors vs non-actor clients and we want to fix that now.
> 
> > And second, why keep both adapters now if its temporary? what goal does that serve?
> 
> It is temporary if we assume the decision has been made - AFAIK we will discuss this in the design review next week.

Very well. let's label this as needs-discussion and merge after we discuss.

artursouza:
> > > > > > > > > > @yaron2 and @AndresRoblesMX
> > > > > > > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > > > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > > > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > > > > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > > > > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > > > > > > 
> > > > > > > 
> > > > > > > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> > > > > > 
> > > > > > 
> > > > > > I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> > > > > > My point is: calling out this change now is a big deal and is not the intent for this PR.
> > > > > 
> > > > > 
> > > > > Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.
> > > > 
> > > > 
> > > > My point is the opposite: keeping both adapters now and moving to the facade later is easier than if we decide the other way around.
> > > 
> > > 
> > > First, can you explain why its easier?
> > > We have the code for HTTP and GRPC with corresponding builders separate already. If we decide to combine those into one is easier, IMO. On the other hand, if we have a facade that partially implements both now and then we decide to split into HTTP and GRPC client we will need a bigger refactoring (IMO). Again, the code as-is is still not doing what I envisioned: the client is split into actors vs non-actor clients and we want to fix that now.
> > 
> > 
> > > And second, why keep both adapters now if its temporary? what goal does that serve?
> > 
> > 
> > It is temporary if we assume the decision has been made - AFAIK we will discuss this in the design review next week.
> 
> Very well. let's label this as needs-discussion and merge after we discuss.

This is a WIP branch, do you really think we cannot merge this? I understand that we should agree before going to master. There is more work pending on top of this change, so if we block this then three things might happen:

1. A new WIP2 branch is created and code is merged there instead and then merged into here once a decision has been made.
2. A forked repo is used as the new WIP branch and changes pile up there instead.
3. The team pauses work on most of the Java SDK until a decision has been made.

What is your expectation by blocking this PR?

yaron2:
> > > > > > > > > > > @yaron2 and @AndresRoblesMX
> > > > > > > > > > > The "Adapter" layer is acting as a Adapter today because it is a translation between the common API to _one_ underlying raw API. Once we do a more complex integration (grpc for one vs http for other), then it becomes a Facade since it is simplifying multiple interfaces into a single common interface.
> > > > > > > > > > > The current strategy of having one adapter per raw client is aligned to the goal of having GRPC only. In that mode, we simply stop offering the http implementation and default to GRPC only, deprecating the HTTP option in the builder. This is not in scope for this milestone.
> > > > > > > > > > > Even if we decide to implement the "facade" solution where we use GRPC for some APIs and HTTP for others - it is just another implementation of the same interface (maybe even using the two adapters created here). I do see this change in line with our goal.
> > > > > > > > > > 
> > > > > > > > > > 
> > > > > > > > > > We don't need a wrapper layer around gRPC or HTTP. As far as I can see, they are redundant, as the only consumer of these is the Adapter layer today.
> > > > > > > > > > So having the Adapter (should probably be only called Client) use the autogenerated gRPC client to call into Dapr for everything besides actors, and making direct HTTP calls for actors is the way to go.
> > > > > > > > > > It simplifies the architecture, removing unneeded interfaces and wrappers and improves maintainability by having a single implementation instead of several.
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > Once we decide to have a single implementation of the common interface, yes we can simplify.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Can we make the decision? the common interface is very similar to the one now, only missing the custom classes per operation (for requests and responses) instead of generics.
> > > > > > > 
> > > > > > > 
> > > > > > > I remember us having this discussion about being GRPC only (except for Actors) before and it was agreed to support both in the SDKs. Let's add this to the review we will have soon. This change is not a one-way road - we can quickly switch to the GRPC+HTTP hybrid if we decide to. I am more concerned about the callback API - that will be heavily impacted if we decide on that.
> > > > > > > My point is: calling out this change now is a big deal and is not the intent for this PR.
> > > > > > 
> > > > > > 
> > > > > > Calling out this change later will be a bigger deal. If there's no reason to not go down the path of a single client using gRPC and HTTP directly, I don't see why we should insert work that we know will change later and will be more difficult to change.
> > > > > 
> > > > > 
> > > > > My point is the opposite: keeping both adapters now and moving to the facade later is easier than if we decide the other way around.
> > > > 
> > > > 
> > > > First, can you explain why its easier?
> > > > We have the code for HTTP and GRPC with corresponding builders separate already. If we decide to combine those into one is easier, IMO. On the other hand, if we have a facade that partially implements both now and then we decide to split into HTTP and GRPC client we will need a bigger refactoring (IMO). Again, the code as-is is still not doing what I envisioned: the client is split into actors vs non-actor clients and we want to fix that now.
> > > 
> > > 
> > > > And second, why keep both adapters now if its temporary? what goal does that serve?
> > > 
> > > 
> > > It is temporary if we assume the decision has been made - AFAIK we will discuss this in the design review next week.
> > 
> > 
> > Very well. let's label this as needs-discussion and merge after we discuss.
> 
> This is a WIP branch, do you really think we cannot merge this? I understand that we should agree before going to master. There is more work pending on top of this change, so if we block this then three things might happen:
> 
> 1. A new WIP2 branch is created and code is merged there instead and then merged into here once a decision has been made.
> 2. A forked repo is used as the new WIP branch and changes pile up there instead.
> 3. The team pauses work on most of the Java SDK until a decision has been made.
> 
> What is your expectation by blocking this PR?

My expectation is we make the best possible choice for the project based on technical merits and agreed upon decisions, and to not continue work on building blocks that may or may not exist after being merged and are awaiting discussion.

Based on what you've written, it seems like this is a major and crucial decision if most work done by the team is blocked on this, emphasizing further the importance of the architectural decision made here.

This is Friday after noon. Is it really that important we merge now if we discuss this beginning of next week?

sdk/src/main/java/io/dapr/actors/client/ActorProxyBuilder.java
AndresRoblesMX: No exposing the Http builder to the above layer will force us to use a client with only default configurations, which could be not very secure.

sdk/src/main/java/io/dapr/actors/runtime/AppToDaprHttpAsyncClient.java
AndresRoblesMX: you mean move this to DaprHttp.java? this is very specific to Actors, probably rename it to ActorsHttpClient and move it package io.dapr.actors.client?

sdk/src/main/java/io/dapr/actors/client/ActorProxyHttpAsyncClient.java
AndresRoblesMX: so for sharing the same httpClient I should be gettting the acual implementation rather than the builder?

artursouza:
Hi Andres,

I guess we are having problems with the refactoring because we have two clients and this is causing confusion. Also, after talking to Yaron, we should provide a single client builder and interface. Here is my proposal:

1.
Merge ActorProxyHttpAsyncClient and AppToDaprHttpAsyncClient into a single class called DaprHttpClientAdapter

2.
Merge AppToDaprAsyncClient and ActorProxyAsyncClient into the existing DaprClient interface

3.
Merge builders into a single builder (DaprClientBuilder) that will use HTTP only – let GRPC client unused for now. If we need to support multiple flavors, we can do it easily after this change.

4.
Change code that uses the classes above need to change to reflect that.

Thanks,
Artur Souza

yaron2:
After discussing, we agreed to:

1. Create a single user facing interface and client
2. Encapsulate the protocol from the user inside the client
3. Allow an _optional_ HashMap of configuration values should we decide to keep protocol support, and throw it out if not needed
4. To get something out quickly, we'll continue with HTTP at the moment, and transition to gRPC when the time comes.

AndresRoblesMX:
@yaron2 @artursouza already addressed the observations made, please take another look, and let me know what you think.

sdk/src/main/java/io/dapr/client/AbstractClientBuilder.java
```diff
@@ -5,55 +5,11 @@
 package io.dapr.client;
 
 import io.dapr.utils.Constants;
+import okhttp3.OkHttpClient;
 
 /**
  * Base class for client builders
  */
 public abstract class AbstractClientBuilder {
```
artursouza: Delete.

sdk/src/main/java/io/dapr/client/DaprClientGrpcAdapter.java
```diff
@@ -19,160 +23,234 @@
...
-     * {@inheritDoc}
-     */
-    @Override
-    public <T> Mono<Void> invokeBinding(T request) {
+  /**
+   * Operation not supported for GRPC
+   * @throws UnsupportedOperationException every time is called.
+   */
+  public <T> Mono<Void> invokeService(String verb, String appId, String method, T request) {
+    return Mono.error(new UnsupportedOperationException("Operation not supported for GRPC"));
```
artursouza: Actually, this is supported.

sdk/src/main/java/io/dapr/client/domain/StateKeyValue.java
```diff
@@ -0,0 +1,56 @@
+package io.dapr.client.domain;
+
+public class StateKeyValue<T> {
+  private T value;
+  private String key;
+  private String etag;
+  private StateOptions options;
```
artursouza: We should not get this argument from user. If this is only for internal use, then there should be another class to represent this composition of KeyValue + StateOptions.

Also, make this class immutable. There should be no need to change it.

sdk/src/main/java/io/dapr/client/domain/StateOptions.java
```diff
@@ -0,0 +1,13 @@
+package io.dapr.client.domain;
+
+public class StateOptions {
```
artursouza: Immutable.

sdk/src/main/java/io/dapr/utils/Constants.java
```diff
@@ -17,13 +17,20 @@
     /**
      * Dapr's default hostname.
      */
-    public static final String DEFAULT_HOSTNAME = "localhost";
+    public static final String DEFAULT_HOSTNAME = "http://localhost";
```
artursouza: hostname is "localhost", consumers of that constant expect it to be an actual hostname (without "http://").
"http://localhost/" should be called DEFAULT_HTTP_BASE_URL or something.

sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -4,70 +4,172 @@
...
+
+  /**
+   * Saves state batch to Dapr.
+   *
+   * @param actorType Type of actor.
+   * @param actorId   Actor Identifier.
+   * @param data      State to be saved.
+   * @return Asynchronous void result.
+   */
+  Mono<Void> saveStateTransactionally(String actorType, String actorId, String data);
```
artursouza: Use Actor is every actor related method for consistency.

sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
```diff
@@ -5,66 +5,59 @@
...
-     * Sets the port to be used by the client
+     * Creates an instance of the builder setting the type of client to be creted
      *
-     * @param port Port to connect to.
-     * @return itself
      */
-    public DaprClientBuilder withPort(Integer port) {
-        this.port = port;
-        return this;
+    public DaprClientBuilder() {
```
artursouza: You probably don't need this constructor anymore.

sdk/src/main/java/io/dapr/client/domain/StateKeyValue.java
AndresRoblesMX: That's there because it's needed like that for GRPC

yaron2: **Important**

This structure is the following:

1. For both gRPC and HTTP (mentioned before)
2. Is not internal. It's provided by the user.

sdk/src/main/java/io/dapr/client/AbstractClientBuilder.java
AndresRoblesMX: Forgot to delete this class, which is no longer required.

AndresRoblesMX:
Comments have been addressed, can you guys let me know if any other comment.

Issue #30, opened 2019-12-09
[java-sdk] New jar to allow Dapr's callback API in SpringBoot


artursouza:
A SpringBoot Application can easily depend on this jar and have Dapr's callback API running with a simple configuration change and implementing a base class/interface.

As an alternative to a base class, annotations can also be used, similar to how the .Net SDK does with ASP.Net.

artursouza:
@ji11er

ji11er:
Hi Artur,
I assume I follow the spec written here:
https://github.com/dapr/docs/tree/master/reference/api

right?,

Thanks

Issue #31, opened 2019-12-09
[java-sdk] Sample app using SpringBoot


artursouza:
Create a sample app with SpringBoot using the SpringBoot specific jar from Issue #30 .

artursouza:
Current examples already use SpringBoot. dapr-sdk-springboot.jar will not be published anymore, users will only import dapr-sdk.jar and existing code from dapr-sdk-springboot.jar will be moved into examples.

PR #69, opened 2020-01-07
Starting package to use SpringBoot as web API for dapr.

artursouza:
# Description

Starts work to use SpringBoot as web api + migrates Actor example to use SpringBoot.

## Issue reference

Please reference the issue this PR will close: #30 #31 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


yaron2:
I'm confused about the term "integration": this is for the SDK using SpringBoot internally to get callbacks from Dapr, no?

yaron2:
Or is this about calling back into the user's app using SpringBoot?

artursouza:
> Or is this about calling back into the user's app using SpringBoot?

This is about using SpringBoot to get calls from Dapr, using a SpringBoot aplication provided by dapr sdk.

yaron2:
It's not a biggie, but I wouldn't necciseraly use the term integration as it generally denotes some sort of feature wise extensions between two systems, where here it's Dapr using Spring Boot for it's server purpose.

Wdyt?

artursouza:
> It's not a biggie, but I wouldn't necciseraly use the term integration as it generally denotes some sort of feature wise extensions between two systems, where here it's Dapr using Spring Boot for it's server purpose.
> 
> Wdyt?

OK, the term "integration" can create ambiguity.

sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
```diff
@@ -0,0 +1,135 @@
...
+ * SpringBoot Controller to handle callback APIs for Dapr.
+ * TODO: use POJOs instead of String when possible.
+ * TODO: JavaDocs.
+ */
+@RestController
+public class DaprController {
+
+  private static final JsonFactory JSON_FACTORY = new JsonFactory();
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
olegz: You should probably autowire it in the constructor and only create a new instance if one is not available. Also, having it being managed would allow user's to immediately provide their own version or add additional configurations

sdk-springboot/src/main/java/io/dapr/springboot/DaprApplication.java
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Starts Dapr's callback in a given port.
+   * @param port Port to listen to.
+   */
+  public static void start(int port) {
+    SpringApplication app = new SpringApplication(DaprApplication.class);
+    Properties properties = new Properties();
+    properties.setProperty("server.port", Integer.toString(port));
+    app.setDefaultProperties(properties);
+    app.run();
```
olegz: You should be able to simply call `main("--server.port=" +  port)` instead of these 5 lines of code.

sdk-springboot/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot</artifactId>
+      <version>2.2.2.RELEASE</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-web</artifactId>
+      <version>2.2.2.RELEASE</version>
```
olegz: You should probably externalise boot versions, but a better approach would probably be to have parent as
```
<parent>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-parent</artifactId>
	<relativePath /> <!-- lookup parent from repository -->
</parent>
```
as pretty much all boot-based projects have. In fact you can go to https://start.spring.io/ and generate a simplest project and us it's pom as a guide on what this one should be

artursouza: We wanted to keep our jars under the same parent and let the user application be a child of SpringBoot. So, I went through the examples and eventually figured out a way to accomplish this.

What do you mean by externalize boot version?

sdk-springboot/src/main/java/io/dapr/springboot/DaprApplication.java
artursouza: Will do.

sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
artursouza: We are discussing the serialization thing here. Thanks for the feedback. What is the advantage of using autowire?

PR #70, opened 2020-01-07
Request body must be null on get for okhttp, empty body does not work…

LMWF:
# Description

This fixes errors on http get calls.  Request body must be null on get for okhttp, empty body does not work. Also, expose the ActorStateManager methods that ought to be exposed to user.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/actors/runtime/ActorStateManager.java
+++ b/sdk/src/main/java/io/dapr/actors/runtime/ActorStateManager.java
@@ -15,7 +15,7 @@ import java.util.*;
  * <p>
  * All changes are cached in-memory until save() is called.
  */
-class ActorStateManager {
+public class ActorStateManager {
 
     /**
      * Provides states using a state store.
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/io/dapr/actors/runtime/ActorStateManager.java
```diff
@@ -44,7 +44,7 @@
      * @param actorTypeName Name of Actor's type.
      * @param actorId       Actor's identifier.
      */
-    ActorStateManager(DaprStateAsyncProvider stateProvider, String actorTypeName, ActorId actorId) {
+    public ActorStateManager(DaprStateAsyncProvider stateProvider, String actorTypeName, ActorId actorId) {
```
artursouza: Keep the constructor package visible only.

sdk/src/main/java/io/dapr/client/AbstractDaprHttpClient.java
```diff
@@ -105,7 +105,7 @@ public AbstractDaprHttpClient(int port, OkHttpClient httpClient) {
 
                         Request request = new Request.Builder()
                                 .url(new URL(this.baseUrl + urlString))
-                                .method(method, body)
+                                .method(method, (json == null && method == "GET") ? null : body)
```
artursouza: Sorry, one more problem here. method.equals("GET").

Issue #26, opened 2019-12-08
New example using Invoke over Grpc


artursouza:
Create a new example using Invoke over Grpc
Add README.md with instructions.

PR #74, opened 2020-01-09
#26 Integretion Testing Initial Example

JuanJose-Herrera:
… to work on MAC and Linux

# Description

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static DaprIntegrationTestingRunner createDaprIntegrationTestingRunner(String successMessage, Class serviceClass, Boolean isGrpc, int port, int sleepTime) {
+        return new DaprIntegrationTestingRunner(successMessage, serviceClass, isGrpc, port, sleepTime);
+    }
+
+
+    public void initializeDapr() throws IOException, InterruptedException, TimeoutException, ExecutionException {
+
+        File file= new File("../");
```
artursouza: No need to use this to depend on current directory location. See my next comment.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,87 @@
...
+        final ExecutorService executor = Executors.newSingleThreadExecutor();
+        final Future future = executor.submit(stuffToDo);
+        executor.shutdown(); // This does not cancel the already-scheduled task.
+        future.get(2, TimeUnit.MINUTES);
+        Thread.sleep(sleepTime);
+
+    }
+
+    private static final String DAPR_RUN = "dapr run --app-id %s ";
+    private static final String DAPR_COMMAND = " -- mvn exec:java -pl=sdk -D exec.mainClass=%s -D exec.classpathScope=\"test\"";
```
artursouza: You don't actually need to start the app here. You can just start dapr without the service and then start the test app directly in the same JVM (separate thread).

Make DAPR_COMMAND = " -- cat" to just block.
Start a new thread, starting the service: io.dapr.e2e.apps.MyService.main()

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private void generateAppName(){
+
+        this.appName=UUID.randomUUID().toString();
+    }
+
+    public  void destroyDapr() throws IOException {
+        rt.exec("dapr stop --app-id " + this.appName);
+        proc.destroy();
```
artursouza: Do you need to wait for sometime before calling destroy?

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+
+    public void initializeDapr() throws IOException, InterruptedException, TimeoutException, ExecutionException {
+
+        File file= new File("../");
+        proc= rt.exec(this.buildDaprCommand(), null,file);
+        InputStream stdin = proc.getInputStream();
+        InputStreamReader isr = new InputStreamReader(stdin);
+        BufferedReader br = new BufferedReader(isr);
```
artursouza: You might need to close these readers, not?

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
```diff
@@ -0,0 +1,44 @@
...
+ * Simple example, to run:
+ * mvn clean install
+ * dapr run --grpc-port 50001 -- mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.Example
+ */
+public class HelloWorldGrpcStateService {
+
+  public static void main(String[] args) {
+    ManagedChannel channel =
+      ManagedChannelBuilder.forAddress("localhost", 50001).usePlaintext().build();
+    DaprBlockingStub client = DaprGrpc.newBlockingStub(channel);
```
artursouza: Indentation.

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.concurrent.TimeoutException;
+
+
+public class HelloWorldClientIT {
+
+
+    private static DaprIntegrationTestingRunner daprIntegrationTestingRunner;
+
+    @BeforeClass
+    public static void init() throws IOException, InterruptedException, TimeoutException, ExecutionException {
```
artursouza: Just throws Exception for simplicity. It is OK for tests.

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
```diff
@@ -0,0 +1,44 @@
...
+      String value = "Hello World";
+      StateRequest req = StateRequest
+        .newBuilder()
+        .setKey(key)
+        .setValue(Any.newBuilder().setValue(ByteString.copyFromUtf8(value)).build())
+        .build();
+      SaveStateEnvelope state = SaveStateEnvelope.newBuilder()
+        .addRequests(req)
+        .build();
+      client.saveState(state);
```
artursouza: For this particular example, you did not need a separate service class because it is not listening on any port, just invoking dapr. See my comment above: there is a way to start Dapr faking an app just to have it running locally.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
JuanJose-Herrera: I am getting the following error with the cat command, even when that work on my terminal
image

But using DAPR on windows 10, I am getting the following:
exec: "cat": executable file not found in %!P(MISSING)ATH%!(NOVERB)

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
JuanJose-Herrera: No, after test cases run, everything is ready to stop the application

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
JuanJose-Herrera: I already remove this unneeded file path

sdk/src/main/java/io/dapr/runtime/Dapr.java
```diff
@@ -64,87 +82,179 @@ public static DaprRuntime getInstance() {
...
       return Mono.error(new DaprException("INVALID_METHOD_OR_TOPIC", "Did not find handler for : " + (name == null ? "" : name)));
     }
 
     try {
-      Map<String, Object> map = parse(payload);
-      String messageId = map.getOrDefault("id", "").toString();
-      String dataType = map.getOrDefault("datacontenttype", "").toString();
-      byte[] data = this.serializer.deserialize(map.get("data"), byte[].class);
-      return handler.apply(new HandleRequest(messageId, dataType, data, metadata)).switchIfEmpty(EMPTY_BYTES_ASYNC);
+      Message message = this.serializer.deserialize(payload, Message.class);
```
LMWF: Does this only work for topics (pub/sub) messages?  It looks like it's used for both subscriptions and method invocations from the way registrations are done, but this call seems to result in a later call deserializeTopicMessage() which looks like it's only for subscriptions.  Also, what format does `Message` correspond to?  It seems to resemble the CloudEventsEnvelope in:

https://github.com/dapr/components-contrib/blob/ede980c7ff320bc38f3284d12793910f4f4f4d93/pubsub/envelope.go

but it doesn't match exactly.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
JuanJose-Herrera: yes, now we use the try(resource) approach
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
JuanJose-Herrera: please review again, let me know if that works

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
JuanJose-Herrera: yes, the idea is to fail with an exception the complete build

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
JuanJose-Herrera: I can not resolve the problem using a command like CAT or something similar, so I create an empty class, that allows running DAPR without any service running. I use this approach to test HTTP state functionality.

sdk/src/main/java/io/dapr/runtime/Dapr.java
LMWF: Sorry this was posted on the wrong PR.

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Simple example, to run:
+ * mvn clean install
+ * dapr run --grpc-port 50001 -- mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.Example
+ */
+public class HelloWorldGrpcStateService {
+
+    public static void main(String[] args) {
+        ManagedChannel channel = ManagedChannelBuilder.forAddress("localhost", 50002).usePlaintext().build();
```
artursouza: DAPR_HTTP_PORT or DAPR_GRPC_PORT env vars.

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
```diff
@@ -0,0 +1,92 @@
...
+                        50002,
+                        2000
+                        );
+        daprIntegrationTestingRunner.initializeDapr();
+    }
+
+    @Test
+    public void testHelloWorldState(){
+        ManagedChannel channel =
+                ManagedChannelBuilder.forAddress("localhost", 50002).usePlaintext().build();
```
artursouza: Here the env variable will not be set. So, we need some code to find a free port,

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
```diff
@@ -0,0 +1,92 @@
...
+    private static DaprIntegrationTestingRunner daprIntegrationTestingRunner;
+
+    @BeforeClass
+    public static void init() throws IOException, InterruptedException, TimeoutException, ExecutionException {
+         daprIntegrationTestingRunner =
+                DaprIntegrationTestingRunner.createDaprIntegrationTestingRunner(
+                        "BUILD SUCCESS",
+                        HelloWorldGrpcStateService.class,
+                        true,
+                        50002,
```
artursouza: Auto detect a free port and send it here.

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
```diff
@@ -0,0 +1,92 @@
...
+                    .build();
+            DaprProtos.GetStateResponseEnvelope response = client.getState(req);
+            String value = response.getData().getValue().toStringUtf8();
+            System.out.println("Got: " + value);
+            Assert.assertEquals("",value);
+        }
+    }
+
+    @Test
+    public void test2(){
```
artursouza: Empty test?

sdk/src/test/java/io/dapr/it/state/HttpStateClientIT.java
```diff
@@ -0,0 +1,113 @@
...
+    private static DaprIntegrationTestingRunner daprIntegrationTestingRunner;
+
+    @BeforeClass
+    public static void init() throws IOException, InterruptedException, TimeoutException, ExecutionException {
+        daprIntegrationTestingRunner =
+                DaprIntegrationTestingRunner.createDaprIntegrationTestingRunner(
+                        "BUILD SUCCESS",
+                        EmptyService.class,
+                        false,
+                        3500,
```
artursouza: send a auto-detected free port.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
artursouza: Technically it can be any blocking command.

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
artursouza: My comment was just to do the following:
public static void init() throws Exception {

Instead of
public static void init() throws IOException, InterruptedException, TimeoutException, ExecutionException {

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
artursouza: OK.

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
JuanJose-Herrera: ok, I will

JuanJose-Herrera: this is done

sdk/src/test/java/io/dapr/it/services/HelloWorldGrpcStateService.java
JuanJose-Herrera: I already change this code, with auto-generate free ports code

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
JuanJose-Herrera: done

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
JuanJose-Herrera: done

sdk/src/test/java/io/dapr/it/state/HelloWorldClientIT.java
JuanJose-Herrera: removed

sdk/src/test/java/io/dapr/it/state/HttpStateClientIT.java
JuanJose-Herrera: done

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final String DAPR_RUN = "dapr run --app-id %s ";
+    private static final String DAPR_COMMAND = " -- mvn exec:java -D exec.mainClass=%s -D exec.classpathScope=\"test\" -Dexec.args=\"-p %d -grpcPort %d -httpPort %d\"";
+
+    private String buildDaprCommand(){
+        StringBuilder stringBuilder= new StringBuilder(String.format(DAPR_RUN, this.appName))
+                .append(this.useAppPort ? "--app-port " + this.daprFreePorts.appPort : "")
+                .append(" --grpc-port ")
+                .append(this.daprFreePorts.grpcPort)
+                .append(" --port ")
+                .append(this.daprFreePorts.httpPort)
```
youngbupark: I would not recommend to set these options. Rather, user code can read `DAPR_HTTP_PORT` and `DAPR_GRPC_PORT` which dapr cli sets.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+
+
+    public DaprFreePorts initializeDapr() throws Exception {
+
+
+        String daprCommand=this.buildDaprCommand();
+        System.out.println(daprCommand);
+        proc= rt.exec(daprCommand);
```
artursouza: Use ProcessBuilder instead. This way we don't need to forcefully set an env variable anymore.
https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,155 @@
...
+            try {
+                runtime.exec("dapr stop --app-id " + this.appName);
+            } catch (IOException e) {
+               throw new RuntimeException(e);
+            }
+        });
+        Optional.ofNullable(proc).ifPresent(process -> process.destroy());
+    }
+
+
```
artursouza: Keep single spaces.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,155 @@
...
+            return grpcPort;
+        }
+
+
+
+        public int getHttpPort() {
+            return httpPort;
+        }
+
+
```
artursouza: reformat code.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -0,0 +1,155 @@
...
+
+        public int getAppPort() {
+            return appPort;
```
artursouza: spacing is "2" and not "4".

sdk/pom.xml
```diff
@@ -62,14 +62,25 @@
...
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>com.github.gmazzo</groupId>
       <artifactId>okhttp-mock</artifactId>
       <version>1.3.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
```
artursouza: See my comment about a solution to avoid this dependency.

Issue #66, opened 2019-12-30
Reach parity with Dapr features


yaron2:
## Describe the proposal

Missing functionality that needs to be addressed:

1. Subscribe to topics
2. Bindings callbacks
3. Service invocation callback

I assume this is likely to be implemented in DaprClient

yaron2:
These are currently manually implemented by the user by either exposing a web server with the required endpoints, or implementing the `daprclient` gRPC interface.

artursouza:
Thanks, Yaron. I am aware of these gaps. Thanks for creating the issue.

LMWF:
The part of this issue related to bindings: https://github.com/dapr/java-sdk/pull/81/

PR #78, opened 2020-01-10
Invoke service + unit tests.

artursouza:
# Description

Invoke service + unit tests.

## Issue reference

Please reference the issue this PR will close: #66 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprClient.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClient.java
@@ -6,6 +6,7 @@ package io.dapr.client;
 
 import io.dapr.client.domain.StateKeyValue;
 import io.dapr.client.domain.StateOptions;
+import io.dapr.client.domain.Verb;
 import reactor.core.publisher.Mono;
 
 import java.util.List;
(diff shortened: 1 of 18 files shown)


examples/src/main/java/io/dapr/examples/invoke/http/HelloWorldClient.java
```diff
@@ -0,0 +1,28 @@
...
+import io.dapr.client.DaprClientBuilder;
+import io.dapr.client.domain.Verb;
+
+/**
+ * 1. Build and install jars:
+ * mvn clean install
+ * 2. Send messages to the server:
+ * dapr run --port 3006 -- mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.invoke.http.HelloWorldClient -Dexec.args="'message one' 'message two'"
+ */
+public class HelloWorldClient {
```
LMWF: This file and class should have a different name, maybe ServiceInvoker or ServiceInvokerExample.

examples/src/main/java/io/dapr/examples/invoke/http/HelloWorldService.java
```diff
@@ -0,0 +1,81 @@
...
+import static io.dapr.examples.DaprExamplesProtos.SayRequest;
+import static io.dapr.examples.DaprExamplesProtos.SayResponse;
+
+/**
+ * 1. Build and install jars:
+ * mvn clean install
+ * 2. Run in server mode:
+ * dapr run --app-id hellohttp --app-port 3000 --port 3005 -- mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.invoke.grpc.HelloWorldService -Dexec.args="-p 3000"
+ */
+public class HelloWorldService {
```
LMWF: Same with this one.

sdk/src/test/java/io/dapr/runtime/DaprRuntimeTest.java
```diff
@@ -189,4 +280,46 @@ private static final String generateMessageId() {
   private static final Map<String, String> generateSingleMetadata() {
     return Collections.singletonMap(UUID.randomUUID().toString(), UUID.randomUUID().toString());
   }
+
+  private static final class Message {
```
LMWF: Does this correspond to CloudEventsEnvelope?  

https://github.com/dapr/components-contrib/blob/ede980c7ff320bc38f3284d12793910f4f4f4d93/pubsub/envelope.go#L23

It doesn't seem to have all the fields.

sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -21,55 +22,82 @@
...
+   * @param appId   The Application ID where the service is.
+   * @param method  The actual Method to be call in the application.
+   * @param request The request to be sent to invoke the service.
+   * @param clazz   the Type needed as return for the call.
+   * @param <T>     the Type of the return, use byte[] to skip serialization.
+   * @param <K>     The Type of the request, use byte[] to skip serialization.
+   * @return A Mono Plan of type clazz.
    */
-  <T, K> Mono<T> invokeService(String verb, String appId, String method, K request, Class<T> clazz);
+  <T, K> Mono<T> invokeService(Verb verb, String appId, String method, K request, Class<T> clazz);
```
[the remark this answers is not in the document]
amanbha: We need methods to cover following:
1- Call with a request Type and response type.
2. Call with request Type but no response type.
3. Call with no request type but with response type.
4. Call without no request type and no response type.

sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -78,44 +106,44 @@
...
    *
-   * @param state        The key of the State to be retrieved
+   * @param state        The key of the State to be retrieved.
    * @param stateOptions
    * @param clazz        the Type of State needed as return.
-   * @param <T>          the Type of the return
-   * @param <K>          The Type of the key of the State
-   * @return A Mono Plan for the requested State
+   * @param <T>          the Type of the return.
+   * @param <K>          The Type of the key of the State.
```
amanbha: K is not needed.

amanbha: stateOptions can be an optional parameter.

sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -78,44 +106,44 @@
...
   /**
    * Save/Update a list of states.
    *
-   * @param states  the States to be saved
-   * @param options the Options to use for each state
-   * @param <T>     the Type of the State
-   * @return a Mono plan of type Void
+   * @param states  the States to be saved.
+   * @param options the Options to use for each state.
+   * @param <T>     the Type of the State.
```
amanbha: For bulk save state, T can be different for each state. Similarly StateOptions can be different for each state. This api signature needs to be revisited.

amanbha: For these state apis, we would need metadata (it could be a map)

sdk/src/main/java/io/dapr/client/DaprClient.java
amanbha: A metadata would be needed for the api call (metadata could be a map of string, string)

examples/src/main/java/io/dapr/examples/invoke/http/HelloWorldClient.java
artursouza: Will do.

examples/src/main/java/io/dapr/examples/invoke/http/HelloWorldService.java
artursouza: Will do.

sdk/src/test/java/io/dapr/runtime/DaprRuntimeTest.java
artursouza: No. This is just for this test. It is all the input we need to test a message.

sdk/src/main/java/io/dapr/client/DaprClient.java
artursouza: By optional, you mean adding it to the comment, right? Java does not have a way to specify a default value for a parameter to make it optional.

sdk/src/main/java/io/dapr/client/DaprClient.java
artursouza: The idea was to have variations where the metadata is optional. If we do all the permutations of request/response types, adding more to this would pollute our APIs with too many combinations (IMO). I can make metadata present in all of them, then.

sdk/src/main/java/io/dapr/client/DaprClient.java
artursouza: Yes, this is not the scope of this change.

sdk/src/main/java/io/dapr/client/DaprClient.java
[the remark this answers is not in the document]
artursouza: What is the value of these combinations? IMO, there is more value in the following (all with generics):
Call with input and response objects.
Call with input but void response.
Call with no input but response object.

examples/src/main/java/io/dapr/examples/invoke/http/HelloWorldClient.java
artursouza: Renamed it to InvokeClient.

examples/src/main/java/io/dapr/examples/invoke/http/HelloWorldService.java
artursouza: Renamed it to DemoService.

sdk/src/test/java/io/dapr/runtime/DaprRuntimeTest.java
artursouza: I got your comment. It is about the other Message class (not private). I will rename it and make it part of the domain for our API. I will also add the other attributes. It is more extensible that way.

sdk/src/main/java/io/dapr/client/DaprClient.java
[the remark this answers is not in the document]
artursouza: As we talked, this is about the input/output and not the generic types themselves. Wiill do all permutations.

On single letter generics in *Java*, see examples of style we follow:

https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheManagerCustomizer.java

https://github.com/Azure/autorest-clientruntime-for-java/blob/master/azure-client-runtime/src/main/java/com/microsoft/azure/AzureServiceFuture.java

artursouza: Will keep metadata in all and permute on Input/Output instead.

sdk/src/test/java/io/dapr/runtime/DaprRuntimeTest.java
LMWF: Right I pointed at the wrong Message class above, meant the other one that it not test-only.

PR #88, opened 2020-01-13
Fixing url builder

mestizoLopez:
# Description

We found and issue, regarding url builder, but with this modifications is fixed


artursouza:
No need for this PR. Issue has been fixed via another PR: #93

PR #102, opened 2020-01-15
Fixing bug, where call should be deferred instead of async

AndresRoblesMX:
# Description

Fixing bug, where call should be deferred instead of async

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprHttp.java
+++ b/sdk/src/main/java/io/dapr/client/DaprHttp.java
@@ -132,7 +132,7 @@ class DaprHttp {
    * @return Asynchronous text
    */
   public Mono<Response> invokeAPI(String method, String urlString, Map<String, String> urlParameters, byte[] content, Map<String, String> headers) {
-    return Mono.fromFuture(CompletableFuture.supplyAsync(
+    return Mono.fromCallable(
         () -> {
           try {
             String requestId = UUID.randomUUID().toString();
@@ -190,7 +190,7 @@ class DaprHttp {
           } catch (Exception e) {
             throw new RuntimeException(e);
           }
-        }, this.pool));
+        });
   }
 
   /**
--- a/sdk/src/main/java/io/dapr/client/DaprHttp.java
+++ b/sdk/src/main/java/io/dapr/client/DaprHttp.java
@@ -84,11 +84,6 @@ class DaprHttp {
   private final OkHttpClient httpClient;
 
   /**
-   * Thread-pool for HTTP calls.
-   */
-  private final ExecutorService pool;
-
-  /**
    * Creates a new instance of {@link DaprHttp}.
    *
    * @param port       Port for calling Dapr. (e.g. 3500)
@@ -97,7 +92,6 @@ class DaprHttp {
   DaprHttp(int port, OkHttpClient httpClient) {
     this.port = port;
     this.httpClient = httpClient;
-    this.pool = Executors.newWorkStealingPool();
   }
 
   /**
@@ -185,7 +179,7 @@ class DaprHttp {
               response.headers().forEach(pair -> {
                 mapHeaders.put(pair.getFirst(), pair.getSecond());
               });
-              return new Response(result, mapHeaders, response.code());
+              return new Response(result.length > 0 ? result : null, mapHeaders, response.code());
             }
           } catch (Exception e) {
             throw new RuntimeException(e);
--- a/sdk/src/test/java/io/dapr/client/DaprHttpTest.java
+++ b/sdk/src/test/java/io/dapr/client/DaprHttpTest.java
@@ -4,6 +4,8 @@
  */
 package io.dapr.client;
 
+import io.dapr.exceptions.DaprException;
+import io.dapr.utils.Constants;
 import io.dapr.utils.ObjectSerializer;
 import okhttp3.*;
 import okhttp3.mock.Behavior;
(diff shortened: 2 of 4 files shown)


sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -190,7 +190,7 @@ public int getStatusCode() {
           } catch (Exception e) {
             throw new RuntimeException(e);
           }
-        }, this.pool));
+        });
```
artursouza: If threadpool is not used anymore, please remove that from the code.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -177,16 +179,48 @@ public void invokePostMethodUnknownError() throws IOException {
...
                 .respond(500, ResponseBody.create(MediaType.parse("application/json"),
                         "{\"errorCode\":\"null\",\"message\":\"null\"}"));
-
         DaprHttp daprHttp = new DaprHttp(3500, okHttpClient);
-
         System.out.println(daprHttp);
+    }
 
+    @Test ()
+    public void testCallbackCalledAtTheExpectedTimeTest() throws IOException {
```
LMWF: Can you add a comment describing what this is validating and how?  Another way to look at this is "why this test is needed"?

Ideally unit tests self-documenting by name and from being short, but this one is more complicated.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -177,16 +179,48 @@ public void invokePostMethodUnknownError() throws IOException {
...
+        Mono<DaprHttp.Response> responseDeleteKey = daprHttp.invokeAPI("DELETE", urlDeleteState, null, null);
+        assertNull(serializer.deserialize(responseDeleteKey.block().getBody(), String.class));
+        mockInterceptor.reset();
+        mockInterceptor.addRule()
+            .get("http://localhost:3500/" +urlDeleteState)
+            .respond(404, ResponseBody.create(MediaType.parse("application/json"),
+                "{\"errorCode\":\"404\",\"message\":\"State Not Fuund\"}"));
+        try {
+            responseDeleted.block();
+            fail("Expected DaprException");
```
LMWF: Does this fail without the change in DaprClientGrpcAdapter above?

AndresRoblesMX: This test is completly isolated to the DaprHttp, not sure what are you asking regarding GRPC

LMWF: One of the changes in this PR is in DaprClientGrpcAdapter.java.  I thought this was validating that.  Sounds like it's not.  

Ok then is this testing the change in this PR in DaprHttp.java?  And does this test fail w/o that change?

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -132,7 +126,7 @@ public int getStatusCode() {
    * @return Asynchronous text
    */
   public Mono<Response> invokeAPI(String method, String urlString, Map<String, String> urlParameters, byte[] content, Map<String, String> headers) {
-    return Mono.fromFuture(CompletableFuture.supplyAsync(
```
artursouza: I've checked with Azure SDK team, let's keep the CompletableFuture (old code) but wrap it with Mono.defer(). This way, we still have a separate threadpool to do the http call and not block on the mono caller's thread.

AndresRoblesMX: using fromCallable will do the same

artursouza: But you will not have a separate threadpool, you would be using the caller's thread instead.

AndresRoblesMX: What's the point of having a separate thread, if we still need to wait for the thread to be completed?

artursouza: To show the benefits we would need to do performance testing. This change is better already because it is fixing a functionality bug, so I will take it. We might end up bringing back the threadpool but that will be separate.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
AndresRoblesMX: Test for the GRPC client has been added @LMWF can you take another look?

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
AndresRoblesMX: Documentation for the case has been added.

LMWF: Thanks for adding documentation.
"Block" in the comments here makes it sound like it is stopping the call from happening (e.g. blocking the network).  You could change the "Block" on 194, 198, 199 to "Block for" or "Wait for".

I'm asking to write down the scenario this method aims to test, because sometimes you see test cases that seem very specific or particular and the reason it was added is lost with time.  For example, it may not be clear to a reader why `responseDeleted` and `responseDeleteKey` are created in one order, then .block() invoked on them in the opposite order.  And why is the `addRule()` on line 227 added after resetting the interceptor?  These both must be to orchestrate a specific scenario - a description of what that is should go above the test method.

AndresRoblesMX: @LMWF can you look at the current documentation?

LMWF: Thanks for expanding.

Issue #86, opened 2020-01-13
Unit tests for stateful actors


artursouza:
Cover unit tests for the following scenarios for stateful actors:

Inheritance | Input Type | Output Type | State management (CRUD operations) | Exception in Actor's methods | Exception in Actor's runtime | Custom Serializer
-- | -- | -- | -- | -- | -- | --
No | String | String | Yes | No | No | No
No | String | String | Yes | No | Yes | No
No | String | String | Yes | Yes | No | No
No | Class | Class | Yes | No | No | No
No | Class | Class | Yes | No | No | Yes
Yes | int | int | Yes | No | No | No
Yes | int | int | Yes | No | Yes | No
Yes | int | int | Yes | Yes | No | No

Validations for all scenario (when applicable):
+ Output is as expected
+ Exception is as expected
+ State is saved as expected.
+ Methods are invoked in the Actor class: onActivate(), onDeactivate(), onPreActorMethod(), onPostActorMethod()
+ Timer is invoked()
+ Reminder is invoked()

PR #104, opened 2020-01-16
Unit tests for stateful actors + fixes + reformatting.

artursouza:
# Description

Unit tests for stateful actors + fixes + reformatting.

Sorry for bundling reformatting with other changes.

## Issue reference

Please reference the issue this PR will close: #86 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk-actors/src/test/java/io/dapr/actors/runtime/DaprInMemoryStateProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+import io.dapr.actors.ActorId;
+import io.dapr.client.DaprClient;
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class DaprInMemoryStateProvider extends DaprStateAsyncProvider {
```
amanbha: Is this added for tests?

LMWF: Yes, it is under `src/test`.

sdk-actors/src/test/java/io/dapr/actors/runtime/DaprInMemoryStateProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+  private static final Map<String, byte[]> stateStore = new HashMap<>();
+
+  private final ActorStateSerializer serializer;
+
+  DaprInMemoryStateProvider(ActorStateSerializer serializer) {
+    super(null, serializer);
+    this.serializer = serializer;
```
LMWF: Is this defined 2x?  It is in the parent as well.

artursouza: Yes. I did not want to change visibility of that property just for the tests. Sometimes I change but this time is avoidable.

sdk-actors/src/test/java/io/dapr/actors/runtime/ActorStatefulTest.java
```diff
@@ -0,0 +1,661 @@
...
+      eq(actorId.toString()),
+      any(),
+      any()))
+      .thenAnswer(invocationOnMock ->
+        this.manager.invokeMethod(
+          new ActorId(invocationOnMock.getArgument(1, String.class)),
+          invocationOnMock.getArgument(2, String.class),
+          context.getActorSerializer().unwrapMethodResponse(
+            invocationOnMock.getArgument(3, String.class), String.class))
+          .map(s -> {
```
LMWF: Can you intent (add 4 spaces) from .map downwards?  Otherwise it looks like it's lined up with unwrapMethodResponse() and acting on the result of that.

LMWF: 2 spaces rather.

sdk-actors/src/test/java/io/dapr/actors/runtime/DaprInMemoryStateProvider.java
LMWF: It's confusing on first read and might be better if it did not have the same name as the hidden parent member.

sdk-actors/src/test/java/io/dapr/actors/runtime/ActorStatefulTest.java
```diff
@@ -0,0 +1,661 @@
...
+    when(daprClient.invokeActorMethod(
+      eq(context.getActorTypeInformation().getName()),
+      eq(actorId.toString()),
+      any(),
+      any()))
+      .thenAnswer(invocationOnMock ->
+        this.manager.invokeMethod(
+          new ActorId(invocationOnMock.getArgument(1, String.class)),
+          invocationOnMock.getArgument(2, String.class),
+          context.getActorSerializer().unwrapMethodResponse(
```
LMWF: This call to `unwrapMethodResponse` and the call a few lines after it to `wrapMethodRequest` read strangely, because at first it looks like you're calling something used to unwrap the response but on the *input* on line 612.  Then the `wrapMethodRequest` on 616 makes it look like you're wrapping the output using a method normally to wrap the *input*.

It looks like it's really filling in some functionality that is needed around the mocking, but a comment would help, because it almost looks "reversed".

sdk-actors/src/test/java/io/dapr/actors/runtime/ActorStatefulTest.java
LMWF: Nevermind, it is formatted correctly, but there is just a lot packed into these surrounding lines.

sdk-actors/src/test/java/io/dapr/actors/runtime/ActorStatefulTest.java
artursouza: Ha. I plan to address this in the next PR when I will be adding custom serializer. This method should be called unwrapData() and the other one wrapData().

LMWF: That's fine.

artursouza: Renamed it in this PR.

sdk-actors/src/test/java/io/dapr/actors/runtime/DaprInMemoryStateProvider.java
artursouza: Passing null to super now.

sdk-actors/src/test/java/io/dapr/actors/runtime/DaprInMemoryStateProvider.java
artursouza: Added comments, just in case.

PR #107, opened 2020-01-16
save should not be called during actor deactivation

LMWF:
# Description

save() should not be called during actor deactivation.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java
```diff
@@ -256,8 +256,7 @@ ActorTimer getActorTimer(String timerName) {
 
         return this.resetState()
                 .then(this.onDeactivate())
-                .then(this.doWriteInfo(TRACE_TYPE, this.id.toString(), "Deactivated"))
-                .then(this.saveState());
+                .then(this.doWriteInfo(TRACE_TYPE, this.id.toString(), "Deactivated"));
```
artursouza: Can you add an unit test that fails without this fix?

LMWF: Functionality, it's the same with and without the fix other than some wasted calls because the first thing save does is no-op if the state change tracker map is empty, which it will be because resetState() was called.

However, removing the no-op save makes from the deactivate call makes it easier for readers to conceptually understand what deactivate is supposed to do.

LMWF:
Will recreate.

Issue #90, opened 2020-01-13
[java-sdk] define release process and set up CD to Nexus repo


youngbupark:
This is to set up cd pipeline to deploy java sdk to GitHub Package for master branch.

youngbupark:
@artursouza @amanbha @shalabhms  @LMWF 

* Release process: https://github.com/dapr/java-sdk/wiki/Release-Process
* Cert Rotation: https://github.com/dapr/java-sdk/wiki/Manage-OSSRH

I'm closing this issue.

PR #113, opened 2020-01-17
[CD] Fix gpg private key import

youngbupark:
# Description

* Fix gpg private key import in CI
* Remove useless properties

## Issue reference

#90 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


sdk-actors/pom.xml
```diff
@@ -15,6 +15,7 @@
   <version>0.2.0-SNAPSHOT</version>
   <name>dapr-sdk</name>
```
artursouza: Take this as a chance to fix this too.

youngbupark: fixed.

PR #114, opened 2020-01-17
Removing springboot jar and adding some unit tests for ActorRuntime.

artursouza:
# Description

Removing springboot jar and adding some unit tests for ActorRuntime.

## Issue reference

Please reference the issue this PR will close: #44 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


examples/pom.xml
```diff
@@ -84,7 +84,7 @@
     </dependency>
     <dependency>
       <groupId>io.dapr</groupId>
-      <artifactId>dapr-sdk-springboot</artifactId>
+      <artifactId>dapr-sdk-actors</artifactId>
```
LMWF: Is this correct is it supposed to be -examples?

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorStateSerializer.java
```diff
@@ -12,188 +12,199 @@
...
+  private String serialize(ActorRuntimeConfig config) throws IOException {
+    try (Writer writer = new StringWriter()) {
+      JsonGenerator generator = JSON_FACTORY.createGenerator(writer);
+      generator.writeStartObject();
+      generator.writeArrayFieldStart("entities");
+      for (String actorClass : config.getRegisteredActorTypes()) {
+        generator.writeString(actorClass);
+      }
+      generator.writeEndArray();
+      // TODO: handle configuration.
```
LMWF: Still todo?

sdk/src/main/java/io/dapr/runtime/Dapr.java
```diff
@@ -58,6 +60,14 @@ public static DaprRuntime getInstance() {
     return instance;
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public String serializeSubscribedTopicList() throws IOException {
+    return new String(this.serializer.serialize(this.getSubscribedTopics()), StandardCharsets.UTF_8);
```
LMWF: Is the last arg needed?  That is the java default.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorStateSerializer.java
artursouza: Yes.

examples/pom.xml
artursouza: This is for dependency, not the name of this jar.

sdk/src/main/java/io/dapr/runtime/Dapr.java
artursouza: It is not needed, but it is good to specify which Charset you need, otherwise (from Java docs):

"Constructs a new String by decoding the specified array of bytes using the platform's default charset."

PR #122, opened 2020-01-18
Add plugin for integration tests

LMWF:
mvn integration-test doesn't seem to work.  Adding this to the pom allows it to function as it should - mvn test will run unit tests and mvn integration-test will run integration tests.

The latter do not pass though - there are a lot of strange errors.  

* [x] Code compiles correctly
* [ ] Created/updated tests


artursouza:
Can you disable the IT tests in code? Use the @Ignore annotation:

https://github.com/junit-team/junit4/wiki/Ignoring-tests

Issue #123, opened 2020-01-18
Use byte[] for customer data and serialization/deserialization


artursouza:
In some places of the code we refer to Object, byte[] or String, including in the deserialization. Be consistent and always use byte[].

--Feedback from design review

PR #124, opened 2020-01-18
Addresses multiple items from design review

artursouza:
# Description

Use byte[] for customer data and serialized data.
Hide actor related APIs from DaprClient.
Simplify Serializer.
Remove Actor interface.
Supports custom serializer.

## Issue reference

Please reference the issue this PR will close: #123 #125 #126 #127 #128 #129 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


examples/src/main/java/io/dapr/examples/bindings/http/OutputBindingExample.java
```diff
@@ -25,7 +24,7 @@ public MyClass(){}
   }
 
   public static void main(String[] args) throws Exception {
-    DaprClient client = new DaprClientBuilder().build();
+    DaprClient client = new DaprClientBuilder(null).build();
```
LMWF: This seems like it would be more natural if there was an overload that took in no args to represent this.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
```diff
@@ -291,13 +325,18 @@ private Object invokeMethod(AbstractActor actor, Method method, Object request)
             this.runtimeContext.getActorTypeInformation().getName()));
       }
 
-      return actor.onPreActorMethodInternal(context)
-        .then(func.apply(actor))
-        .flatMap(result -> actor.onPostActorMethodInternal(context).thenReturn(result))
+      Mono<Object> result = actor.onPreActorMethodInternal(context)
+        .then((Mono<Object>)func.apply(actor));
```
LMWF: By separating the chain rollback() is no longer called now if there's a failure.  Why change it from what it was previously?

LMWF: Nevermind, I see now.  I still think it should be one chain of .'s though.

examples/src/main/java/io/dapr/examples/bindings/http/OutputBindingExample.java
artursouza: I agree. On the other hand, the feedback we got is to make it a required argument, so users must pass in something and be aware that they are OK with the default behavior.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
artursouza: OK. I was debugging. I will put these together.

examples/src/main/java/io/dapr/examples/bindings/http/OutputBindingExample.java
AndresRoblesMX: Passing null into the constructor of a builder makes the attribute not required, having a withSerializer method to add the serializer will be more intuitive for users of the API, regarding what´s expected, By defualt a Builders builds base on default behaviors, only if you want to change something you send it before building, but only Required attributes that do not have a defuault behavior are passes in the constructor.

artursouza: I will make it required and handle null as exception.

Issue #132, opened 2020-01-21
Have separate serializer for state store


artursouza:
Today, the code is supporting a single custom serializer for both transient and persisted objects. The goal is to allow customers to provide two different serializers instead: one for transient objects and another one for persisted objects.

At this point, the same serializer interface can be used for both.

PR #135, opened 2020-01-21
Separate serializer for state + handle String-based serializer.

artursouza:
# Description

Separate serializer for state.
Handle String-based serializer.

## Issue reference

Please reference the issue this PR will close: #132 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -112,29 +112,37 @@ public static ActorRuntime getInstance() {
...
    * @param actorFactory     An optional factory to create actors. This can be used for dependency injection.
-   * @param serializer       Serializer for Actor's state and transient objects.
+   * @param serializer       Serializer for Actor's transient objects.
+   * @param stateSerializer  Serializer for Actor's state objects.
    * @param <T>              Actor class type.
    */
   public <T extends AbstractActor> void registerActor(
-    Class<T> clazz, ActorFactory<T> actorFactory, DaprObjectSerializer serializer) {
+    Class<T> clazz, ActorFactory<T> actorFactory,
+    DaprObjectSerializer serializer,
```
youngbupark: Nit: can we rename `serializer` to be more specific name.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -112,29 +112,37 @@ public static ActorRuntime getInstance() {
...
    * @param clazz            The type of actor.
-   * @param objectSerializer Serializer for Actor's state and transient objects.
+   * @param objectSerializer Serializer for Actor's transient objects.
+   * @param stateSerializer  Serializer for Actor's state objects.
    * @param <T>              Actor class type.
    */
-  public <T extends AbstractActor> void registerActor(Class<T> clazz, DaprObjectSerializer objectSerializer) {
-    registerActor(clazz, null, objectSerializer);
+  public <T extends AbstractActor> void registerActor(
+    Class<T> clazz, DaprObjectSerializer objectSerializer, DaprObjectSerializer stateSerializer) {
```
youngbupark: I know why we let users to pass these serializers, but as a developer, I do not like to set serializer explicitly all the time :) this is my personal opinion.

examples/src/main/java/io/dapr/examples/actors/http/DemoActorService.java
```diff
@@ -36,7 +36,8 @@ public static void main(String[] args) throws Exception {
     int port = Integer.parseInt(cmd.getOptionValue("port"));
 
     // Register the Actor class.
-    ActorRuntime.getInstance().registerActor(DemoActorImpl.class, new DefaultObjectSerializer());
+    ActorRuntime.getInstance().registerActor(
+      DemoActorImpl.class, new DefaultObjectSerializer(), new DefaultObjectSerializer());
```
LMWF: Since the examples are using the default, many uses will simply copy and past what's there.  There should be an obvious comment both in the examples *and* in docs that discusses the tradeoffs.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntimeContext.java
```diff
@@ -21,7 +21,7 @@
   private final ActorRuntime actorRuntime;
 
   /**
-   * Serializer.
+   * Serializer for transient objects.
```
LMWF: I don't think this term has an obvious meaning and can be easily misinterpreted.

examples/src/main/java/io/dapr/examples/actors/http/DemoActorService.java
artursouza: I agree. Please, look at the future PR for docs as well.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: We had a strong feedback on this. On the other hand, we can quickly add overloaded methods hiding this based on user feedback.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: I am not sure what a better name would be. Let me find something and change it.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntimeContext.java
artursouza: Let me think about something else.

artursouza: See the latest version, please.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
artursouza: See the latest version, please.

sdk/src/main/java/io/dapr/client/DaprClientGrpcAdapter.java
```diff
@@ -40,19 +38,30 @@
...
+   * @param serializer      Serializer for transient objects.
+   * @param stateSerializer Serializer for state objects.
+   * @see DaprClientBuilder
    */
-  DaprClientGrpcAdapter(DaprGrpc.DaprFutureStub futureClient, DaprObjectSerializer serializer) {
-    client = futureClient;
-    objectSerializer = serializer;
+  DaprClientGrpcAdapter(
+    DaprGrpc.DaprFutureStub futureClient,
+    DaprObjectSerializer serializer,
```
youngbupark: I would rename `serializer` to `objectSerializer` here.

artursouza: It is internal class but renamed it too. Please, approve again.

PR #138, opened 2020-01-22
Add gRPC State Integration Test

JuanJose-Herrera:
# Description

Add GRP State Integration Test

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
```diff
@@ -85,6 +85,10 @@ public DaprFreePorts initializeDapr() throws Exception {
   private String buildDaprCommand(){
     StringBuilder stringBuilder= new StringBuilder(String.format(DAPR_RUN, this.appName))
       .append(this.useAppPort ? "--app-port " + this.DAPR_FREEPORTS.appPort : "")
+            .append(" --grpc-port ")
+            .append(this.DAPR_FREEPORTS.grpcPort)
+            .append(" --port ")
+            .append(this.DAPR_FREEPORTS.httpPort)
```
youngbupark: Do we need this ? dapr cli finds the free ports for http and grpc and set the envrionment variables by itself. 

@LMWF

sdk/src/test/java/io/dapr/it/BaseIT.java
```diff
@@ -27,6 +27,7 @@
   @BeforeClass
   public static void setEnvironmentVariables(){
     environmentVariables.set("DAPR_HTTP_PORT", String.valueOf(DAPR_FREEPORTS.getHttpPort()));
+    environmentVariables.set("DAPR_GRPC_PORT", String.valueOf(DAPR_FREEPORTS.getGrpcPort()));
```
youngbupark: this should be set by dapr cli.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
JuanJose-Herrera: yes, but IT does not aware of that, does not read the environment variables

sdk/src/test/java/io/dapr/it/BaseIT.java
JuanJose-Herrera: no, this is a Mock of the environment variables, is not the actual ones, because, at this moment, IT test does not know about that variables, and actually there is another problem where the client is a singleton, so even if dapr change the port, the client is not aware of that.

artursouza: This code runs outside dapper, so it is needed.

sdk/src/test/java/io/dapr/it/state/GRPCStateClientIT.java
```diff
@@ -0,0 +1,546 @@
...
+
+    //Create deferred action to get the sate without an etag
+    response = daprClient.getState(new State(stateKey, null, null), MyData.class);
+    myDataResponse = response.block();
+
+    //Review that the response is null, because the state was deleted
+    assertNull(myDataResponse.getValue());
+  }
+
+  @Ignore
```
artursouza: Explain why.

sdk/src/test/java/io/dapr/it/DaprIntegrationTestingRunner.java
artursouza: JuanJose is correct. We need to know the ports outside Dapr too.

LMWF: If the IT needs to know what they are then this is fine.  Otherwise it would be preferable if dapr decides.

Note having hyphens in the command line here is only a problem if they're inside the -Dexec.args portion of the mvn command.  Here they are not.  If they were being used to populate DAPR_COMMAND (the immediate next line), they would and that would be a problem.

Issue #128, opened 2020-01-20
Simplify Serializer


artursouza:
Change default serializer to be straight JSON and not have a custom logic to handle primitive types.

PR #139, opened 2020-01-22
Fix bug in Actor due to timer serialization.

artursouza:
# Description

Fix bug in Actor due to timer serialization.

## Issue reference

Please reference the issue this PR will close: #128 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java
@@ -163,7 +163,7 @@ public abstract class AbstractActor {
           this.actorRuntimeContext.getActorTypeInformation().getName(),
           this.id.toString(),
           actorTimer.getName(),
-          this.actorRuntimeContext.getObjectSerializer().serialize(actorTimer));
+          INTERNAL_SERIALIZER.serialize(actorTimer));
       } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java
```diff
@@ -163,7 +163,7 @@ protected ActorStateManager getActorStateManager() {
           this.actorRuntimeContext.getActorTypeInformation().getName(),
           this.id.toString(),
           actorTimer.getName(),
-          this.actorRuntimeContext.getObjectSerializer().serialize(actorTimer));
+          INTERNAL_SERIALIZER.serialize(actorTimer));
```
LMWF: That's pointing to io.dapr.client.ObjectSerializer.  Don't you want io.dapr.runtime.ObjectSerializer?  That is the one with the custom method for Timers.

Outside scope for this PR, but parent and child shouldn't have the same name, it's error prone.

LMWF: Ok nevermind, it is actually pointing at io.dapr.runtime's version, so it's good.

But like I said, having the name ObjectSerializer for 2 classes is confusing.

Issue #72, opened 2020-01-08
Review all 3rd party dependencies


artursouza:
## Expected Behavior

Dapr's java-sdk has 3rd party dependencies. 

## Actual Behavior

Move them to an approved version/alternative when possible and identify exceptions.

Approved libraries should be in one of the references below:
https://azure.github.io/azure-sdk/java_implementation.html#dependencies
https://github.com/Azure/azure-sdk-for-java
https://github.com/Azure/azure-cosmosdb-java
(any other official Azure SDK)

## Steps to Reproduce the Problem

Look at all our direct dependencies.

artursouza:
$ mvn dependency:list -DexcludeTransitive | egrep "compile|test|provided|runtime" | sort | uniq
[INFO]    com.fasterxml.jackson.core:jackson-databind:jar:2.9.10.1:compile
State: Old version.
Action: Upgrade to 2.10.1 as per Azure's Java SDK.

[INFO]    com.github.gmazzo:okhttp-mock:jar:1.3.2:test
[INFO]    com.github.gmazzo:okhttp-mock:jar:1.3.2:test
State: Not in Azure's SDK. MIT License
URL: https://github.com/gmazzo/okhttp-client-mock/blob/master/LICENSE
Action: Keep it since it is only for test and not shipped.

[INFO]    com.github.os72:protoc-jar-maven-plugin:jar:3.10.1:compile
State: Not in Azure's SDK. Apache 2 license.
URL: https://github.com/os72/protoc-jar-maven-plugin/blob/master/LICENSE
Action: Keep it as a required exception for GRPC support.

[INFO]    com.google.protobuf:protobuf-java-util:jar:3.11.0:compile
State: Not in Azure's SDK. Custom License.
URL: https://github.com/google/protobuf-gradle-plugin/blob/master/LICENSE
Action: Remove it since we found it to be unused.

[INFO]    com.squareup.okhttp3:okhttp:jar:4.2.1:compile
State: In Azure's SDK as a transient dependency.

[INFO]    commons-cli:commons-cli:jar:1.4:compile
[INFO]    commons-cli:commons-cli:jar:1.4:test
State: Not in Azure's SDK, but they have other jars from Apache Commons.  Apache License V2.
URL: https://github.com/apache/commons-cli/blob/master/LICENSE.txt
Action: Keep it since it is useful to parse command line arguments in examples.

[INFO]    io.grpc:grpc-api:jar:1.25.0:compile
[...]

PR #164, opened 2020-01-29
Removing some dependencies after review.

artursouza:
# Description

Removing dependencies after review.

## Issue reference

Please reference the issue this PR will close: #72 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk-tests/pom.xml
+++ b/sdk-tests/pom.xml
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.12</version>
+      <version>4.13-beta-3</version>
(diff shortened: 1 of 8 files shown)


sdk-tests/pom.xml
```diff
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.12</version>
+      <version>4.13-beta-3</version>
```
youngbupark: nit: should we use beta version?

artursouza: Same version as in Azure SDK. I am fine since it is for test only.

Issue #100, opened 2020-01-14
Integration tests for method invocation


artursouza:
Validate method invocation in Java SDK.

AndresRoblesMX:
@artursouza is anything else needed for method invocation that what I did for the Activate/Deactivate Integration test?
https://github.com/dapr/java-sdk/pull/144

artursouza:
This is for non Actor API.

PR #194, opened 2020-02-01
Add Integration Test for Method Invoke

JuanJose-Herrera:
# Description

Add Integration Test for Method Invoke

## Issue reference

https://github.com/dapr/java-sdk/issues/100

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-tests/pom.xml
+++ b/sdk-tests/pom.xml
@@ -65,12 +65,6 @@
       <version>2.2.2.RELEASE</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.github.javafaker</groupId>
-      <artifactId>javafaker</artifactId>
-      <version>0.15</version>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 
   <build>
--- a/sdk-tests/pom.xml
+++ b/sdk-tests/pom.xml
@@ -65,6 +65,12 @@
       <version>2.2.2.RELEASE</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.javafaker</groupId>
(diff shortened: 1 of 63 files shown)


sdk-tests/pom.xml
```diff
@@ -65,6 +65,12 @@
       <version>2.2.2.RELEASE</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.javafaker</groupId>
```
artursouza: This seems to be a convenient package to generate data. We've finished a review of 3rd party dependencies. I do not see this one adding enough value to justify adding another 3rd party dependency.

sdk-tests/src/test/java/io/dapr/it/methodinvoke/http/MethodInvokeController.java
```diff
@@ -0,0 +1,76 @@
...
+    public void deleteMessages(@PathVariable Integer messageId){
+        messagesReceived.remove(messageId);
+    }
+
+    @GetMapping(path = "/messages")
+    public Map<Integer, String> getMessages() {
+        return messagesReceived;
+    }
+
+
```
artursouza: Excessive spacing.

sdk-tests/pom.xml
JuanJose-Herrera: That was removed

sdk-tests/src/test/java/io/dapr/it/methodinvoke/http/MethodInvokeController.java
JuanJose-Herrera: spaces removed

PR #205, opened 2020-02-05
Review changes of 0.2.0 beta

RicardoNiepel:
# Description

Added and changed small parts under examples
- added debugging instructions with VSCode sample config
- fixed some out-of-sync between readmes and code
- fixed some typos in readmes

### DIFF
--- a/examples/src/main/java/io/dapr/examples/actors/http/README.md
+++ b/examples/src/main/java/io/dapr/examples/actors/http/README.md
@@ -126,8 +126,14 @@ For VSCode you can find a sample launch.json which includes:
 ...
 ```
 
-Use the following command to run the Dapr sidecar:
+Use the following commands to run the Dapr sidecar.
 
+For Linux and MacOS:
+```sh
+dapr run --app-id demoactorservice --app-port 3000 --port 3005 --grpc-port 5001 -- cat
+```
+
+For Windows:
 ```sh
 dapr run --app-id demoactorservice --app-port 3000 --port 3005 --grpc-port 5001 -- waitfor FOREVER
 ```
--- a/examples/src/main/java/io/dapr/examples/actors/http/README.md
+++ b/examples/src/main/java/io/dapr/examples/actors/http/README.md
@@ -105,7 +105,7 @@ An actor inherits from `AbstractActor` and implements the constructor to pass th
 Now, execute the following script in order to run DemoActorService:
 ```sh
 cd to [repo-root]
-dapr run --app-id demoactorservice --app-port 3000 --port 3005 -- mvn exec:java -pl=examples -D exec.mainClass=io.dapr.examples.actors.http.DemoActorService -D exec.args="-p 3000"
+dapr run --app-id demoactorservice --app-port 3000 --port 3005 -- mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.actors.http.DemoActorService -Dexec.args="-p 3000"
 ```
 
 ### Debugging the Demo actor service
@@ -188,7 +188,7 @@ Use the follow command to execute the DemoActorClient:
 
 ```sh
 cd to [repo-root]
-dapr run --app-id demoactorclient --port 3006 -- mvn exec:java -pl=examples -D exec.mainClass=io.dapr.examples.actors.http.DemoActorClient
+dapr run --app-id demoactorclient --port 3006 -- mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.actors.http.DemoActorClient
 ```
 
 Once running, the `DemoActorClient` logs will start displaying the different steps: 
--- a/examples/src/main/java/io/dapr/examples/actors/http/README.md
+++ b/examples/src/main/java/io/dapr/examples/actors/http/README.md
@@ -55,7 +55,7 @@ public class DemoActorService {
 This application uses `ActorRuntime.getInstance().registerActor()` in order to register `DemoActorImpl` as an actor in the Dapr Actor runtime. Internally, it is using `DefaultObjectSerializer` for two properties: `objectSerializer` is for Dapr's sent and received objects, and `stateSerializer` is for objects to be persisted.
  
 
-`DaprApplication.start()` method will run the Spring Boot [DaprApplication](../../../springboot/DaprApplication.java), which registers the Dapr Spring Boot controller [DaprController](../../springboot/DaprController.java). This controller contains all Actor methods implemented as endpoints. The Dapr's sidecar will call into the controller.
+`DaprApplication.start()` method will run the Spring Boot [DaprApplication](../../../springboot/DaprApplication.java), which registers the Dapr Spring Boot controller [DaprController](../../../springboot/DaprController.java). This controller contains all Actor methods implemented as endpoints. The Dapr's sidecar will call into the controller.
 
 See [DemoActorImpl](DemoActorImpl.java) for details on the implementation of an actor:
 ```java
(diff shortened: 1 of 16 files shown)


examples/src/main/java/io/dapr/examples/pubsub/http/Subscriber.java
```diff
@@ -17,7 +17,7 @@
  * mvn clean install
  * 2. Run the server:
  * dapr run --app-id subscriber --app-port 3000 --port 3005 -- \
- *   mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.pubsub.http.Subscriber -Dexec.args="-p 3000"
+ *   mvn exec:java -pl=examples -Dexec.mainClass=io.dapr.examples.pubsub.http.Subscriber -D exec.args="-p 3000"
```
youngbupark: why do we need space here?

examples/src/main/java/io/dapr/examples/actors/http/README.md
```diff
@@ -108,6 +108,30 @@ cd to [repo-root]
...
+    "projectName": "dapr-sdk-examples",
+    "args": "-port=3000"
+},
+...
+```
+
+Use the following command to run the Dapr sidecar:
+
+```sh
+dapr run --app-id demoactorservice --app-port 3000 --port 3005 --grpc-port 5001 -- waitfor FOREVER
```
youngbupark: waitfor is not a builtin command in shell. users needs to install it separately.

examples/src/main/java/io/dapr/examples/pubsub/http/Subscriber.java
RicardoNiepel: It's not required - but at all readmes it's with a space - so my suggestion is to align it

examples/src/main/java/io/dapr/examples/actors/http/README.md
RicardoNiepel: I just realized after the PR that at each debugging doc we need to separate between Unix and Windows if we go with ```dapr``` and dummy commands.
Thus my suggestion is to change it to ```daprd``` - see #212

youngbupark: @RicardoNiepel for now, let's use `cat` for linux and mac. please add additional command in this doc. if we use daprd, then we may need to update the existing docs. Rather than this, I make dapr cli to run alone without app. 

Also Are you supposed to use vscode even for java?

examples/src/main/java/io/dapr/examples/pubsub/http/Subscriber.java
youngbupark: I think the other docs are wrong. we need to update the other readmes. I would remove space here.

examples/src/main/java/io/dapr/examples/actors/http/README.md
LMWF: We're using `waitfor` on Windows and `cat` on MacOS/Linux just to have a dummy command so the user can start their app separately through an IDE/debugger.  These two options are listed in the README.md in the root of the repo ie

```
 For Linux and MacOS:

dapr run --app-id testapp --app-port 3000 --port 3500 --grpc-port 5001 -- cat
For Windows:

dapr run --app-id testapp --app-port 3000 --port 3500 --grpc-port 5001 -- waitfor FOREVER
```


We used `cat` on MacOS/Linux but unfortunately it's not built-in on Windows, so we used `waitfor`, which is built into Win7 and later.

In places where the command differs between OSes, we've tried to list the command to run per OS.  It would be good to list the commands per OS like the text I pasted above.

examples/src/main/java/io/dapr/examples/actors/http/README.md
```diff
@@ -108,6 +108,30 @@ cd to [repo-root]
 dapr run --app-id demoactorservice --app-port 3000 --port 3005 -- mvn exec:java -pl=examples -D exec.mainClass=io.dapr.examples.actors.http.DemoActorService -D exec.args="-p 3000"
 ```
 
+### Debugging the Demo actor service
+
+If you want to debug the `Subscriber`, you have to make sure to provide the port as an argument.
```
LMWF: Subscriber is a confusing name here, it should say actor service.

examples/src/main/java/io/dapr/examples/bindings/http/README.md
```diff
@@ -92,6 +92,30 @@ cd to [repo-root]/examples
...
+    "projectName": "dapr-sdk-examples",
+    "args": "-port=3000"
+},
+...
+```
+
+Use the following command to run the Dapr sidecar:
+
+```sh
+dapr run --app-id inputbinding --app-port 3000 --port 3005 --grpc-port 5001 -- waitfor FOREVER
```
LMWF: It would be good to have MacOS/Linux instructions too.  Same as above, just `cat` instead of `waitfor FOREVER`

examples/src/main/java/io/dapr/examples/invoke/grpc/README.md
```diff
@@ -26,7 +26,7 @@ mvn install
 
 ### Running the example's service
 
-The first component is the service. It has a simple API with the `Say` method. This method will print out each message received from the client. The proto file below contains the description of the HelloWorld service found in the `./proto/examples/helloworld.proto` file:
+The first component is the service. It has a simple API with the `Say` method. This method will print out each message received from the client. The proto file below contains the description of the HelloWorld service found in the `./examples/proto/helloworld.proto` file:
```
LMWF: Lowercase "say"

examples/src/main/java/io/dapr/examples/invoke/grpc/README.md
```diff
@@ -103,7 +124,7 @@ private static class HelloWorldClient {
 }
 ```
 
-First, it creates an instance of `DaprClient` via `DaprClientBuilder`. The protocol used by DaprClient is transparent to the application. The HTTP and GRPC ports used by Dapr's sidecar are automatically chosen and exported as environment variables: `DAPR_HTTP_PORT` and `DAPR_GRPC_PORT`. Dapr's Java SDK references these environment variables when communicating to Dapr's sidecar.
+First, it creates an instance of `DaprClient` via `DaprClientBuilder`. The protocol used by DaprClient can be set with the JVM's system property `dapr.grpc.enabled` or environment variable `DAPR_GRPC_ENABLED`. If both are not set, the default is to use HTTP. The HTTP and GRPC ports used by Dapr's sidecar are automatically chosen and exported as environment variables: `DAPR_HTTP_PORT` and `DAPR_GRPC_PORT`. Dapr's Java SDK references these environment variables when communicating to Dapr's sidecar.
```
LMWF: Just to clarify - 

..."exported as environment variables" => "exported as environment variables to the app"

The app will start with these 2 environment variables, not daprd.

examples/src/main/java/io/dapr/examples/pubsub/http/Subscriber.java
LMWF: For some reason in the past this command did not work for some on Windows without the space in the past, but the issue doesn't appear to repro now.  This is something we talked about on the other call earlier today.

I think we can remove the space from all docs separately, this PR can keep the spaces.

examples/src/main/java/io/dapr/examples/actors/http/README.md
artursouza: There is significant logic in Dapr cli before it can invoke daprd with the right arguments.

examples/src/main/java/io/dapr/examples/pubsub/http/Subscriber.java
RicardoNiepel: I will clean it up

examples/src/main/java/io/dapr/examples/actors/http/README.md
RicardoNiepel: I will do this

examples/src/main/java/io/dapr/examples/actors/http/README.md
RicardoNiepel: Done

Issue #211, opened 2020-02-05
Unpredictable behaviour at the state management sample


RicardoNiepel:
## Environment
```powershell
PS C:\Sources\dapr-java-sdk> mvn -v
  Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
  Maven home: C:\Program Files\apache-maven-3.6.3\bin\..
  Java version: 11.0.6, vendor: AdoptOpenJDK, runtime: C:\Users\riniepel\AppData\Local\Programs\AdoptOpenJDK
  Default locale: de_DE, platform encoding: Cp1252
  OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
PS C:\Sources\dapr-java-sdk>dapr --version
  CLI version: 0.3.0
  Runtime version: 0.3.0
```

**Branch:** release-0.2.0-beta

## Expected Behavior

The [State management sample](https://github.com/dapr/java-sdk/tree/master/examples/src/main/java/io/dapr/examples/state) should not lead to Maven errors and warnings and each execution should bring the same successful outcome.

## Actual Behavior

As you can see below, multiple Maven Errors and Warnings are coming up.

In addition, sometimes the client is unable to connect to Dapr (the second run) and thus is not predictable.

### First Run

```
dapr run --port 3006 -- mvn exec:java -D exec.mainClass=io.dapr.examples.state.StateClient -D exec.args="'my message'"
Starting Dapr with id Pigwinter-Hide. HTTP Port: 3006. gRPC Port: 55991
You're up and running! Both Dapr and your app logs will appear here.

== DAPR == time="2020-02-05T14:01:58+01:00" level=info msg="starting Dapr Runtime -- version 0.3.0 -- commit 
v0.3.0-rc.0-1-gfe6c306-dirty"
== DAPR == time="2020-02-05T14:01:58+01:00" level=info msg="log level set to: info"
== DAPR == time="2020-02-05T14:01:58+01:00" level=info msg="standalone mode configured"
== DAPR == time="2020-02-05T14:01:58+01:00" level=info msg="dapr id: Pigwinter-Hide"
== DAPR == time="2020-02-05T14:01:58+01:00" level=info msg="loaded component messagebus (pubsub.redis)" 
== DAPR == time="2020-02-05T14:01:58+01:00" level=info msg="loaded component sample123 (bindings.kafka)"
[...]

Issue #209, opened 2020-02-05
Http DaprClient unable to connect to Dapr


RicardoNiepel:
## Environment
```powershell
PS C:\Sources\dapr-java-sdk> mvn -v
  Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
  Maven home: C:\Program Files\apache-maven-3.6.3\bin\..
  Java version: 11.0.6, vendor: AdoptOpenJDK, runtime: C:\Users\riniepel\AppData\Local\Programs\AdoptOpenJDK
  Default locale: de_DE, platform encoding: Cp1252
  OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
PS C:\Sources\dapr-java-sdk>dapr --version
  CLI version: 0.3.0
  Runtime version: 0.3.0
```

**Branch:** release-0.2.0-beta

## Expected Behavior

The [Method invocation HTTP sample](https://github.com/dapr/java-sdk/tree/master/examples/src/main/java/io/dapr/examples/invoke/http) should be possible to follow and start the ```DemoService``` and ```InvokeClient``` without any errors.

## Actual Behavior

Starting the ```InvokeClient``` brings the following error

```sh
The InvokeClient was unable to connect to Dapr:
== APP == java.lang.RuntimeException: java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:3006
== APP ==     at io.dapr.client.DaprHttp.lambda$invokeApi$3 (DaprHttp.java:214)
== APP ==     at reactor.core.publisher.MonoCallable.call (MonoCallable.java:91)
== APP ==     at reactor.core.publisher.FluxFlatMap.trySubscribeScalarMap (FluxFlatMap.java:126)
== APP ==     at reactor.core.publisher.MonoFlatMap.subscribeOrReturn (MonoFlatMap.java:53)
== APP ==     at reactor.core.publisher.Mono.subscribe (Mono.java:4090)
== APP ==     at reactor.core.publisher.Mono.block (Mono.java:1662)
== APP ==     at io.dapr.examples.invoke.http.InvokeClient.main (InvokeClient.java:34)
```

But I was able to send manually a POST successful to
	http://localhost:3006/v1.0/invoke/invokedemo/method/say
at the same time

## Steps to Reproduce the Problem

Start ```DemoService``` and ```InvokeClient```.

### DemoService
```
[...]

PR #213, opened 2020-02-05
Move from localhost to 127.0.0.1

yaron2:
This PR changes the address to talk to Dapr from localhost to 127.0.0.1.

This is the recommended way to avoid local OS specific DNS lookups that might hinder performance, and is a possible fix (needs validation) for https://github.com/dapr/java-sdk/issues/211 and https://github.com/dapr/java-sdk/issues/209.


youngbupark:
@artursouza @LMWF @yaron2 I will cherry-pick it to release-0.2.0-beta.

cc/ @lukekim @msfussell

artursouza:
Thanks. We can release beta-2 today.

Issue #178, opened 2020-01-30
Reduce verbosity for Grpc libraries in Java SDK


artursouza:
## Expected Behavior

When using GRPC (IT or examples), the output is verbose because it prints DEBUG

## Actual Behavior

Only print INFO and above.

## Steps to Reproduce the Problem

Run any example or IT that uses Grpc.

marcosreyes05:
I can review this one

artursouza:
Issue resolved now via reducing verbosity. Log level is now INFO for those classes.

PR #221, opened 2020-02-07
Reducing verbosity by setting log level to INFO in samples and tests

marcosreyes05:
# Description

Reducing verbosity in logs by configuring both Spring Boot's configuration properties and logback for recular java classes called via dapr

## Issue reference

Please reference the issue this PR will close: #178 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/examples/src/main/resources/application.properties
+++ b/examples/src/main/resources/application.properties
@@ -1,0 +1,1 @@
+logging.level.root=INFO
--- a/examples/src/main/resources/application.properties
+++ b/examples/src/main/resources/application.properties
@@ -1,1 +1,0 @@
-logging.level.root=INFO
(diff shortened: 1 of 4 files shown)


artursouza:
I deleted the properties files since it does not seem to make a difference. The xml config seems to be what is solving the issue. @marcosreyes05 Let me know if the files I removed are really needed before we merge this.

artursouza:
Merging. @marcosreyes05  Let me know if we really need the property files I deleted. In my tests for ITs and examples, everything worked fine without them.

marcosreyes05:
Hi @artursouza yes, it will work fine. I added them in case we needed to separate the log control between Spring boot based code and non-Spring code. But a single logback config would work as well

Issue #13, opened 2019-11-30
Steps to regenerate Protobuf classes


brunoborges:
## Describe the proposal
What were the steps followed to generate the protobuf classes?

artursouza:
I am new to this code base as well. I see you have a PR and it re-generated some protobuf code. Can you share some insights?

artursouza:
I have changed the code to not version control the auto-generated files and copies the proto files instead. A simple compile task will generate the protobuf files. The documentation should be updated to reflect that and mention that to update the proto files, there should be a copy from dapr/dapr into this repo (maybe also create a maven task to download this automatically).

brunoborges:
Using git submodules is not ideal as it requires constant update of the commit hash where the submodule refers to.

A task to download may be the best solution, unless the protobuf generator can consume remote URLs

artursouza:
I had discussed this with some folks. The Java proto files might be behind the master proto intentionally. The scenario is the one we have now where there is a breaking change in master branch but it is still not stable yet. We might need to download from a release branch to kind of guarantee an immutable proto file.

artursouza:
It should be possible to automatically download the file from a URL. Then we just need to point to a file within a given commit ref in the URL and we have an immutable file to download from and have consistent behavior of the build pipeline.

@JuanJose-Herrera is looking into this now.

artursouza:
We will merge the PR to automatically download the proto from dapr/dapr repo - we are just waiting for milestone 0.4 to be released this week.

PR #224, opened 2020-02-07
Download proto

artursouza:
# Description

Downloads proto files into target dir and generate code from there. It keeps a fixed reference and can easily be updated.

## Issue reference

Please reference the issue this PR will close: #13 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


artursouza:
Only merge after code freeze is over. This is not a needed fix to release 0.4.

pom.xml
```diff
@@ -17,6 +17,8 @@
     <grpc.version>1.25.0</grpc.version>
     <protobuf.version>3.11.0</protobuf.version>
     <protoc.version>3.10.0</protoc.version>
+    <dapr.proto.url>https://raw.githubusercontent.com/dapr/dapr/e540a7166aeaf115773ccc4c7a1056ae7eed073b/pkg/proto/dapr/dapr.proto</dapr.proto.url>
```
youngbupark: JFYI - This url is snapshot, which means it will not be a up-to-date file

artursouza: This is by design. We want to have a fixed version. So this build has predictable behavior.

youngbupark: How about the version tag instead of hash? e.g. https://raw.githubusercontent.com/dapr/cli/v0.3.0/cmd/dapr.go

artursouza: Yes, that is preferred but this ref is for the 0.4 milestone and we did not have one at that time.

artursouza: Updated to use v0.4.0

Issue #185, opened 2020-01-31
Test .net SDK with Java SDK for Actor methods


artursouza:
## Expected Behavior

1. Actor in .Net calls method on Actor in Java for String type.
2. Actor in .Net calls method on Actor in Java for Integer type.
3. Actor in .Net calls method on Actor in Java for Class type.
4. Actor in Java calls method on Actor in .Net for String type.
5. Actor in Java calls method on Actor in .Net for Integer type.
6. Actor in Java calls method on Actor in .Net for Class type.

LMWF:
Clarification - in the list above, the type indicated is the type for both input and output.

LMWF:
These work as expected other than for the case where the input is a string type.  That scenario encounters the known extra pair of quotes issue.

PR #225, opened 2020-02-07
Remove special case for String in serializer + serialize state value …

artursouza:
…"directly".

# Description

Remove special case for String in serializer.
Serialize state value as part of the request payload, so it does not have JSON inside string but JSON directly instead.

## Issue reference

Please reference the issue this PR will close: #185 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
```diff
@@ -177,22 +180,17 @@
...
-
-    if (raw.length == 0) {
-      return raw;
-    }
-
+  private <T> T deserializeDaprStateResponse(byte[] raw, Class<T> clazz) throws IOException {
     try {
-      return OBJECT_MAPPER.readValue(raw, String.class).getBytes();
+      // Try to deserialize after quoted string first.
+      return this.stateSerializer.deserialize(OBJECT_MAPPER.readValue(raw, String.class).getBytes(), clazz);
```
LMWF: How does this work?  Can you provide an example of what format it would have when this works and another for when it uses the other case?

artursouza: I just realized we don't need this anymore. This was needed because the server side was running with an older version. Actually, the data is much cleaner now.

artursouza: image

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -346,10 +345,14 @@
...
         byte[] data = this.stateSerializer.serialize(state.getValue());
-        if (this.isStateString) {
-          internalStateObjects.add(
-              new State<>(data == null ? null : new String(data), state.getKey(), state.getEtag(), state.getOptions()));
+        if (this.isDefaultStateSerializer) {
+          // If default serializer is being used, we just pass the object through to be serialized directly.
+          // This avoids a JSON object from being quoted inside a string.
+          // We WANT this: { "value" : { "myField" : 123 } }
+          // We DON't WANT this: { "value" : "{ \"myField\" : 123 }" }
+          internalStateObjects.add(state);
```
LMWF: This isn't referred to after this.

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
```diff
@@ -55,15 +55,15 @@
   DaprStateAsyncProvider(DaprClient daprClient, DaprObjectSerializer stateSerializer) {
     this.daprClient = daprClient;
     this.stateSerializer = stateSerializer;
-    this.isStateString = stateSerializer.getClass().getAnnotation(StringContentType.class) != null;
+    this.isDefaultStateSerializer = stateSerializer instanceof DefaultObjectSerializer;
```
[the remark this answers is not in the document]
artursouza: You are right.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
artursouza: Fixed.

Issue #248, opened 2020-03-08
Add GetSecret API to Java SDK


artursouza:
There is a new API: GetSecret. Work to be done:

1. Update URL to newest proto in pom.xml
2. Add API to GRPC and HTTP client.
3. Add unit testing and integration testing.

PR #249, opened 2020-03-10
Support for GetSecret + ITs + sample.

artursouza:
# Description

Add support for GetSecret API.
Integration tests for GetSecret API.
Sample for GetSecret API.

## Issue reference

Please reference the issue this PR will close: #248 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


.github/workflows/build.yml
```diff
@@ -24,8 +24,8 @@ jobs:
       JDK_VER: 13.0.x
       DAPR_RUNTIME_VER: 0.4.0-rc.1
       DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/f84566fb2bf5a599252ab9d6bd82fc78faf94dba/install/install.sh
-      DAPR_CLI_REF:
-      DAPR_REF:
+      DAPR_CLI_REF: eb76cbf2200058f849814649f23212b3247a952c
+      DAPR_REF: 034b82ebfbb7c76e5801ae6d90c2edd29f629df8
```
artursouza: These overrides are useful for integration tests to use an unreleased version via building from source. This feature was used in 0.4 release.

.github/workflows/build.yml
```diff
@@ -80,6 +80,12 @@ jobs:
       run: |
         docker-compose -f ./sdk-tests/deploy/local-test-kafka.yml up -d
         docker ps
+    - name: Install Local Hashicorp Vault using docker-compose
+      run: |
+        docker-compose -f ./sdk-tests/deploy/local-test-vault.yml up -d
+        docker ps
+    - name: Setup Vault's test token
+      run: echo myroot > /tmp/.hashicorp_vault_token
```
artursouza: This is a test only token. The same one is used in the docker compose file.

examples/.hashicorp_vault_token
```diff
@@ -0,0 +1 @@
+myroot
```
artursouza: This is the one for examples. I am adding this file here, so user does not have to create it while running the examples.

examples/components/hashicorp_vault.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: vault
+spec:
+  type: secretstores.hashicorp.vault
+  metadata:
+  - name: vaultAddr
+    value: "http://127.0.0.1:8200"
+  - name: skipVerify
+    value : true
+  - name: vaultTokenMountPath
+    value : ".hashicorp_vault_token"
```
artursouza: It expects this file to present at the same folder that the user runs `dapr run`, which is the file created above. Running this sample from this folder is documented as a required step (same as we have for bindings to have kafka setup).

pom.xml
```diff
@@ -17,8 +17,8 @@
     <grpc.version>1.25.0</grpc.version>
     <protobuf.version>3.11.0</protobuf.version>
     <protoc.version>3.10.0</protoc.version>
-    <dapr.proto.url>https://raw.githubusercontent.com/dapr/dapr/v0.4.0/pkg/proto/dapr/dapr.proto</dapr.proto.url>
-    <dapr.client.proto.url>https://raw.githubusercontent.com/dapr/dapr/v0.4.0/pkg/proto/daprclient/daprclient.proto</dapr.client.proto.url>
+    <dapr.proto.url>https://raw.githubusercontent.com/dapr/dapr/3b792de734594463f6a16ce8c952fe2ed96c0dc0/pkg/proto/dapr/dapr.proto</dapr.proto.url>
```
artursouza: To make sure we download the latest pre-release version. Not pointing to master because we want this to be immutable by design.

sdk-tests/components/vault.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: vault
+spec:
+  type: secretstores.hashicorp.vault
+  metadata:
+  - name: vaultAddr
+    value: "http://127.0.0.1:8200"
+  - name: skipVerify
+    value : true
+  - name: vaultTokenMountPath
+    value : "/tmp/.hashicorp_vault_token"
```
artursouza: For tests, we auto-generate this token file in the build steps.

sdk-tests/deploy/local-test-vault.yml
```diff
@@ -0,0 +1,10 @@
+version: '2'
+services:
+  hashicorp_vault:
+    image: vault
+    ports:
+      - "8200:8200"
+    cap_add:
+      - IPC_LOCK
+    environment:
+      VAULT_DEV_ROOT_TOKEN_ID: "myroot"
```
artursouza: Seems duplicate but this one is for tests. Sample might use another vault in the future, but we keep this one for tests. They are the same today.

sdk-tests/pom.xml
```diff
@@ -65,6 +65,12 @@
       <version>2.2.2.RELEASE</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.bettercloud</groupId>
+      <artifactId>vault-java-driver</artifactId>
```
artursouza: This is need for us to programmatically add secrets in integration tests.

sdk-tests/src/test/java/io/dapr/it/secrets/SecretsClientIT.java
```diff
@@ -0,0 +1,127 @@
...
+  private static Vault vault;
+
+  /**
+   * Parameters for this test.
+   * Param #1: useGrpc.
+   * @return Collection of parameter tuples.
+   */
+  @Parameterized.Parameters
+  public static Collection<Object[]> data() {
+    return Arrays.asList(new Object[][] { { false }, { true } });
```
artursouza: We run the same tests via HTTP and GRPC to keep the SDK working on both protocols.

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -221,7 +221,7 @@ public int getStatusCode() {
    * @return DaprError or null if could not parse.
    */
   private static DaprError parseDaprError(byte[] json) throws IOException {
-    if (json == null) {
+    if ((json == null) || (json.length == 0)) {
```
artursouza: Bugfix for issue found during unit testing.

sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
```diff
@@ -971,7 +1081,10 @@ public synchronized void onSuccess(@NullableDecl T result) {
     public synchronized void onFailure(Throwable throwable) {
       assertFalse(wasCalled);
       wasCalled = true;
-      assertEquals(failure, throwable);
+      assertEquals(failure == null, throwable == null);
```
artursouza: Better assertions to validate exception. Forcing the equals() was too strong for this check (IMO).

examples/src/main/java/io/dapr/examples/secrets/README.md
```diff
@@ -0,0 +1,112 @@
...
+This sample includes two files:
+
+* docker-compose-vault.yml (Starts Hashicorp's Vault as a container)
+* SecretClient.java (Reads a secret from Dapr's Secret Store)
+* Existing Dapr component file in `< repo dir >/examples/components/hashicorp_vault.yaml`
+* Existing token file in `< repo dir >/examples/.hashicorp_vault_token` (Consumed by `daprd`'s vault component above)
+
+Visit [this](https://github.com/dapr/docs/tree/master/concepts/secrets) link for more information about secret stores in Dapr.
+ 
+## Binding sample using the Java-SDK
```
LMWF: It looks like this should be about secret stores, not bindings.

examples/src/main/java/io/dapr/examples/secrets/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+Then build the Maven project:
+
+```sh
+# make sure you are in the `java-sdk` directory.
+mvn install
+```
+### Setting Vault locally
+
+Before getting into the application code, follow these steps in order to setup a local instance of Kafka. This is needed for the local instances. Steps are:
```
LMWF: Kafka -> Vault

sdk-tests/pom.xml
LMWF: I think this one should be added as an actual comment inside the xml.

artursouza: ok.

examples/src/main/java/io/dapr/examples/secrets/README.md
artursouza: Yes, copy and pasta.

examples/src/main/java/io/dapr/examples/secrets/README.md
artursouza: Same.

sdk-tests/pom.xml
artursouza: Done.

examples/src/main/java/io/dapr/examples/secrets/README.md
artursouza: Done.

examples/src/main/java/io/dapr/examples/secrets/README.md
artursouza: Done.

Issue #259, opened 2020-03-27
Add health check endpoint for actors


amanbha:
## Describe the proposal
Add healthz endpoint for actors as per Dapr runtime change https://github.com/dapr/dapr/pull/1313

artursouza:
Thanks. Will do.

PR #260, opened 2020-03-30
Health check API.

artursouza:
# Description

Adds health check API.
Adds Java Docs.

## Issue reference

Please reference the issue this PR will close: #259 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
+++ b/sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
@@ -16,6 +16,8 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RestController;
 import reactor.core.publisher.Mono;
 
+import java.io.IOException;
+
 /**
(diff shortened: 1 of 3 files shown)


sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
```diff
@@ -27,28 +29,65 @@
    */
   private static final DefaultObjectSerializer SERIALIZER = new DefaultObjectSerializer();
 
+  /**
+   * Callback API for health checks from Dapr's sidecar.
+   */
+  @GetMapping("/healthz")
+  public void healthz() {
```
amanbha: can you help me understand how and where does this results in a  200 OK response?

artursouza: It will handle a get request on /healthz, do nothing and return nothing. So, it is a successful request for Springboot = 200.

Issue #281, opened 2020-05-11
Provide Spring Boot starter


JMayrbaeurl:
## Describe the proposal
Provide a Spring Boot starter that adds a fully configured DaprClient instance to the application context by just putting the starter artifact in the dependency list of Maven (pom.xml file)

artursouza:
I saw the PR, can this be done without adding all the new jars? Can this be done in the existing sdk-springboot package?

JMayrbaeurl:
No. Because the basic idea is to rename the existing 'sdk-springboot' package to 'sdk-spring' for generic Spring support and introduce a new one for Spring Boot support. For the implementation I've followed the guidelines from the Spring. And I think it's much better to have the samples as close as possible to related sdk.

artursouza:
Your idea is fundamentally different from our vision for the SDK for the reasons below:

1. We minimize the dependency on Spring and Springboot and its footprint to keep the SDK as neutral as possible.

2. The samples are organized per SDK, fragmenting it further would make this SDK different. We need to review this proposal.

Long story short, to make these changes you will need to prepare a casa and review with the team. We had multiple sessions to review the proposals to reach the current form.

JMayrbaeurl:
A lot of big enterprises are using Spring Boot for their Cloud native developments and I thought it's a great idea to provide really great support in Dapr for Spring Boot. Currently Dapr has no Spring Boot support. Just a sample showing how to use it in Spring Boot. The basic Spring support is minimal, too.

And Microsoft has recently invested a lot in supporting Spring and Spring Boot on Azure. So I was expecting the same kind of support from Dapr.

JMayrbaeurl:
Moved my code in a new repo at https://github.com/JMayrbaeurl/dapr-spring-extensions

Issue #279, opened 2020-05-07
Samples really working? - Build with Java 11, rest is build with Java 8


JMayrbaeurl:
## Expected Behavior

Running Method invocation sample as described [here](https://github.com/dapr/java-sdk/tree/master/examples/src/main/java/io/dapr/examples/invoke/http)

## Actual Behavior

DemoService crashes on start with UnsupportedClassVersionError exception:
`dapr run --app-id invokedemo --app-port 3004 --port 3005 -- java -jar examples/target/dapr-java-sdk-examples-exec.jar io.dapr.examples.invoke.http.DemoService -p 3004
Starting Dapr with id invokedemo. HTTP Port: 3005. gRPC Port: 56180
←[94;1m== APP == Exception in thread "main" java.lang.UnsupportedClassVersionError: io/dapr/springboot/DaprMainApplication has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
←[0m
←[94;1m== APP ==        at java.lang.ClassLoader.defineClass1(Native Method)
←[0m
←[94;1m== APP ==        at java.lang.ClassLoader.defineClass(Unknown Source)
←[0m
←[94;1m== APP ==        at java.security.SecureClassLoader.defineClass(Unknown Source)
←[0m
←[94;1m== APP ==        at java.net.URLClassLoader.defineClass(Unknown Source)
←[0m
←[94;1m== APP ==        at java.net.URLClassLoader.access$100(Unknown Source)
←[0m
←[94;1m== APP ==        at java.net.URLClassLoader$1.run(Unknown Source)
←[0m
←[94;1m== APP ==        at java.net.URLClassLoader$1.run(Unknown Source)
←[0m
←[94;1m== APP ==        at java.security.AccessController.doPrivileged(Native Method)
←[0m
←[94;1m== APP ==        at java.net.URLClassLoader.findClass(Unknown Source)
←[0m
←[94;1m== APP ==        at java.lang.ClassLoader.loadClass(Unknown Source)
←[0m
←[94;1m== APP ==        at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:92)
←[0m
←[94;1m== APP ==        at java.lang.ClassLoader.loadClass(Unknown Source)
←[0m
←[94;1m== APP ==        at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:46)
←[0m
[...]

PR #282, opened 2020-05-11
Build fix for examples not working and Spring Boot starter for DaprClient

JMayrbaeurl:
# Description

Fixed problem with examples. They were built with Java 11, but need to be build with Java 8 as the rest of the code to work correctly. And added a Spring Boot starter for a DaprClient instance (see [Spring Boot docu for starters](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using-boot-starter) )

## Issue reference

#279 Fix for examples not working
#281 Spring Boot starter

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


pom.xml
```diff
@@ -253,6 +253,7 @@
     <module>sdk-actors</module>
     <module>sdk-springboot</module>
     <module>examples</module>
+    <module>dapr-spring-boot</module>
```
artursouza, line 256: Why are you creating all these new packages? We have purposely minimized our footprint on Spring boot. Before we can review this PR, we need a justification for the added support. We can hop into a Teams call for you to present the added benefits.

JMayrbaeurl: Devs using Spring Boot are used to get Spring Boot starters to integrate other libs or frameworks. This was the basic idea. Just providing a starter for the DaprClient instance creation that can be used with autowiring afterwards in the source code. I've added a sample just showing this with the name 'dapr-spring-boot-samples-autostarter' in the samples folder of 'dapr-spring-boot-samples'. And this is fully optional. 

The Maven modules and resulting packages approach just follows the guidelines of the official Spring Boot documentation for the development of starters. Especially for samples I dont think your current approach to provide them all in one folder called 'examples', resulting in one jar package, is a good idea, because it completly screws up the dependencies and doesnt alllow for real application context setups, that are normally necessary in Spring based application. I definitely recommend to either put the examples in another repo or to follow the folder structure approach I've shown in 'dapr-spring-boot-samples', using one common parent pom but having an extra package/module/folder for each example.

I'm in the CEST time zone and ready to jump on a Teams call to discuss my proposed changes anytime, e.g. today 10 am Redmond time.

JMayrbaeurl:
Closed because of too many different changes, some that need more discussion. I will split it up to several PRs

Issue #288, opened 2020-05-29
Update Java SDK for latest proto for 0.8 and bidirectional binding


artursouza:
## Expected Behavior

Java SDK should work with the master branch planned for release 0.8.

## Actual Behavior

Latest proto are not backwards compatible with this SDK.
Binding's API is not backwards compatible with this SDK.

## Steps to Reproduce the Problem

Try to use this SDK with the latest master from dapr-cli and dapr.

youngbupark:
@artursouza @mukundansundar are you working on this?

artursouza:
I am working on it today.

artursouza:
Please, take a look at the PR above.

PR #289, opened 2020-05-31
Bi-directional output binding + 'data' to refer to PubSub event

artursouza:
# Description

Bi-directional output binding.
Use 'data' to refer to PubSub event

## Issue reference

Please reference the issue this PR will close: #288 and https://github.com/dapr/dapr/issues/1370

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/examples/src/main/java/io/dapr/examples/bindings/http/OutputBindingExample.java
+++ b/examples/src/main/java/io/dapr/examples/bindings/http/OutputBindingExample.java
@@ -28,6 +28,8 @@ public class OutputBindingExample {
 
   static final String BINDING_NAME = "sample123";
 
+  static final String BINDING_OPERATION = "create";
+
   /**
    * The main method of this app.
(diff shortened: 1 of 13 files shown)


examples/src/main/java/io/dapr/examples/bindings/http/OutputBindingExample.java
```diff
@@ -28,6 +28,8 @@ public MyClass() {
 
   static final String BINDING_NAME = "sample123";
 
+  static final String BINDING_OPERATION = "create";
```
youngbupark, line 31: nit: I would use `BINDING_CREATE_OPERATION` or something more more specific name.

artursouza: I am following the same pattern as above.

Issue #296, opened 2020-06-29
Wrong content type on dapr/config API from SDK springboot


artursouza:
## Expected Behavior

dap/config should contain content-type header of application/json

## Actual Behavior

dap/config contains content-type header of application/octet-stream

## Steps to Reproduce the Problem

Run any app that has PubSub and look at the logs from sidecar.

javageek79:
"wrong" info in log message
```
{
	"id": "AQAAAXLr-Am4h2JXKQAAAABBWExyLUE3VFdBR1FDZTJ3TmNBQg",
	"content": {
		"timestamp": "2020-06-25T14:53:20.696Z",
		"tags": [
		],
		"host": "OUR_HOST",
		"service": "daprd",
		"message": "time=\"2020-06-25T14:53:20.663898976Z\" level=debug msg=\"dapr/config returns invalid content_type: application/octet-stream\" app_id=group-app instance=group-84898cb7b5-lpnwr scope=dapr.runtime type=log ver=0.8.0",
		"attributes": {
			"ver": "0.8.0",
			"level": "debug",
			"msg": "dapr/config returns invalid content_type: application/octet-stream",
			"instance": "group-84898cb7b5-lpnwr",
			"scope": "dapr.runtime",
			"time": "2020-06-25T14:53:20.663898976Z",
			"type": "log",
			"app_id": "group-app",
			"status": "debug"
		}
	}
}
```

PR #300, opened 2020-07-08
Fixing content-type for Spring controller.

artursouza:
# Description

Fixing content-type for Spring controller.

## Issue reference

Please reference the issue this PR will close: #296 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
+++ b/sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
@@ -7,6 +7,7 @@ package io.dapr.springboot;
 
 import io.dapr.actors.runtime.ActorRuntime;
 import io.dapr.serializer.DefaultObjectSerializer;
+import org.springframework.http.MediaType;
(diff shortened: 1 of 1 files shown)


sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
```diff
@@ -76,7 +77,7 @@ public void healthz() {
    * @param body Raw request body.
    * @return Raw response body.
    */
-  @PutMapping(path = "/actors/{type}/{id}/method/{method}")
+  @PutMapping(path = "/actors/{type}/{id}/method/{method}", produces = MediaType.APPLICATION_JSON_VALUE)
```
mukundansundar, line 80: is this always going to json and not any other type?

artursouza: I double checked and this could be something other than JSON. Let me revert this one.

Issue #242, opened 2020-02-20
Change APIs to support HTTPExtension


artursouza:
Change the ServiceInvocation APIs to accept a HTTPExtension argument which contains specific details needed for HTTP calls. This argument will not be needed for GRPC. Similar to [.NET SDK](https://github.com/dapr/dotnet-sdk/blob/master/src/Dapr.Client/DaprClient.cs#L79). 

Update documentation, samples and IT for the same.

PR #312, opened 2020-07-27
Change Service Invocation API. Fix metadata bug.

mukundansundar:
# Description

Change Service Invocation API to accept HttpExtension object.

Fix the bug where metadata passed in is used as urlParameters, instead use queryString from HttpExtension object.

## Issue reference

Please reference the issue this PR will close: #242 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprClient.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClient.java
@@ -46,7 +46,8 @@ public interface DaprClient {
    * @param appId         The Application ID where the service is.
    * @param method        The actual Method to be call in the application.
    * @param request       The request to be sent to invoke the service, use byte[] to skip serialization.
-   * @param httpExtension Additional fields that are needed if the receiving app is listening on HTTP.
+   * @param httpExtension Additional fields that are needed if the receiving app is listening on
+   *                      HTTP, {@link HttpExtension#NONE} otherwise.
    * @param metadata      Metadata (in GRPC) or headers (in HTTP) to be sent in request.
    * @param type          The Type needed as return for the call.
    * @param <T>           The Type of the return, use byte[] to skip serialization.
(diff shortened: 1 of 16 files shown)


sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -43,130 +43,134 @@
...
-   * @param method   The actual Method to be call in the application.
-   * @param request  The request to be sent to invoke the service, use byte[] to skip serialization.
-   * @param metadata Metadata (in GRPC) or headers (in HTTP) to be send in request.
-   * @param type     The Type needed as return for the call.
-   * @param <T>      The Type of the return, use byte[] to skip serialization.
-   * @return A Mono Plan of type type .
+   * @param appId         The Application ID where the service is.
+   * @param method        The actual Method to be call in the application.
+   * @param request       The request to be sent to invoke the service, use byte[] to skip serialization.
+   * @param httpExtension Additional fields that are needed if the receiving app is listening on HTTP.
```
mukundansundar, line 49: will update doc to read that should be `NONE` otherwise

sdk/src/main/java/io/dapr/client/HttpExtension.java
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * HTTP querystring.
+   */
+  private Map<String, String> queryString;
+
+  /**
+   * Convenience HttpExtension object for the NONE Verb with empty queryStreing.
+   */
+  public static final HttpExtension NONE = new HttpExtension(DaprHttp.HttpMethods.NONE, new HashMap<>());
```
artursouza, line 32: nit: move the static final properties to the top.

sdk/src/main/java/io/dapr/client/HttpExtension.java
```diff
@@ -0,0 +1,97 @@
...
+  public static final HttpExtension OPTIONS = new HttpExtension(DaprHttp.HttpMethods.OPTIONS, new HashMap<>());
+  /**
+   * Convenience HttpExtension object for the TRACE Verb with empty queryStreing.
+   */
+  public static final HttpExtension TRACE = new HttpExtension(DaprHttp.HttpMethods.TRACE, new HashMap<>());
+
+  /**
+   * Default no-op constructor.
+   */
+  private HttpExtension() {
```
artursouza, line 69: Why do you need this?

sdk/src/main/java/io/dapr/utils/Constants.java
```diff
@@ -15,6 +15,11 @@
    */
   public static final String API_VERSION = "v1.0";
 
+  /**
+   * Dapr's default protocol.
+   */
+  public static final String DEFAULT_PROTOCOL = "http";
```
artursouza, line 21: No, it's GRPC. Where is this being used? There is a Property to use GRPC by default and this conflicts with that.

mukundansundar: URL Builder scheme. Will rename to `DEFAULT_HTTP_SCHEME` ?

artursouza: If this is used only in the Http client, keep it as a private constant there, not here.

PR #321, opened 2020-08-17
Remove EmptyService and use dapr run -- instead.

artursouza:
# Description

Since now dapr cli allows running sidecar only (no app), we don't need EmptyService anymore. Removing it and using `dapr run` directly instead.

## Issue reference

Please reference the issue this PR will close: None.

## Checklist

* [ ] Code compiles correctly
* [X] Created/updated tests


mukundansundar:
@artursouza Can you link to the root issue if possible?

artursouza:
This is a refactoring. There is no root issue.

Issue #309, opened 2020-07-20
Support multi pub/sub


mukundansundar:
## Describe the proposal

Dependent on runtime changes for https://github.com/dapr/dapr/issues/1709
Protobuf files changes, samples update and IT update.

PR #320, opened 2020-08-15
Support W3C tracing + example.

artursouza:
# Description

Adds support for W3C tracing via OpenTelemetrySDK.
Adds example for W3C tracing.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #298 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation

PR #323, opened 2020-08-19
Add support for multi-pubsub (#309)

wcs1only:
# Description

Support multi-pub/sub in the java sdk

## Issue reference

Please reference the issue this PR will close: #309

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,10 +22,10 @@ jobs:
       GOARCH: amd64
       GOPROXY: https://proxy.golang.org
       JDK_VER: 13.0.x
-      DAPR_RUNTIME_VER: 0.9.0-rc.1
-      DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/e498de9f7dd92c30aa592d6e6761dc924bb53cc2/install/install.sh
-      DAPR_CLI_REF: 35b9a824d2fd9e2dcf8e75d1d49ce59a39c7cf5b
-      DAPR_REF: b53586d1e5a880a3f87044975ace25b9ae51daec
+      DAPR_RUNTIME_VER: 0.10.0-rc.0
+      DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/e7c9a643dfefbcfff0c2c26c12029259e6e81180/install/install.sh
+      DAPR_CLI_REF: e7c9a643dfefbcfff0c2c26c12029259e6e81180
+      DAPR_REF: 98365b1b9ade55e7cf46cbc2313f0625318c0977
       OSSRH_USER_TOKEN: ${{ secrets.OSSRH_USER_TOKEN }}
       OSSRH_PWD_TOKEN: ${{ secrets.OSSRH_PWD_TOKEN }}
       GPG_KEY: ${{ secrets.GPG_KEY }}
@@ -66,15 +66,15 @@ jobs:
         cd ..
     - name: Initialize Dapr runtime ${{ env.DAPR_RUNTIME_VER }}
       run: |
-        sudo dapr init --runtime-version ${{ env.DAPR_RUNTIME_VER }}
-        echo "Showing dapr version..."
-        dapr --version
+        dapr uninstall --all
+        dapr init --runtime-version ${{ env.DAPR_RUNTIME_VER }}
     - name: Build and override daprd with referenced commit.
       if: env.DAPR_REF != ''
       run: |
         cd dapr
         make
-        sudo cp dist/linux_amd64/release/daprd /usr/local/bin/daprd
+        mkdir -p $HOME/.dapr/bin/
+        cp dist/linux_amd64/release/daprd $HOME/.dapr/bin/daprd
         cd ..
     - name: Override placement service.
       if: env.DAPR_REF != ''
--- a/sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
+++ b/sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
@@ -5,6 +5,7 @@
 
 package io.dapr.springboot;
 
+
 import io.dapr.actors.runtime.ActorRuntime;
 import io.dapr.serializer.DefaultObjectSerializer;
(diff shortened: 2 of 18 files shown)


wcs1only:
Reminder: update examples/src/main/java/io/dapr/examples/pubsub/http/README.md

.github/workflows/build.yml
```diff
@@ -66,15 +66,15 @@ jobs:
         cd ..
     - name: Initialize Dapr runtime ${{ env.DAPR_RUNTIME_VER }}
       run: |
+        sudo dapr uninstall
```
LMWF, line 69: I think you want to pass `--all` here.

sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
```diff
@@ -18,6 +19,7 @@
 import reactor.core.publisher.Mono;
 
 import java.io.IOException;
+import java.nio.charset.Charset;
```
mukundansundar, line 22: seems unused.

.github/workflows/build.yml
yaron2: Do we still need sudo now that the binary doesn't reside in /use/local/bin?

sdk-tests/src/test/java/io/dapr/it/pubsub/http/PubSubIT.java
```diff
@@ -30,6 +30,7 @@
   //Number of messages to be sent: 10
   private static final int NUM_MESSAGES = 10;
 
+  private static final String PUBSUB_NAME = "messagebus";
```
LMWF, line 33: You may want to make this public so you don't have to write "messagebus" each time in other files like SubscriberController.java.

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -100,20 +100,24 @@
...
+  public Mono<Void> publishEvent(String pubsubName, String topic, Object data, Map<String, String> metadata) {
     try {
       // TODO: handle metadata.
       DaprProtos.PublishEventRequest envelope = DaprProtos.PublishEventRequest.newBuilder()
-          .setTopic(topic).setData(ByteString.copyFrom(objectSerializer.serialize(data))).build();
+          .setTopic(topic)
+          .setPubsubName(pubsubName)
+          .setData(ByteString.copyFrom(objectSerializer.serialize(data))).build();
 
+      System.out.println("GRPCPublish " + pubsubName + " " + topic + " "
```
mukundansundar, line 119: remove

wcs1only: Whoops, nice catch.

.github/workflows/build.yml
LMWF: It shouldn't be needed if it's a clean environment.  But I'm noticing that if there was a previous dapr installation it may have set permissions in some folders that dapr init needs to write under ~/.dapr/ and some of those may not be overridable without a sudo.

sdk-springboot/src/main/java/io/dapr/springboot/DaprController.java
wcs1only: Right you are. I wonder why the static analyzer didn't pick up on this.

.github/workflows/build.yml
wcs1only: > Do we still need sudo now that the binary doesn't reside in /use/local/bin?

I had left it in case we were upgrading from an older version, but we probably don't need it going forward.

wcs1only:
So this will have conflicts with #320. It will probably be easier to deal with the conflicts if we merge #320 first.

yaron2:
It might be riskier to merge #320, I gather?

wcs1only:
> It might be riskier to merge #320, I gather?

Yes, I'll keep a branch around with my changes on top of master. If we run into any trouble, we can revert both and roll forward this change without #320.

amanbha:
> > It might be riskier to merge #320, I gather?
> 
> Yes, I'll keep a branch around with my changes on top of master. If we run into any trouble, we can revert both and roll forward this change without #320.

We should merge this first and when we decide to take #320, we can resolve the conflicts at that time.

wcs1only:
> > > It might be riskier to merge #320, I gather?
> > 
> > 
> > Yes, I'll keep a branch around with my changes on top of master. If we run into any trouble, we can revert both and roll forward this change without #320.
> 
> We should merge this first and when we decide to take #320, we can resolve the conflicts at that time as merging #320 is not signed off for this release.

Ok, I have no strong feeling about the order, so if we don't think #320 is ready, I'm fine either way.

.github/workflows/build.yml
```diff
@@ -22,10 +22,10 @@ jobs:
...
       GOPROXY: https://proxy.golang.org
       JDK_VER: 13.0.x
-      DAPR_RUNTIME_VER: 0.9.0-rc.1
-      DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/e498de9f7dd92c30aa592d6e6761dc924bb53cc2/install/install.sh
-      DAPR_CLI_REF: 35b9a824d2fd9e2dcf8e75d1d49ce59a39c7cf5b
-      DAPR_REF: b53586d1e5a880a3f87044975ace25b9ae51daec
+      DAPR_RUNTIME_VER: 0.10.0-rc.0
+      DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/e7c9a643dfefbcfff0c2c26c12029259e6e81180/install/install.sh
+      DAPR_CLI_REF: e7c9a643dfefbcfff0c2c26c12029259e6e81180
+      DAPR_REF: 98365b1b9ade55e7cf46cbc2313f0625318c0977
```
mukundansundar, line 28: i think this Dapr Ref is not needed as it refers to the same runtime version

wcs1only: We can fix this when we cut the release today.

PR #330, opened 2020-08-25
Add support for DAPR_API_TOKEN

pruthvidhodda:
# Description

Added support for DAPR_API_TOKEN

## Issue reference

Please reference the issue this PR will close: #303 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/pom.xml
```diff
@@ -84,6 +84,12 @@
       <version>1.3.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
```
artursouza, line 88: There is no need to add this dependency. All env variables can be override by Java properties in our Properties solutions in this project. That was done on purpose.

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -258,6 +259,12 @@ private Response doInvokeApi(String method,
     } else {
       requestBuilder.method(method, body);
     }
+
+    String daprApiToken = Properties.getStringOrDefault(Constants.DAPR_API_TOKEN, Constants.DAPR_API_TOKEN, null);
```
artursouza, line 263: Do not use the same key for Java property and Env Variable. The naming convention is different. Instead, add it directly to the Properties class as a new public static field. I recommend to refactor all Properties.get*OrDefault() to private.

sdk/src/main/java/io/dapr/utils/Constants.java
```diff
@@ -60,6 +60,16 @@
    */
   public static final String ACTOR_TIMER_RELATIVE_URL_FORMAT = ACTORS_BASE_URL + "/%s/%s/timers/%s";
 
+  /**
+   * Environment variable name for dapr api token.
+   */
+  public static final String DAPR_API_TOKEN = "DAPR_API_TOKEN";
```
artursouza, line 66: Please, do not add constants here. I actually think this class should be gone and constants should be declared only when they are used. There might be a few shared constants here but it should be kept to a minimum.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -27,6 +29,9 @@
 
 public class DaprHttpTest {
 
+  @Rule
+  public final EnvironmentVariables environmentVariables = new EnvironmentVariables();
```
artursouza, line 33: No need for this since you can achieve the override by system properties.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -42,6 +47,38 @@ public void setUp() throws Exception {
     okHttpClient = new OkHttpClient.Builder().addInterceptor(mockInterceptor).build();
   }
 
+  @Test
+  public void invokeApi_daprApiToken_present() throws IOException {
+    mockInterceptor.addRule()
+        .post("http://127.0.0.1:3500/v1.0/state")
+        .hasHeader(Constants.DAPR_API_TOKEN_HEADER)
+        .respond(serializer.serialize(EXPECTED_RESULT));
+    environmentVariables.set(Constants.DAPR_API_TOKEN, "xyz");
```
artursouza, line 56: Do not set environment variable, set the property instead and the behavior should be the same.

sdk/src/main/java/io/dapr/utils/Constants.java
pruthvidhodda: Is there a reason why we don't want to do this? This constant is being used in three files, so I find it reasonable to declare here.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
pruthvidhodda: Right we could use the system properties but the main workflow we are trying to test is: when a user sets DAPR_API_TOKEN env variable then java-sdk needs to set dapr-api-token header in all dapr requests.

If we just test using system properties and let's say someone changes the behavior of Properties.getValueOrDefault() to not check environment variables then the test would still pass but [this use case](https://github.com/dapr/docs/tree/master/howto/enable-dapr-api-token-based-authentication#self-hosted) would be broken.

sdk/src/main/java/io/dapr/client/DaprHttp.java
pruthvidhodda: Makes sense, added it as property.

sdk/src/main/java/io/dapr/utils/Properties.java
```diff
@@ -35,6 +35,13 @@
    */
   private static final String DEFAULT_STRING_CHARSET = StandardCharsets.UTF_8.name();
 
+  /**
+   * API token for Dapr after checking system property and environment variable.
+   */
+  public static final Supplier<String> DAPR_API_TOKEN = () -> getStringOrDefault(
+      "dapr.api.token",
+      Constants.DAPR_API_TOKEN, null);
```
artursouza, line 43: This is the point I was trying to make. See how this constant is only used here and in the unit tests. It can be easily removed from the Constants class and be hardcoded here. The test class can have its own constant too.

If constants need to be referenced outside the class, it is better to have a contextualized class instead of a generic class called "Constants". This is why I am trying to stop adding to the Constants class.

sdk/src/main/java/io/dapr/utils/Constants.java
artursouza: If constants are shared, it should be an owner with the right context. Having this wild card class with every constant in the project hurts readability. This was imported from the .Net class and needs refactoring. This is why I am avoiding adding more to it.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
artursouza: You are right about the use case test, on the other hand, unit test assume abstractions from other classes. The scenario you are describing would be better tested in an integration test. I am OK with this now mainly because the dependency is for tests only and not released.

PR #333, opened 2020-09-01
Refactoring properties and constants.

artursouza:
# Description

Move constants to correct corresponding owner class.
Remove Constants class.
Add SIDECAR_IP env var and property (defaults to 127.0.0.1).
Refactor properties to support reference to name and not only value.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


pruthvidhodda:
Looks like there are some more places where Constants reference needs to be removed, else the changes LGTM

sdk-tests/src/test/java/io/dapr/actors/runtime/DaprClientHttpUtils.java
```diff
@@ -0,0 +1,21 @@
...
+ * Exposes useful methods for IT in DaprClientHttp.
+ */
+public class DaprClientHttpUtils {
+
+  public static void unregisterActorReminder(
+      DaprHttp client,
+      String actorType,
+      String actorId,
+      String reminderName) throws Exception {
+    new DaprHttpClient(client).unregisterActorReminder(actorType, actorId, reminderName).block();
```
mukundansundar, line 19: Is a new client creation necessary for each call?

sdk/src/main/java/io/dapr/config/Property.java
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @return Value from system property (1st) or env variable (2nd) or default (last).
+   */
+  public T get() {
+    String propValue = System.getProperty(this.name);
+    if (propValue != null && !propValue.trim().isEmpty()) {
+      try {
+        return this.parse(propValue);
+      } catch (Exception e) {
+        e.printStackTrace();
```
mukundansundar, line 68: Log that error instead of printing stack trace.

sdk/src/main/java/io/dapr/config/Property.java
```diff
@@ -0,0 +1,92 @@
...
+        // OK, we tried. Falling back to system environment variable.
+      }
+    }
+
+    String envValue = System.getenv(this.envName);
+    if (envValue != null && !envValue.trim().isEmpty()) {
+      try {
+        return this.parse(envValue);
+      } catch (Exception e) {
+        e.printStackTrace();
```
mukundansundar, line 78: Same here.

sdk-tests/src/test/java/io/dapr/actors/runtime/DaprClientHttpUtils.java
artursouza: This is just for ITs.

sdk/src/main/java/io/dapr/config/Property.java
artursouza: We do not log anything in the SDK, we throw exceptions and expect user code to handle those. I will just silence the exception here.

mukundansundar (dismissed):

mukundansundar (approved):

Issue #299, opened 2020-07-07
Handle passing metadata in GRPC


mukundansundar:
Handle passing of the HTTP verb and content-type metadata for GRPC calls

PR #334, opened 2020-09-03
Add content type metadata support in gRPC invoke service calls

pruthvidhodda:
# Description

Added content type metadata support in gRPC invoke service calls

## Issue reference

Please reference the issue this PR will close: #299 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
```diff
@@ -110,52 +101,101 @@
...
+      Class<T> clazz) {
     return this.invokeService(appId, method, request, httpExtension, null, TypeRef.get(clazz));
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public <T> Mono<T> invokeService(String appId, String method, Object request,
+      HttpExtension httpExtension, Class<T> clazz, String contentType) {
```
artursouza, line 195: contenttype is determined by the serialization of the request object. every call to this method will be serialized to the same type. It feels that contenttype fits better in the constructor for when users pass in a custom serializer. The default serializer will always be application/json.

Another option is to add a new method to the serializer: getContentType(), so there is no overload needed here.

Also, the request object is a place to avoid adding more overloads but I don't think we should add contenttype as a new parameter here anyway.

PS: This comment applies to every method with contenttype param here.

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -382,14 +386,19 @@
...
    * @param httpExtension Object for HttpExtension
    * @param appId         The application id to be invoked
+   * @param contentType   The content type of the request body
    * @param method        The application method to be invoked
    * @param request       The body of the request to be send as part of the invocation
    * @param <K>           The Type of the Body
    * @return The object to be sent as part of the invocation.
    * @throws IOException If there's an issue serializing the request.
    */
   private <K> DaprProtos.InvokeServiceRequest buildInvokeServiceRequest(
```
artursouza, line 396: This method invokes objectSerializer, so it can get the contenttype via objectSerializer.getContentType().

artursouza (changes requested):

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
pruthvidhodda: Makes sense, I'll make the changes now.

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
pruthvidhodda: Removed it.

artursouza (approved):

Issue #396, opened 2020-11-26
Review and ensure metadata is correctly set and transport 


skyao:
## Expected Behavior

metadata should be correctly set and passed to dapr.

## Actual Behavior

At first, in state save method, there is no metadata to set. 

See file `src/main/java/io/dapr/client/domain/State.java`

I will submit a PR for this and later check other processes.

RELEASE NOTE: **RESOLVED** validated handling of metadata and fixed when applicable.

skyao:
@artursouza Another PR has raised to add metadata parameter in get bulk request which is missing in java sdk. Please help to review. And now I only update the grpc api, for HTTP API, should we also support metadata in get bulk request (and also there is a ongoing PR to add metadata in get/bulk get response).

skyao:
@artursouza A new more PR for metadata just raised.

artursouza:
@skyao Thanks a lot. I will review the PR. Please, keep feature parity for the GRPC and HTTP clients for Dapr.

artursouza:
I am handling all the remaining metadata logic in the PR above.

PR #397, opened 2020-11-26
Add metadata for state save operation

skyao:
# Description

In state save operation, metadata is missing in request.

## Issue reference

https://github.com/dapr/java-sdk/issues/396

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/src/main/java/io/dapr/client/domain/State.java
```diff
@@ -191,6 +230,10 @@ public boolean equals(Object o) {
       return false;
     }
 
+    if (getMetadata() != null ? !getMetadata().equals(that.getMetadata()) : that.getMetadata() != null) {
```
artursouza, line 233: Does this equal method work for maps? Will it make a deep comparison? This might be worth unit testing. Can you add unit test for this class to validate equals and hashcode?

artursouza (changes requested):

mukundansundar (commented):
@skyao this change should also be done for `executeTransaction`, not only `saveStates` call, for both gRPC and HTTP.

skyao (commented):
Hashcode() and equals() method verified. 

But it depends on the map implementation: java.util.HashMap is OK, but for other implementation the result will depends on the impementation's equals() and hashcode() method.

I think this is not the problem of dapr itself, since we can only use Map interface here, not HashMap impmentation.

skyao:
> @skyao this change should also be done for `executeTransaction`, not only `saveStates` call, for both gRPC and HTTP.

Update for executeTransaction, only need to change HTTP. For gRPC, it reused the same buildStateRequest() method .

artursouza (approved):

PR #406, opened 2020-12-10
IT for recovering reminder after sidecar restart.

artursouza:
# Description

IT for recovering reminder after sidecar restart.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/2554

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


mukundansundar (approved):
for `countReminderCalls` could the filter be applied to filter only `Enter` based entry logs. thus you need not divide by two. But i can see that you are also tracking exit of the reminder so its fine.

PR #407, opened 2020-12-10
Fix actor sample + remove HTTP reference in docs and code.

artursouza:
# Description

Fix actor sample + remove HTTP reference in docs and code.

## Issue reference

Please reference the issue this PR will close: #408 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/examples/src/main/java/io/dapr/examples/actors/DemoActorService.java
+++ b/examples/src/main/java/io/dapr/examples/actors/DemoActorService.java
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-package io.dapr.examples.actors.http;
+package io.dapr.examples.actors;
 
 import io.dapr.actors.runtime.ActorRuntime;
 import io.dapr.springboot.DaprApplication;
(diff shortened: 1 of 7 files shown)


examples/src/main/java/io/dapr/examples/actors/DemoActorService.java
```diff
@@ -20,8 +20,7 @@
  * mvn clean install
  * 2. Run the server:
  * dapr run --components-path ./components --app-id demoactorservice --app-port 3000 --dapr-http-port 3005 \
```
mukundansundar, line 22: should the http port be removed here?

artursouza: Yes. Thanks for reminding me.

mukundansundar (approved):

mukundansundar (changes requested):
integration tests are failing. 
Additionally there is some small mistake with logging. 
In receiveReminder test
```
Size of timer count list is %d, which means it's been invoked half that many times6
```

```
 System.out.printf(
        "Size of reminder count list is %d, which means it's been invoked half that many times.", calls.size());
```

mukundansundar (approved):

PR #411, opened 2020-12-11
IT for failover actor reminder.

artursouza:
# Description

IT for failover actor reminder.

## Issue reference

Relates to: https://github.com/dapr/dapr/issues/2554

## Checklist

* [ ] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/actors/ActorReminderFailoverIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/actors/ActorReminderFailoverIT.java
@@ -1,0 +1,164 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.it.actors;
+
+import io.dapr.actors.ActorId;
+import io.dapr.actors.client.ActorProxy;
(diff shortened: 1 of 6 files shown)


mukundansundar:
the reminder fire IT added recently seems to be flaky. A rerun of build passed the test https://github.com/dapr/java-sdk/pull/407/checks?check_run_id=1534796822 when before it was failing.

artursouza:
I can look at it again when it fails.

mukundansundar (approved):
Looks good. But one question. 
@artursouza Is there a scenario where the logs we parse contain only 2 exit logs and the minimum is 1 call so we say the reminder fired because of 2 exit logs with no entry log ... do we need to parse filter and count the Entry and Exit logs separately?

artursouza:
> Looks good. But one question.
> @artursouza Is there a scenario where the logs we parse contain only 2 exit logs and the minimum is 1 call so we say the reminder fired because of 2 exit logs with no entry log ... do we need to parse filter and count the Entry and Exit logs separately?

It is unlikely the exit log to be fired without the entry log. We can change the count logic to check for in enter with a matching exit only. This is very unlikely to happen though.

PR #414, opened 2020-12-15
Set etag to be null if it's empty

### LINKED ISSUE #405: States without etag is handled differently in gRPC and HTTP APIs
## Expected Behavior

Get a state without eTag in gRPC and HTTP, eTag attribute should be null.

## Actual Behavior

eTag is null in HTTP and empty string in gRPC.

## Steps to Reproduce the Problem

Get state that does not have eTag.
Try the same in getBulkState too.

## Release Note

RELEASE NOTE: **FIX** missing eTag response for state APIs.

arghya88:
Signed-off-by: Arghya Sadhu <arghya88@gmail.com>

# Description

Convert empty string to null for GRPC implementation to keep it consistent with HTTP implementation

## Issue reference

Fixes #405

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/state/AbstractStateClientIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/state/AbstractStateClientIT.java
@@ -74,13 +74,11 @@ public abstract class AbstractStateClientIT extends BaseIT {
     Assert.assertNotNull(state);
     Assert.assertEquals("unknownKey", state.getKey());
     Assert.assertNull(state.getValue());
-    // gRPC returns empty eTag while HTTP returns null.
-    // TODO(artursouza): https://github.com/dapr/java-sdk/issues/405
-    Assert.assertTrue(state.getEtag() == null || state.getEtag().isEmpty());
+    Assert.assertNull(state.getEtag());
   }
 
   @Test
-  public void saveAndGetBulkStates() {
+  public void saveAndGetBulkState() {
     final String stateKeyOne = UUID.randomUUID().toString();
     final String stateKeyTwo = UUID.randomUUID().toString();
     final String stateKeyThree = "NotFound";
@@ -114,8 +112,8 @@ public abstract class AbstractStateClientIT extends BaseIT {
 
     assertEquals(stateKeyThree, result.stream().skip(2).findFirst().get().getKey());
     assertNull(result.stream().skip(2).findFirst().get().getValue());
-    assertEquals("", result.stream().skip(2).findFirst().get().getEtag());
-    assertNull("not found", result.stream().skip(2).findFirst().get().getError());
+    assertNull(result.stream().skip(2).findFirst().get().getEtag());
+    assertNull(result.stream().skip(2).findFirst().get().getError());
   }
 
   @Test
--- a/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
@@ -320,6 +320,9 @@ public class DaprClientGrpc extends AbstractDaprClient {
     byte[] data = payload == null ? null : payload.toByteArray();
     T value = stateSerializer.deserialize(data, type);
     String etag = item.getEtag();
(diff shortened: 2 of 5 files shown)


mukundansundar (commented):
lgtm.
@artursouza The issue linked to this should be a breaking change since the response is changing.

artursouza:
Yes, LGTM. Let's mark it as breaking change - even though it is minor.

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -332,6 +332,9 @@
     byte[] data = payload == null ? null : payload.toByteArray();
     T value = stateSerializer.deserialize(data, type);
     String etag = response.getEtag();
+    if (etag.equals("")) {
```
artursouza, line 335: Bulk state needs this change too. See method above this one.

artursouza (changes requested):

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
mukundansundar: good catch. missed that ..

arghya88: Done

sdk-tests/src/test/java/io/dapr/it/state/AbstractStateClientIT.java
```diff
@@ -109,12 +109,12 @@ public void saveAndGetBulkStates() {
 
     assertEquals(stateKeyTwo, result.stream().skip(1).findFirst().get().getKey());
     assertNull(result.stream().skip(1).findFirst().get().getValue());
-    assertEquals("1", result.stream().skip(1).findFirst().get().getEtag());
+    assertNull(result.stream().skip(1).findFirst().get().getEtag());
```
arghya88, line 112: @artursouza I wonder how this assert was successful before.

artursouza: Ops. Let me see if this is a bug that was fixed or created.

artursouza: I see that his change is not in anymore. So I assume it was a red herring.

arghya88:
@artursouza I debugged why `saveAndGetBulkState` test in `AbstractStateClientIT` is failing. It turns out

1. For the second state object both GRPC and HTTP client returns `1` for Etag even though we set `null` in the object. 
2. For the third object where we don't even save it in state store HTTP client returns empty and GRPC client returns `null` for Etag.

I am not sure if 1 is expected to be this way. If this is expected then we can keep it this way and let the test pass.
For 2 the tests used to pass before my change because GRPC client used to return empty before my change. With my change it returns null instead of empty and hence  test for GRPC client is failing.

arghya88:
@artursouza Pls review

artursouza:
> @artursouza I debugged why `saveAndGetBulkState` test in `AbstractStateClientIT` is failing. It turns out
> 
> 1. For the second state object both GRPC and HTTP client returns `1` for Etag even though we set `null` in the object.
> 2. For the third object where we don't even save it in state store HTTP client returns empty and GRPC client returns `null` for Etag.
> 
> I am not sure if 1 is expected to be this way. If this is expected then we can keep it this way and let the test pass.
> For 2 the tests used to pass before my change because GRPC client used to return empty before my change. With my change it returns null instead of empty and hence test for GRPC client is failing.
> 
> 2 is not happening in unit tests because units tests don't cover a scenario of trying to retrieve a state object which is not saved at all.

The reason why the etag is "1" and not null even when we set it to null is because Dapr maps etag to version in Redis. Version is always present in Redis, so it defaults to "1" when we set a value: https://github.com/dapr/components-contrib/blob/master/state/redis/redis.go

This is why the integration tests asserts "1" for the null case.

artursouza (approved):

Issue #427, opened 2020-12-23
HTTP input is not properly encoded.


artursouza:
## Expected Behavior

If the value of a state store (for example) has a "/", then it does not affect the URL path and the Dapr API works as expected.

## Actual Behavior

Dapr returns `Method Not Allowed` error because "/" (and other special chars) affect the URL path.

## Steps to Reproduce the Problem

Try to save state where state store contains "/" in the middle of the name.

## Release Note

RELEASE NOTE: **FIXED** input encoding for HTTP calls.

mukundansundar:
breaking change since constants are removed.

artursouza:
This is not breaking change because those constants are public simply because they are used in other packages. In Java there is no way to mark that as internal.

PR #428, opened 2020-12-23
Fix input encoding and error parsing for HTTP calls.

artursouza:
# Description

Fix input encoding and error parsing for HTTP calls.

## Issue reference

Please reference the issue this PR will close: #427 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
@@ -36,9 +36,9 @@ class DaprHttpClient implements DaprClient {
    */
   @Override
   public Mono<byte[]> invoke(String actorType, String actorId, String methodName, byte[] jsonPayload) {
-    String[] path = new String[] { DaprHttp.API_VERSION, "actors", actorType, actorId, "method", methodName };
+    String[] pathSegments = new String[] { DaprHttp.API_VERSION, "actors", actorType, actorId, "method", methodName };
     Mono<DaprHttp.Response> responseMono =
-          this.client.invokeApi(DaprHttp.HttpMethods.POST.name(), path, null, jsonPayload, null, null);
+          this.client.invokeApi(DaprHttp.HttpMethods.POST.name(), pathSegments, null, jsonPayload, null, null);
     return responseMono.map(r -> r.getBody());
   }
 }
--- a/sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
@@ -16,16 +16,6 @@ import reactor.core.publisher.Mono;
 class DaprHttpClient implements DaprClient {
 
   /**
-   * Base URL for Dapr Actor APIs.
-   */
-  private static final String ACTORS_BASE_URL = DaprHttp.API_VERSION + "/" + "actors";
-
-  /**
-   * String format for Actors method invocation relative url.
-   */
-  private static final String ACTOR_METHOD_RELATIVE_URL_FORMAT = ACTORS_BASE_URL + "/%s/%s/method/%s";
-
-  /**
    * The HTTP client to be used.
    *
    * @see DaprHttp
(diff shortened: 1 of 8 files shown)


sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -153,26 +153,26 @@ public int getStatusCode() {
...
+   * @param path          Array of path segments ("/a/b/c" maps to ["a", "b", "c"]).
    * @param urlParameters URL parameters
    * @param headers       HTTP headers.
    * @param context       OpenTelemetry's Context.
    * @return Asynchronous text
    */
   public Mono<Response> invokeApi(
       String method,
-      String urlString,
+      String[] path,
```
rynowak, line 164: suggestion - `pathSegments` might be a clearer name. Path is already a concept that's defined by URIs and most people wouldn't think of it as an array

wcs1only (commented):
Changes look good, but do you have a test that reproduces #427 that is now passing?

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -317,7 +318,12 @@ private static DaprError parseDaprError(byte[] json) throws IOException {
     if ((json == null) || (json.length == 0)) {
       return null;
     }
-    return OBJECT_MAPPER.readValue(json, DaprError.class);
+
+    try {
+      return OBJECT_MAPPER.readValue(json, DaprError.class);
+    } catch (JsonParseException e) {
+      throw new DaprException("UNKNOWN", new String(json, StandardCharsets.UTF_8));
+    }
```
rynowak, line 326: Is this supposed to be here? What are the cases where this is called?

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -215,8 +215,8 @@ public void invokePostMethodUnknownError() throws IOException {
   public void testCallbackCalledAtTheExpectedTimeTest() throws IOException {
     String deletedStateKey = "deletedKey";
     String existingState = "existingState";
-    String urlDeleteState = DaprClientHttp.STATE_PATH + "/" + deletedStateKey;
-    String urlExistingState = DaprClientHttp.STATE_PATH + "/" + existingState;
+    String urlDeleteState = STATE_PATH + "/" + deletedStateKey;
+    String urlExistingState = STATE_PATH + "/" + existingState;
```
rynowak, line 219: It seems little wierd to do the work in the tests of building a path and then splitting it again because the API takes `string[]`

sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
```diff
@@ -46,9 +36,9 @@
    */
   @Override
   public Mono<byte[]> invoke(String actorType, String actorId, String methodName, byte[] jsonPayload) {
-    String url = String.format(ACTOR_METHOD_RELATIVE_URL_FORMAT, actorType, actorId, methodName);
+    String[] path = new String[] { DaprHttp.API_VERSION, "actors", actorType, actorId, "method", methodName };
     Mono<DaprHttp.Response> responseMono =
-          this.client.invokeApi(DaprHttp.HttpMethods.POST.name(), url, null, jsonPayload, null, null);
+          this.client.invokeApi(DaprHttp.HttpMethods.POST.name(), path, null, jsonPayload, null, null);
```
rynowak, line 41: Is there a test for the case that's being fixed? (Reserved URL Path) character in this state store or method name?

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
artursouza: It was to minimize changes.

sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
artursouza: Will do.

sdk/src/main/java/io/dapr/client/DaprHttp.java
artursouza: This can happen if Dapr does not unescape correctly and returns a bad response at the HTTP server prior to the request being processed by dapr logic. I reproduced this with daprd with a bad patch from another PR - so I thought it would be good to have a better experience in the SDK if that kind of corner case happens.

sdk/src/main/java/io/dapr/client/DaprHttp.java
artursouza: I can rename it.

wcs1only (approved):
Thanks for adding a test.

Issue #434, opened 2021-01-04
Blocking calls for gRPC and HTTP even though using Reactor


xiazuojie:
## Expected Behavior
Implementations of DaprClient API being non-blocking.

## Actual Behavior

[DaprClient](https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/DaprClient.java) API is reactive with Reactor. However, its implementations (gPRC and HTTP/1.x) at the moment  are blocking, which defeats the purpose of using Reactive API.
[gRPC](https://github.com/dapr/java-sdk/blob/6a44d51279d7d6d5ccdfd63346b4c83d00e7eeb8/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java#L677):
```java
return Mono.fromCallable(wrap(context, () -> {
        ListenableFuture<DaprProtos.GetStateResponse> futureResponse = client.getState(envelope);
        return buildStateKeyValue(get(futureResponse), key, options, type);
      })).map(s -> new Response<>(context, s));
```

```java
  private static <V> V get(ListenableFuture<V> future) {
    try {
      return future.get();
    } catch (Exception e) {
      DaprException.wrap(e);
    }

    return null;
  }
```
[HTTP/1.1](https://github.com/dapr/java-sdk/blob/6a44d51279d7d6d5ccdfd63346b4c83d00e7eeb8/sdk/src/main/java/io/dapr/client/DaprHttp.java#L208):
```java
  public Mono<Response> invokeApi(
          String method,
          String[] pathSegments,
          Map<String, String> urlParameters,
          byte[] content,
          Map<String, String> headers,
          Context context) {
    return Mono.fromCallable(() -> doInvokeApi(method, pathSegments, urlParameters, content, headers, context));
  }
```

Note ``Mono.fromCallable`` only adapts the API to reactive. It still blocks a thread, be it the calling thread or the thread the Mono is subscribed on.

## Release Note

RELEASE NOTE: **FIXED** implementations of DaprClient API were blocking for gRPC and HTTP.

artursouza:
@xiazuojie Do you have a proposal to fix this? The same also happens in the gRPC client. Since Java does not have async/await like .Net, I wonder how the Java community solves this.

xiazuojie:
> @xiazuojie Do you have a proposal to fix this? Since Java does not have async/await like .Net, I wonder how the Java community solves this.

Async is natively supported by gRPC java. For HTTP/1.x it's much more complicated.
[...]

PR #435, opened 2021-01-07
Fix blocking calls for gRPC even though using Reactor

xiazuojie:
# Description

Make DaprClientGrpc truly async and update unit tests.

## Issue reference

Reference: #434 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/src/main/java/io/dapr/exceptions/DaprException.java
```diff
@@ -151,6 +145,37 @@ public static void wrap(Exception exception) {
     return Mono.empty();
   }
 
+  /**
+   * Wraps an exception into DaprException (if not already DaprException).
+   *
+   * @param exception Exception to be wrapped.
+   * @return wrapped RuntimeException
+   */
+  public static RuntimeException propagate(Throwable exception) {
```
artursouza, line 154: We don't want to wrap throwable. It will catch stuff like https://docs.oracle.com/javase/7/docs/api/java/lang/OutOfMemoryError.html

sdk/src/main/java/io/dapr/exceptions/DaprException.java
```diff
@@ -37,15 +37,15 @@ public DaprException(DaprError daprError) {
    *                  permitted, and indicates that the cause is nonexistent or
    *                  unknown.)
    */
-  public DaprException(DaprError daprError, Exception cause) {
+  public DaprException(DaprError daprError, Throwable cause) {
```
artursouza, line 40: No throwable, just Exception.

sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
```diff
@@ -167,44 +162,29 @@ public void publishEventSerializeException() throws IOException {
...
+    }).when(client).publishEvent(any(DaprProtos.PublishEventRequest.class), any());
+
     Mono<Void> result = adapter.publishEvent("pubsubname","topic", "object");
-    settableFuture.set(Empty.newBuilder().build());
     result.block();
-    assertTrue(callback.wasCalled);
-  }
-
-  @Test
-  public void publishEventNoHotMono() {
```
artursouza, line 182: Please, bring back all the hot mono tests. These are critical.

artursouza (changes requested):
Thanks for the PR, it is promising. There are two points of contention for me here. First, the unit tests for hot mono should stay - that is a big deal when dealing with Reactor and we found it the hard way early in this project. Second, we should not wrap (or catch) throwables otherwise we end up in situations where we handle or wrap OutOfMemoryError, for example.

sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
xiazuojie: I do not see the point of unit testing "hot" mono for every scenery. Once get familiar with reactive programming, one would know that nothing happens until a stream is subscribed.

sdk/src/main/java/io/dapr/exceptions/DaprException.java
xiazuojie: I've added ```Exceptions.throwIfFatal(exception);``` from project reactors.
Reactor deals with throwable wrapping with ```reactor.core.Exceptions#propagate```.

sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
artursouza: The problem was that the code was actually invoking a service call and just wrapping the result with Mono.just(). These tests are here to make sure this issue does not come back. I prefer to have the test over trusting that everyone touching this code going forward will avoid this bug.

sdk/src/main/java/io/dapr/exceptions/DaprException.java
artursouza: OK.

artursouza:
The only thing now is to bring back the hot mono unit tests.

sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
xiazuojie: Ok. Added "NoHotMono" unit tests

artursouza (approved):

Issue #444, opened 2021-01-17
@ActorMethod to override method's name for client and service


artursouza:
Today, we have Actor for Dapr implemented in 3 languages: C#, Java and Python. Each language has their own convention in how a method should be named. Then, actor method invocation does not work cross language without the caller breaking the language's convention when defining the Actor's interface.

Example of Actor interface in the service side in C#:
```C#
    public interface IMyActor : IActor
    {      
        Task<int> IncrementAndGetAsync(int delta);
    }
```

Now, the same interface could be represented at the caller's side in Java:
```Java
@ActorType(name = "MyActor")
public interface MyActor {
  @ActorMethod(name = "IncrementAndGetAsync")
  int incrementAndGet(int delta);
}
```

Instead of:
```Java
@ActorType(name = "MyActor")
public interface MyActor {
  int IncrementAndGetAsync(int delta);
}
```

This feature should also work the other way around, where the Java service would accept "IncrementAndGetAsync" instead of the native method name.

Note: The SDK already contains the `@ActorMethod` annotation but it does not have the `name` attribute.

Relates to: https://github.com/dapr/dapr/issues/1165

artursouza:
This issue does not affect C# because C# does not use native method invocation for cross language invocation (without remoting), it uses the following API instead:
```C#
      var proxy = ActorProxy.Create(new ActorId(actorId), actorType);
      return await proxy.InvokeAsync<int, int>("incrementAndGet", 1);
```

artursouza:
Python SDK already has this feature: https://github.com/dapr/python-sdk/blob/master/examples/demo_actor/demo_actor/demo_actor_interface.py

PR #448, opened 2021-01-19
Support ActorMethod to change method's name.

artursouza:
# Description

Support ActorMethod to change method's name.

## Issue reference

Please reference the issue this PR will close: #444

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/examples/src/main/java/io/dapr/examples/actors/README.md
+++ b/examples/src/main/java/io/dapr/examples/actors/README.md
@@ -126,7 +126,7 @@ public interface DemoActor {
 
 ```
 
-The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. The `say()` method, in this example, is renamed to `echo_message()` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
+The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. The `say` method, in this example, is renamed to `echo_message` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
 Some methods can return a `Mono` object. In these cases, the `@ActorMethod` annotation is used to hint the Dapr Java SDK of the type encapsulated in the `Mono` object. You can read more about Java generic type erasure [here](https://docs.oracle.com/javase/tutorial/java/generics/erasure.html).  
 
 
--- a/examples/src/main/java/io/dapr/examples/actors/README.md
+++ b/examples/src/main/java/io/dapr/examples/actors/README.md
@@ -126,7 +126,9 @@ public interface DemoActor {
 
 ```
 
-The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. The `say` method, in this example, is renamed to `echo_message` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
+The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. 
+
+The `@ActorMethod` annotation can be applied to an interface method to specify configuration for that method. In this example, the `say` method, is renamed to `echo_message` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
 Some methods can return a `Mono` object. In these cases, the `@ActorMethod` annotation is used to hint the Dapr Java SDK of the type encapsulated in the `Mono` object. You can read more about Java generic type erasure [here](https://docs.oracle.com/javase/tutorial/java/generics/erasure.html).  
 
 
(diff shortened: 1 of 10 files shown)


sdk-actors/src/main/java/io/dapr/actors/ActorMethod.java
```diff
@@ -21,5 +21,12 @@
    *
    * @return Actor's method return type.
    */
-  Class returns();
+  Class returns() default Void.class;
```
artursouza, line 24: To make this optional, a default value must be provided. On the other hand, `null` is not allowed in Java for annotation attributes. The best I found was to use `Void.class` here. The idea is that this is only needed when the result object is Mono<T>, so T can be hinted to the Actor runtime - this is a limitation of the Java programming language that we need to overcome.

examples/src/main/java/io/dapr/examples/actors/README.md
```diff
@@ -123,7 +126,8 @@ public interface DemoActor {
 
 ```
 
-The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. Some methods can return a `Mono` object. In these cases, the `@ActorMethod` annotation is used to hint the Dapr Java SDK of the type encapsulated in the `Mono` object. You can read more about Java generic type erasure [here](https://docs.oracle.com/javase/tutorial/java/generics/erasure.html).  
+The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. The `say()` method, in this example, is renamed to `echo_message()` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
```
rynowak, line 129:
```suggestion
The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. The `say` method, in this example, is renamed to `echo_message` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
```

Suggestion, remove the parenthesis from the example because you won't use parenthesis to invoke the actor method by name, it's slightly confusing.

rynowak: Suggestion, make `@ActorType` and `@ActorMethod` separate paragraphs. Since this immediately follows the description of `@ActorType` it's not clear what it's related to. 

`@ActorMethod` hasn't been mentioned yet. So mention that :)

sdk-actors/src/main/java/io/dapr/actors/ActorMethod.java
rynowak: The only other option I could think of would be to declare your own sentinel value.

I'm fine either way.

sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
```diff
@@ -136,29 +136,33 @@ public Object invoke(Object proxy, Method method, Object[] args) {
...
 
     if (method.getReturnType().equals(Mono.class)) {
-      ActorMethod actorMethodAnnotation = method.getDeclaredAnnotation(ActorMethod.class);
       if (actorMethodAnnotation == null) {
-        return invokeMethod(method.getName(), args[0]);
+        return invokeMethod(methodName, args[0]);
       }
 
-      return invokeMethod(method.getName(), args[0], actorMethodAnnotation.returns());
+      return invokeMethod(methodName, args[0], actorMethodAnnotation.returns());
```
rynowak, line 162: Do we need any special case logic for the case for `Void.class` as the value of `returns()` the old behavior (not specifying return type for `Mono`) would send it down the code path at line 159

artursouza: Void is a valid option. So we don't know if customers intended Void or if they just forgot and left the default there.

sdk-actors/src/main/java/io/dapr/actors/ActorMethod.java
artursouza: I am leaning towards using a sentinel class, called `Default` for that.

artursouza: Created an "internal" class called "Undefined". It reads better now. Please, take a look.

sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
artursouza: Updated with logic now that we have the "Undefined" class.

sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
```diff
@@ -21,6 +21,8 @@
  */
 class ActorProxyImpl implements ActorProxy, InvocationHandler {
 
+  private static final String UNDEFINED_CLASS_NAME = "io.dapr.actors.Undefined";
```
artursouza, line 24: We cannot refer to the class since it is internal to the package in which it was declared.

examples/src/main/java/io/dapr/examples/actors/README.md
```diff
@@ -123,7 +126,8 @@ public interface DemoActor {
 
 ```
 
-The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. Some methods can return a `Mono` object. In these cases, the `@ActorMethod` annotation is used to hint the Dapr Java SDK of the type encapsulated in the `Mono` object. You can read more about Java generic type erasure [here](https://docs.oracle.com/javase/tutorial/java/generics/erasure.html).  
+The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. The `say` method, in this example, is renamed to `echo_message` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
```
rynowak, line 129: This still runs together.

```suggestion
The `@ActorType` annotation indicates the Dapr Java SDK that this interface is an Actor Type, allowing a name for the type to be defined. 

The `@ActorMethod` annotation can be applied to an interface method to specify configuration for that method. In this example, the `say` method, is renamed to `echo_message` - this can be used when invoking an actor method implemented in a different programming language (like C# or Python) and the method name does not match Java's naming conventions.
```

rynowak (approved):

Issue #438, opened 2021-01-13
Handle service invocation over HTTP only


artursouza:
## Describe the proposal

Today, mapping gRPC to HTTP is not correct and we don't have a mapping that can work yet (as per https://github.com/dapr/dapr/issues/2342). So, all API calls should be made over HTTP by default in the SDK, even if the other APIs work over gRPC.

This should improve experience for most users since HTTP endpoint is most common. The gRPC to gRPC or HTTP to gRPC scenarios are way less common since dapr requires a custom proto to be implemented in the app.

Parent: https://github.com/dapr/dapr/issues/2342

PR #450, opened 2021-01-23
Makes service method API invocations over HTTP by default.

artursouza:
# Description

Makes service method API invocations over HTTP by default.

## Issue reference

Please reference the issue this PR will close: #438 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
@@ -321,6 +321,10 @@ public class ActorRuntime implements Closeable {
    * @return GRPC managed channel or null.
    */
   private static ManagedChannel buildManagedChannel() {
+    if (Properties.API_PROTOCOL.get() != DaprApiProtocol.GRPC) {
+      return null;
+    }
+
     int port = Properties.GRPC_PORT.get();
     if (port <= 0) {
       throw new IllegalStateException("Invalid port.");
--- a/sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
@@ -7,6 +7,7 @@ package io.dapr.actors.runtime;
 
 import io.dapr.actors.ActorId;
 import io.dapr.actors.ActorTrace;
+import io.dapr.client.DaprApiProtocol;
 import io.dapr.client.DaprHttpBuilder;
 import io.dapr.config.Properties;
 import io.dapr.serializer.DaprObjectSerializer;
@@ -307,7 +308,7 @@ public class ActorRuntime implements Closeable {
    * @throws java.lang.IllegalStateException if any required field is missing
    */
   private static DaprClient buildDaprClient(ManagedChannel channel) {
-    if (Properties.USE_GRPC.get()) {
+    if (Properties.API_PROTOCOL.get() == DaprApiProtocol.GRPC) {
       return new DaprGrpcClient(channel);
     }
 
@@ -320,10 +321,6 @@ public class ActorRuntime implements Closeable {
    * @return GRPC managed channel or null.
    */
   private static ManagedChannel buildManagedChannel() {
-    if (!Properties.USE_GRPC.get()) {
-      return null;
-    }
-
     int port = Properties.GRPC_PORT.get();
     if (port <= 0) {
       throw new IllegalStateException("Invalid port.");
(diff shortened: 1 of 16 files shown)


sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
```diff
@@ -97,11 +103,30 @@ public DaprClientBuilder withStateSerializer(DaprObjectSerializer stateSerialize
...
+   */
+  private DaprClient buildDaprClient(DaprApiProtocol protocol) {
+    if (protocol == null) {
+      throw new IllegalArgumentException("Protocol is required.");
+    }
+
+    switch (protocol) {
+      case GRPC: return buildDaprClientGrpc();
+      case HTTP: return buildDaprClientHttp();
+      default: throw new IllegalArgumentException("Unsupported protocol: " + protocol.name());
```
mukundansundar, line 128: Why is this exception different and worded differently from the exception in `ActorProxyBuilder`?
There it was:
```java
switch (this.apiProtocol) {
      return new DaprGrpcClient(DaprGrpc.newFutureStub(this.channel));	      case GRPC: return new DaprGrpcClient(DaprGrpc.newFutureStub(this.channel));
      case HTTP: return new DaprHttpClient(daprHttpBuilder.build());
      default: throw new IllegalStateException("Invalid Dapr API protocol.");
}
```

sdk/src/main/java/io/dapr/config/Properties.java
```diff
@@ -24,14 +26,19 @@
...
-   * Dapr's default GRPC port.
+   * Dapr's default use of gRPC or HTTP.
+   */
+  private static final DaprApiProtocol DEFAULT_API_PROTOCOL = DaprApiProtocol.GRPC;
+
+  /**
+   * Dapr's default use of gRPC or HTTP for Dapr's method invocation APIs.
    */
-  private static final Boolean DEFAULT_GRPC_ENABLED = true;
+  private static final DaprApiProtocol DEFAULT_API_METHODINVOCATION_PROTOCOL = DaprApiProtocol.HTTP;
```
mukundansundar, line 41:
```suggestion
  private static final DaprApiProtocol DEFAULT_API_METHOD_INVOCATION_PROTOCOL = DaprApiProtocol.HTTP;

```

sdk/src/main/java/io/dapr/config/Properties.java
```diff
@@ -68,12 +75,22 @@
...
+
+  /**
+   * Determines if Dapr client should use gRPC or HTTP for Dapr's service method invocation APIs.
    */
-  public static final Property<Boolean> USE_GRPC = new BooleanProperty(
-      "dapr.grpc.enabled",
-      "DAPR_GRPC_ENABLED",
-      DEFAULT_GRPC_ENABLED);
+  public static final Property<DaprApiProtocol> API_METHODINVOCATION_PROTOCOL = new GenericProperty<>(
+      "dapr.api.methodInvocation.protocol",
```
mukundansundar, line 90:
```suggestion
      "dapr.api.method.invocation.protocol",

```

mukundansundar (commented):
I can understand why you left it as methodInvocation in the propery and METHODINVOCATION in the env var, but I think method.invocation and METHOD_INVOCATION looks better and follows convention.

sdk/src/main/java/io/dapr/config/Properties.java
artursouza: This is wrong because method is not a child of api.

sdk/src/main/java/io/dapr/config/Properties.java
artursouza: I can make this change simply because it is internal and reads as per JAVA convention.

sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
artursouza: OK. Made them the same. IllegalState because it is the state of the builder that does not depend on a param.

sdk/src/main/java/io/dapr/config/Properties.java
artursouza: I also renamed another one and renamed the env variables.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -320,10 +321,6 @@ private static DaprClient buildDaprClient(ManagedChannel channel) {
    * @return GRPC managed channel or null.
    */
   private static ManagedChannel buildManagedChannel() {
-    if (!Properties.USE_GRPC.get()) {
```
mukundansundar, line 323: why is this removed?

artursouza: Actually, this is needed in the case actors communicate over HTTP. Good catch.

mukundansundar (approved):

PR #451, opened 2021-01-24
Use DaprChannel instead of closeable ActorProxyBuilder.

artursouza:
# Description

This change creates a class called `DaprChannel` for actor communication in Dapr. It will avoid making `ActorProxyBuilder` a closeable class - it was odd to have a builder class that could not be disposed. Now, `DaprChannel` is the class where users must keep a reference and close in the end. If we decide to use HTTP instead of gRPC, the DaprChannel still holds the Client reference, making it an effective singleton.

Summary:
+ An instance of `ActorProxyBuilder` is now disposable and does not need to exist while DaprClient is being used.
+ `DaprChannel` holds the "channel" to Dapr and must be referenced while app is using Dapr Actor APIs.
+ `DaprChannel` is closeable, so users must close it after done with using Dapr Actor APIs.
+ `DaprChannel` is encourage to be a singleton in the app, creating many instances can trash the application.
+ `DaprChannel` is specific to Actor APIs because `DaprClient` in the regular SDKs is already closeable and must be reused.

## Issue reference

Please reference the issue this PR will close: None

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

RELEASE NOTE: **REFACTOR** Created `DaprChannel` for actor APIs, making `ActorProxyBuilder` not closeable anymore.

### DIFF
--- a/examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
+++ b/examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
@@ -6,8 +6,8 @@
 package io.dapr.examples.actors;
 
 import io.dapr.actors.ActorId;
+import io.dapr.actors.client.ActorClient;
 import io.dapr.actors.client.ActorProxyBuilder;
-import io.dapr.actors.client.DaprChannel;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -31,8 +31,8 @@ public class DemoActorClient {
    * @throws InterruptedException If program has been interrupted.
    */
   public static void main(String[] args) throws InterruptedException {
-    try (DaprChannel channel = new DaprChannel()) {
-      ActorProxyBuilder<DemoActor> builder = new ActorProxyBuilder(DemoActor.class, channel);
+    try (ActorClient client = new ActorClient()) {
+      ActorProxyBuilder<DemoActor> builder = new ActorProxyBuilder(DemoActor.class, client);
       List<Thread> threads = new ArrayList<>(NUM_ACTORS);
 
       // Creates multiple actors.
--- a/examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
+++ b/examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
@@ -7,6 +7,7 @@ package io.dapr.examples.actors;
 
 import io.dapr.actors.ActorId;
 import io.dapr.actors.client.ActorProxyBuilder;
+import io.dapr.actors.client.DaprChannel;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -30,7 +31,8 @@ public class DemoActorClient {
    * @throws InterruptedException If program has been interrupted.
    */
   public static void main(String[] args) throws InterruptedException {
-    try (ActorProxyBuilder<DemoActor> builder = new ActorProxyBuilder(DemoActor.class)) {
+    try (DaprChannel channel = new DaprChannel()) {
+      ActorProxyBuilder<DemoActor> builder = new ActorProxyBuilder(DemoActor.class, channel);
       List<Thread> threads = new ArrayList<>(NUM_ACTORS);
 
       // Creates multiple actors.
--- a/sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
@@ -17,7 +17,7 @@ import reactor.core.publisher.Mono;
 /**
  * Holds a client for Dapr sidecar communication. ActorClient should be reused.
  */
-public class ActorClient implements DaprClient, AutoCloseable {
+public class ActorClient implements AutoCloseable {
 
   /**
    * gRPC channel for communication with Dapr sidecar.
@@ -56,10 +56,15 @@ public class ActorClient implements DaprClient, AutoCloseable {
   }
 
   /**
-   * {@inheritDoc}
+   * Invokes an Actor method on Dapr.
+   *
+   * @param actorType   Type of actor.
+   * @param actorId     Actor Identifier.
+   * @param methodName  Method name to invoke.
+   * @param jsonPayload Serialized body.
+   * @return Asynchronous result with the Actor's response.
    */
-  @Override
-  public Mono<byte[]> invoke(String actorType, String actorId, String methodName, byte[] jsonPayload) {
+  Mono<byte[]> invoke(String actorType, String actorId, String methodName, byte[] jsonPayload) {
     return daprClient.invoke(actorType, actorId, methodName, jsonPayload);
   }
 
--- a/sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
@@ -12,16 +12,12 @@ import io.dapr.v1.DaprGrpc;
 import io.grpc.Channel;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
+import reactor.core.publisher.Mono;
 
 /**
- * Holds a channel for Dapr sidecar communication. Channel should be reused.
+ * Holds a client for Dapr sidecar communication. ActorClient should be reused.
  */
-public class DaprChannel implements AutoCloseable {
-
-  /**
-   * Determine if channel is for GRPC clients or HTTP clients.
-   */
-  private final DaprApiProtocol apiProtocol;
+public class ActorClient implements DaprClient, AutoCloseable {
 
   /**
    * gRPC channel for communication with Dapr sidecar.
@@ -36,7 +32,7 @@ public class DaprChannel implements AutoCloseable {
   /**
    * Instantiates a new channel for Dapr sidecar communication.
    */
-  public DaprChannel() {
+  public ActorClient() {
     this(Properties.API_PROTOCOL.get());
   }
 
@@ -45,8 +41,8 @@ public class DaprChannel implements AutoCloseable {
    *
    * @param apiProtocol    Dapr's API protocol.
    */
-  private DaprChannel(DaprApiProtocol apiProtocol) {
-    this(apiProtocol,  buildManagedChannel(apiProtocol));
+  private ActorClient(DaprApiProtocol apiProtocol) {
+    this(apiProtocol, buildManagedChannel(apiProtocol));
   }
 
   /**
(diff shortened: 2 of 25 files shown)


examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
```diff
@@ -30,7 +31,8 @@
    * @throws InterruptedException If program has been interrupted.
    */
   public static void main(String[] args) throws InterruptedException {
-    try (ActorProxyBuilder<DemoActor> builder = new ActorProxyBuilder(DemoActor.class)) {
+    try (DaprChannel channel = new DaprChannel()) {
```
artursouza, line 34: This is why this is a breaking change. This is only applicable for Actors because the proxy is per actor instance, so there must be a "client" shared some how. I think having this separate class is more intuitive then making the ActorProxyBuilder to own the channel behind the scenes.

rynowak: Two questions about this approach...

---

I assume that it's not feasible to make the generated proxies (`actor` in this example) implement `Closable` because it's an interface that the user provided and may not include it.  

So having the user do the equivalent of (my Java is rusty, be nice!):

```java
if (actor instanceof Closable) {
    ((Closable)actor).Close()
}
```

Just sounds terrible from a usability standpoint. A beginner would not understand what to do.

Equally bad is requiring the actor interface to implement `Closable` because then they end up with a *do-nothing* implementation of that interface on the server, since the proxy interface is shared.

---

Second question.

Have you considered using a static instance of `DaprChannel` and making the constructor arg optional via overloading? This way by default users can share a single channel instance that's process-scoped and can opt-in to more complexity if they want eager resource cleanup.

---

I also considered the approach of returning a *handle* type, which would not only be a massive breaking change, but it would also require ref-counting (yuck).

sdk-actors/src/main/java/io/dapr/actors/client/DaprChannel.java
```diff
@@ -0,0 +1,114 @@
...
+import io.dapr.config.Properties;
+import io.dapr.v1.DaprGrpc;
+import io.grpc.Channel;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+
+/**
+ * Holds a channel for Dapr sidecar communication. Channel should be reused.
+ */
+public class DaprChannel implements AutoCloseable {
```
rynowak, line 19: Would a better name for this be `ActorChannel`? or something similar since it's scoped to actors not dapr as a whole.

sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyBuilder.java
```diff
@@ -132,7 +105,7 @@ public T build(ActorId actorId) {
             this.actorType,
             actorId,
             this.objectSerializer,
-            buildDaprClient());
+            this.channel.getDaprClient());
```
rynowak, line 108: If the channel just calls `getDaprClient()` is there a ready why it needs to be a separate class and not **just** a dapr client?

rynowak (commented):
Actual code changes look straightforward, left a few questions ...

sdk-actors/src/main/java/io/dapr/actors/client/DaprChannel.java
artursouza: It is scoped for Actors because it is in the `io.dapr.actors.client` package.

artursouza: We have the same for the `DaprClient` interface where we have on in `io.dapr.actors.client` and one in `io.dapr.client` and they are different.

sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyBuilder.java
artursouza: At this point, it makes sense to make it simply implement the DaprClient interface and forward the calls to the actual client implementation underneath. It reads better and gives sense of simplicity to users: `new ActorClient()`. I also renamed it to `ActorClient` as I think it is the best name for this class right now. `DaprClient` (No `I` in Java's naming convention here) is the interface and I did not want to use `DaprClientImpl`.

examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
artursouza: First question, making the proxy or the class user provided to be closeable does not make sense because the resource to be closed (grpcChannel) is shared and reused. So a single instance of ActorProxy cannot close it because there might be other ActorProxies still using it.

The idea is to give users freedom of where to create and when to close the managed resource. If I create one for them. Having a static instance will also load it even when users don't use it. Imagine having this managed channel created for the application that just serves the actor runtime without calling into other actors? Users are also used to do this in Springboot - actually being opinionated about resource management probably belongs to the springboot package we have instead of here.

No ref counting :)

sdk-actors/src/main/java/io/dapr/actors/client/DaprChannel.java
artursouza: Since it now implements the `DaprClient` interface, I decided to name it `ActorClient` similar to how we name `ActorProxy`. So, it should resolve this ambiguity now.

artursouza:
I made ActorClient not implement the DaprClient interface anymore since we don't want to expose that method to applications but made it implement a method with same signature and delegate to DaprClient instead. Now, apps only see the `close()` method for ActorClient which is a more desirable surface area.

sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
```diff
@@ -0,0 +1,113 @@
...
+import io.dapr.v1.DaprGrpc;
+import io.grpc.Channel;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import reactor.core.publisher.Mono;
+
+/**
+ * Holds a client for Dapr sidecar communication. ActorClient should be reused.
+ */
+public class ActorClient implements AutoCloseable {
```
mukundansundar, line 20: Have this implement `DaprClient` as well?

sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyBuilder.java
```diff
@@ -84,22 +56,23 @@ public ActorProxyBuilder(String actorType, Class<T> actorTypeClass) {
...
-  private ActorProxyBuilder(String actorType, Class<T> actorTypeClass, DaprApiProtocol apiProtocol) {
+  public ActorProxyBuilder(String actorType, Class<T> actorTypeClass, ActorClient actorClient) {
     if ((actorType == null) || actorType.isEmpty()) {
       throw new IllegalArgumentException("ActorType is required.");
     }
     if (actorTypeClass == null) {
       throw new IllegalArgumentException("ActorTypeClass is required.");
     }
+    if (actorClient == null) {
+      throw new IllegalArgumentException("Channel is required.");
```
mukundansundar, line 69:
```suggestion
      throw new IllegalArgumentException("ActorClient is required.");
```

sdk-actors/src/test/java/io/dapr/actors/client/ActorProxyBuilderTest.java
```diff
@@ -7,43 +7,56 @@
...
   @Test(expected = IllegalArgumentException.class)
   public void buildWithNullSerializer() {
-    new ActorProxyBuilder("MyActor", Object.class)
+    new ActorProxyBuilder("MyActor", Object.class, actorClient)
       .withObjectSerializer(null)
       .build(new ActorId("100"));
+  }
 
+  @Test(expected = IllegalArgumentException.class)
+  public void buildWithNullChannel() {
```
mukundansundar, line 53:
```suggestion
  public void buildWithNullClient() {
```

sdk-actors/src/test/java/io/dapr/actors/client/ActorProxyForTestsImpl.java
```diff
@@ -10,7 +10,7 @@
 
 public class ActorProxyForTestsImpl extends ActorProxyImpl {
```
mukundansundar, line 11: nitpick
```suggestion
public class ActorProxyImplForTests extends ActorProxyImpl {

```

sdk-actors/src/test/java/io/dapr/actors/client/ActorProxyImplTest.java
```diff
@@ -23,7 +23,7 @@
 
   @Test()
   public void constructorActorProxyTest() {
-    final DaprClient daprClient = mock(DaprClient.class);
+    final ActorClient daprClient = mock(ActorClient.class);
```
mukundansundar, line 26: my suggestion is let ActorClient implement DaprClient interface.

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
```diff
@@ -26,26 +26,6 @@
...
-  /**
-   * Sets the read timeout duration for the instance to be built.
-   *
-   * <p>Instead, set environment variable "DAPR_HTTP_CLIENT_READTIMEOUTSECONDS",
-   *   or system property "dapr.http.client.readtimeoutseconds".
-   *
-   * @param duration Read timeout duration.
-   * @return Same builder instance.
-   */
-  @Deprecated
```
mukundansundar, line 43: quick question ... why was this deprecated?

mukundansundar (commented):
left some comments.

sdk-actors/src/test/java/io/dapr/actors/client/ActorProxyImplTest.java
artursouza: I had this in a version but changed. The reason is that it would expose DaprClient method publicly while it is an internal interface. It would allow users to bypass the SDK and invoke the actor APIs directly without using ActorProxy.

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
artursouza: You cannot really set this per instance since the instance is singleton.

sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
artursouza: I had this in a version but it would expose the invokeActorMethod publicly while it is supposed to be internal do the ActorProxy's package.

mukundansundar: oh ok .

sdk-actors/src/test/java/io/dapr/actors/client/ActorProxyForTestsImpl.java
artursouza: Done.

examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java
rynowak: > Users are also used to do this in Springboot - actually being opinionated about resource management probably belongs to the springboot package we have instead of here.

So does this mean we have a recipe for using these types in a DI-enabled way? Or is that up the user to figure out?

rynowak (approved):

Issue #441, opened 2021-01-14
Why not implement java-sdk-springboot as a spring boot starter?


darren-wang:
## Describe the proposal
Hi, I'm a java learner of dapr, i wonder why not implement  dapr-sdk-springboot according to spring boot common starter rules, that would be nicer for spring boot users.

artursouza:
@darren-wang Can you provide an example of how it would work for the end user? I am not a spring boot expert, so I am eager to learn how we can better integrate with it.

darren-wang:
> 
> 
> @darren-wang Can you provide an example of how it would work for the end user? I am not a spring boot expert, so I am eager to learn how we can better integrate with it.

HI, please check PR:

https://cla.opensource.microsoft.com/dapr/java-sdk?pullRequest=457

artursouza:
@darren-wang We are closing on the release, so I took the liberty to have a PR where we incorporate your solution into  the sdk-springboot package directly. It is OK for that package to be springboot specific - that is the intent. We might refactor spring only stuff out of it later. The priority now is on springboot apps. Please, take a look a the PR.

darren-wang:
Closed by https://github.com/dapr/java-sdk/pull/457 and https://github.com/dapr/java-sdk/pull/459

artursouza:
Reopening since PR is not merged yet.

PR #457, opened 2021-01-26
Adding Dapr Spring Boot Starter.

darren-wang:
# Description

Adding a spring boot starter for dapr to better integrate with Java spring boot applicaitons.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/441

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #459, opened 2021-01-27
Adds autoconfiguration for springboot apps.

artursouza:
# Description

Adds autoconfiguration for springboot apps.

## Issue reference

Please reference the issue this PR will close: #441

Inspired by https://github.com/dapr/java-sdk/pull/457

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk-tests/src/test/java/io/dapr/it/actors/app/TestApplication.java
```diff
@@ -11,7 +11,7 @@
 /**
  * Dapr's HTTP callback implementation via SpringBoot.
  */
-@SpringBootApplication(scanBasePackages = {"io.dapr.springboot", "io.dapr.it.actors.app"})
```
mukundansundar, line 14: Why is `"io.dapr.it.actors.app"` no longer needed?

sdk-tests/src/test/java/io/dapr/it/pubsub/http/SubscriberService.java
```diff
@@ -12,7 +12,7 @@
 /**
  * Service for subscriber.
  */
-@SpringBootApplication(scanBasePackages = {"io.dapr.springboot", "io.dapr.it.pubsub.http"})
```
mukundansundar, line 15: Why is `"io.dapr.it.pubsub.http"` no longer needed?

artursouza: The package where the application is declared is included by default. The springboot path is added automatically now with this change.

sdk-tests/src/test/java/io/dapr/it/actors/app/TestApplication.java
artursouza: same answer as below.

sdk-springboot/pom.xml
```diff
@@ -29,8 +29,21 @@
...
+      <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-dependencies</artifactId>
+        <version>${springboot.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
```
darren-wang, line 46: Be careful of this segment, this will override `okhttp` version set in `dapr-sdk`, because spring boot also makes their version declaration in `spring-boot-depencies`, which conficts with dapr's.

sdk-springboot/src/main/java/io/dapr/springboot/DaprAutoConfiguration.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Dapr's Spring Boot AutoConfiguration.
+ */
+@Configuration
+@ConditionalOnWebApplication
+@ComponentScan("io.dapr.springboot")
+public class DaprAutoConfiguration {
+}
```
darren-wang, line 19: You can use `@ComponentScan` here, also, you can replace it with  `@Bean` method and `@ConditionalOnProperty` to provide finer grained control of beans, e.g. you can use property like `dapr.controller.enabled` to control whether to create the controller bean.

darren-wang (commented):
better with boot apps.

sdk-springboot/src/main/java/io/dapr/springboot/DaprAutoConfiguration.java
artursouza: Would this be a breaking change? If not, you could send a PR after this. You have more knowledge about spring boot than me.

sdk-springboot/pom.xml
artursouza: For this one I am OK since users would need to solve a version conflict too. If this becomes a problem we would know in our integration tests. Thanks for pointing this out. Learned one more thing.

mukundansundar (approved):

sdk-springboot/src/main/java/io/dapr/springboot/DaprAutoConfiguration.java
darren-wang: No, we can release this version this time, and make additional improments in the future.

artursouza: Thanks. That will also give you time to prepare a PR with your ideas. This one was great BTW. Thanks a lot.

Issue #461, opened 2021-01-28
Error handling for http invocation of gRPC service


artursouza:
## Expected Behavior

Error code is handled correctly when deserialized.

## Actual Behavior

Code crashes and Mono does not get a response.

## Steps to Reproduce the Problem

Invoke a gRPC endpoint that throws an exception, using the http client, you will not get a response back.

## Release Note

RELEASE NOTE: **FIX** error handling for service invocation with HTTP client and gRPC service.

PR #462, opened 2021-01-28
Fix error handling in http invoke.

artursouza:
# Description

Fix error handling in http invoke.
ITs to validate invoke errors on http and grpc services.
Handle serialization of proto objects.

## Issue reference

Please reference the issue this PR will close: #461 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/DaprRun.java
+++ b/sdk-tests/src/test/java/io/dapr/it/DaprRun.java
@@ -20,7 +20,7 @@ public class DaprRun implements Stoppable {
 
   private static final String DAPR_SUCCESS_MESSAGE = "You're up and running!";
 
-  private static final String DAPR_RUN = "dapr run --app-id %s --components-path ./components";
+  private static final String DAPR_RUN = "dapr run --app-id %s --app-protocol %s --components-path ./components";
 
   // the arg in -Dexec.args is the app's port
   private static final String DAPR_COMMAND =
@@ -45,11 +45,12 @@ public class DaprRun implements Stoppable {
                   String successMessage,
                   Class serviceClass,
                   int maxWaitMilliseconds,
-                  DaprApiProtocol protocol) {
+                  DaprApiProtocol protocol,
+                  DaprApiProtocol appProtocol) {
     // The app name needs to be deterministic since we depend on it to kill previous runs.
     this.appName = serviceClass == null ? testName : String.format("%s_%s", testName, serviceClass.getSimpleName());
     this.startCommand =
-        new Command(successMessage, buildDaprCommand(this.appName, serviceClass, ports, protocol));
+        new Command(successMessage, buildDaprCommand(this.appName, serviceClass, ports, protocol, appProtocol));
     this.listCommand = new Command(
       this.appName,
       "dapr list");
@@ -170,16 +171,17 @@ public class DaprRun implements Stoppable {
   }
 
   private static String buildDaprCommand(
-      String appName, Class serviceClass, DaprPorts ports, DaprApiProtocol protocol) {
-    StringBuilder stringBuilder = new StringBuilder(String.format(DAPR_RUN, appName))
-        .append(ports.getAppPort() != null ? " --app-port " + ports.getAppPort() : "")
-        .append(ports.getHttpPort() != null ? " --dapr-http-port " + ports.getHttpPort() : "")
-        .append(ports.getGrpcPort() != null ? " --dapr-grpc-port " + ports.getGrpcPort() : "")
-        .append(serviceClass == null ? "" :
-            String.format(DAPR_COMMAND, serviceClass.getCanonicalName(),
-                ports.getAppPort() != null ? ports.getAppPort().toString() : "",
-                Properties.API_PROTOCOL.getName(), protocol,
-                Properties.API_METHOD_INVOCATION_PROTOCOL.getName(), protocol));
+      String appName, Class serviceClass, DaprPorts ports, DaprApiProtocol protocol, DaprApiProtocol appProtocol) {
+    StringBuilder stringBuilder =
+        new StringBuilder(String.format(DAPR_RUN, appName, appProtocol.toString().toLowerCase()))
+            .append(ports.getAppPort() != null ? " --app-port " + ports.getAppPort() : "")
+            .append(ports.getHttpPort() != null ? " --dapr-http-port " + ports.getHttpPort() : "")
+            .append(ports.getGrpcPort() != null ? " --dapr-grpc-port " + ports.getGrpcPort() : "")
+            .append(serviceClass == null ? "" :
+                String.format(DAPR_COMMAND, serviceClass.getCanonicalName(),
+                    ports.getAppPort() != null ? ports.getAppPort().toString() : "",
+                    Properties.API_PROTOCOL.getName(), protocol,
+                    Properties.API_METHOD_INVOCATION_PROTOCOL.getName(), protocol));
     return stringBuilder.toString();
   }
 
@@ -210,17 +212,21 @@ public class DaprRun implements Stoppable {
 
     private DaprApiProtocol protocol;
 
+    private DaprApiProtocol appProtocol;
+
     Builder(
         String testName,
         Supplier<DaprPorts> portsSupplier,
         String successMessage,
         int maxWaitMilliseconds,
-        DaprApiProtocol protocol) {
+        DaprApiProtocol protocol,
+        DaprApiProtocol appProtocol) {
       this.testName = testName;
       this.portsSupplier = portsSupplier;
       this.successMessage = successMessage;
       this.maxWaitMilliseconds = maxWaitMilliseconds;
       this.protocol = protocol;
+      this.appProtocol = appProtocol;
     }
 
     public Builder withServiceClass(Class serviceClass) {
@@ -235,7 +241,8 @@ public class DaprRun implements Stoppable {
               this.successMessage,
               this.serviceClass,
               this.maxWaitMilliseconds,
-              this.protocol);
+              this.protocol,
+              this.appProtocol);
     }
 
     /**
@@ -257,7 +264,8 @@ public class DaprRun implements Stoppable {
               DAPR_SUCCESS_MESSAGE,
               null,
               this.maxWaitMilliseconds,
-              this.protocol);
+              this.protocol,
+              this.appProtocol);
 
       return new ImmutablePair<>(appRun, daprRun);
     }
--- a/sdk-tests/src/test/java/io/dapr/it/DaprRun.java
+++ b/sdk-tests/src/test/java/io/dapr/it/DaprRun.java
@@ -140,18 +140,28 @@ public class DaprRun implements Stoppable {
       System.getProperties().setProperty(Properties.GRPC_PORT.getName(), String.valueOf(this.ports.getGrpcPort()));
     }
     System.getProperties().setProperty(Properties.API_PROTOCOL.getName(), DaprApiProtocol.GRPC.name());
+    System.getProperties().setProperty(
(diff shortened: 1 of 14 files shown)


sdk/src/main/java/io/dapr/client/ObjectSerializer.java
```diff
@@ -54,6 +56,11 @@ protected ObjectSerializer() {
       return (byte[]) state;
     }
 
+    // Proto buffer class is serialized directly.
+    if (state instanceof MessageLite) {
+      return ((MessageLite) state).toByteArray();
```
artursouza, line 61: This allows this to serialize proto objects.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
```diff
@@ -103,13 +110,24 @@ protected ObjectSerializer() {
...
       return (T) CloudEvent.deserialize(content);
     }
 
+    if (javaType.isTypeOrSubTypeOf(MessageLite.class)) {
+      try {
+        Method method = javaType.getRawClass().getDeclaredMethod("parseFrom", byte[].class);
+        if (method != null) {
+          return (T) method.invoke(null, content);
+        }
+      } catch (ReflectiveOperationException e) {
```
artursouza, line 126: Let me make this more strict.

sdk/src/main/java/io/dapr/exceptions/DaprError.java
```diff
@@ -20,12 +24,20 @@
    */
   private String message;
 
+  /**
+   * Error code from gRPC.
+   */
+  private Integer code;
```
artursouza, line 30: This is needed because the resulting JSON is different for a gRPC service.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
artursouza: Done.

sdk-tests/src/test/java/io/dapr/it/DaprRun.java
```diff
@@ -170,16 +171,17 @@ public String getAppName() {
...
+    StringBuilder stringBuilder =
+        new StringBuilder(String.format(DAPR_RUN, appName, appProtocol.toString().toLowerCase()))
+            .append(ports.getAppPort() != null ? " --app-port " + ports.getAppPort() : "")
+            .append(ports.getHttpPort() != null ? " --dapr-http-port " + ports.getHttpPort() : "")
+            .append(ports.getGrpcPort() != null ? " --dapr-grpc-port " + ports.getGrpcPort() : "")
+            .append(serviceClass == null ? "" :
+                String.format(DAPR_COMMAND, serviceClass.getCanonicalName(),
+                    ports.getAppPort() != null ? ports.getAppPort().toString() : "",
+                    Properties.API_PROTOCOL.getName(), protocol,
+                    Properties.API_METHOD_INVOCATION_PROTOCOL.getName(), protocol));
```
mukundansundar, line 184: should this be appProtocol?

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
```diff
@@ -0,0 +1,145 @@
...
+                GetMessagesRequest.newBuilder().build(),
+                HttpExtension.POST, GetMessagesResponse.class).block().getMessagesMap();
+            assertEquals(10, messages.size());
+
+            // Delete one message.
+            client.invokeMethod(
+                daprRun.getAppName(),
+                "deleteMessage",
+                DeleteMessageRequest.newBuilder().setId(1).build(),
+                HttpExtension.POST).block();
```
mukundansundar, line 97: `HTTPExtension.DELETE` ?

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
```diff
@@ -0,0 +1,145 @@
...
+                PostMessageRequest req = PostMessageRequest.newBuilder().setId(i).setMessage(message).build();
+                client.invokeMethod(daprRun.getAppName(), "postMessage", req, HttpExtension.POST).block();
+                System.out.println("Invoke method messages : " + message);
+            }
+
+            Map<Integer, String> messages = client.invokeMethod(
+                daprRun.getAppName(),
+                "getMessages",
+                GetMessagesRequest.newBuilder().build(),
+                HttpExtension.POST, GetMessagesResponse.class).block().getMessagesMap();
```
mukundansundar, line 89: `HttpExtension.GET` ?

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Test
+    public void testInvokeException() throws Exception {
+        try (DaprClient client = new DaprClientBuilder().build()) {
+            SleepRequest req = SleepRequest.newBuilder().setSeconds(-9).build();
+            DaprException exception = assertThrows(DaprException.class, () ->
+                client.invokeMethod(daprRun.getAppName(), "sleep", req.toByteArray(), HttpExtension.POST).block());
+
+            assertEquals("UNKNOWN", exception.getErrorCode());
+            assertEquals("UNKNOWN: HTTP status code: 500", exception.getMessage());
```
mukundansundar, line 142: is this because the method invocation is always `http` ? 
if it is gRPC why is the error message saying HTTP.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
mukundansundar: I think this is similar to how python handles gRPC messages too.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
```diff
@@ -103,13 +110,24 @@ protected ObjectSerializer() {
...
     if (content.length == 0) {
-      return (T) null;
+      return null;
     }
 
     if (javaType.hasRawClass(CloudEvent.class)) {
       return (T) CloudEvent.deserialize(content);
     }
 
+    if (javaType.isTypeOrSubTypeOf(MessageLite.class)) {
```
mukundansundar, line 120: add a comment on why this clause is required?

is there anything that the sub type needs to implement or is it automatically implemented? for example the `parseFrom` method? Where is it coming from? Is it guaranteed that all types that satisfy this clause will have a `parseFrom` method if not we will be throwing an exception? 

Should this be documented separately?

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
mukundansundar: Should his be documented somewhere?

sdk/src/main/java/io/dapr/exceptions/DaprError.java
mukundansundar: How is this field set? 
I think this is set on json parsing?

sdk/src/main/java/io/dapr/exceptions/DaprError.java
```diff
@@ -5,9 +5,13 @@
 
 package io.dapr.exceptions;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import io.grpc.Status;
+
 /**
  * Represents an error message from Dapr.
  */
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
```
mukundansundar, line 14: Will it be better to have getter/setter for code instead of this annotation? To make it consistent?

sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
```diff
@@ -189,13 +190,80 @@ public void invokeServiceIllegalArgumentException() {
...
+
+    assertEquals("MYCODE", exception.getErrorCode());
+    assertEquals("MYCODE: My Message", exception.getMessage());
+  }
+
+  @Test
+  public void invokeServiceDaprErrorFromGRPC() {
+    mockInterceptor.addRule()
+        .post("http://127.0.0.1:3000/v1.0/invoke/myapp/method/mymethod")
+        .respond(500,
```
mukundansundar, line 214: please have 403 to match gRPC code.

sdk-tests/src/test/java/io/dapr/it/DaprRun.java
artursouza: No, this is for the method API protocol, which is which protocol we use when invoking a service.

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
artursouza: gRPC does not have a notion of HTTP method.

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
artursouza: gRPC does not have a notion of HTTP method.

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
artursouza: Let me check. It should not be since I am testing with gRPC client too.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
artursouza: Let me check with another class. Not all MessageLite will have this. Since we invoke a static method, there is no interface for that.

sdk/src/main/java/io/dapr/exceptions/DaprError.java
artursouza: The internal property is private. I don't want to expose that on purpose.

sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
artursouza: Actually, 500 has nothing to do with the http gRPC status code. We do not match that at all.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
artursouza: Keeping this class will maintain consistency with the toByteArray() from above. This is a best effort since we check for null. I fixed an issue to handle MethodNotFoundException here, so it will proceed with the usual path.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
artursouza: I will work on documentation later. This SDK actually does not have a nice doc website like the python SDK - yet.

sdk/src/main/java/io/dapr/exceptions/DaprError.java
artursouza: It is set on JSON parsing. I found out about this in ITs.

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
artursouza: Fixed. Our ITs where not switching to gRPC.

mukundansundar (approved):

mukundansundar:
We need to be really careful on upgrading dependencies ...

Issue #453, opened 2021-01-25
Remove dependency from OpenTelemetry SDK - still unstable


xiazuojie:
## UPDATE

Remove dependency from OpenTelemetry SDK completely, a consequence of that is the fix described below.

Keep OpenTelemetrySDK in example only.

## Describe the proposal
Currently, ``io.dapr.client.domain.Response`` is in a lot of State APIs.
```java
<T> Mono<Response<List<State<T>>>> getBulkState(GetBulkStateRequest request, TypeRef<T> type);

Mono<Response<Void>> saveBulkState(SaveStateRequest request);
```
The sole purpose of a Response is to hold a ``io.opentelemetry.context.Context`` and the object. However, project reactor already has an integrated [reactor.util.context.Context](https://projectreactor.io/docs/core/release/reference/#context), which serves the same purpose of ``io.opentelemetry.context.Context``, i.e.  carry scoped-values across API boundaries and between threads.

I suggest remove ``io.dapr.client.domain.Response`` altogether and use ``reactor.util.context.Context``.

artursouza:
@xiazuojie This started an interesting discussion. Our SDK depends on OpenTelemetry v0.10.0. Since OpenTelemetry is not stable yet, it is not compatible with version 0.14.0. It also means that we force apps to use the same OpenTelemetry SDK version as we do - this is a problem for us to call our SDK "stable". We will work on this issue with urgency but the scope is bigger: we want to remove dependency from OpenTelemetry SDK.

artursouza:
@xiazuojie I am about to create a PR removing opentelemetry and opencensus as a dependency for the sdk (examples will still use that). I wonder if the Response object can actually be removed. I was looking into how to do it with Reactor only but it seems that the Context must be set *after* the reader of the Context in the chain. It is odd, I am not sure why. I reproduced it with the example discussed here: https://stackoverflow.com/questions/54779831/mono-subscribercontext-returning-old-context

it means that it would not be possible because it would require applications to pass in their mono object as an input (see example above).

In the mean time, I will keep Response class but with the reactor's Context instead.

Removing Response would still be a breaking change if we decide to go that way. Please, let me know if you have an idea on how to remove the Response class and use reactor only.

artursouza:
@xiazuojie We closed this issue since we removed dependency from OpenTelemetry and OpenCensus SDKs. See my comment above regarding removing `Response` object. I am happy to reopen this issue and discuss it further.

xiazuojie:
[...]

PR #463, opened 2021-01-29
Removing OpenTelemetry and OpenCensus SDK.

artursouza:
# Description

Removing OpenTelemetry and OpenCensus SDK.

## Issue reference

Please reference the issue this PR will close: #453 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -24,7 +24,7 @@
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <maven.deploy.skip>true</maven.deploy.skip>
     <spotbugs.fail>false</spotbugs.fail>
-    <opentelemetry.version>0.10.0</opentelemetry.version>
+    <opentelemetry.version>0.14.0</opentelemetry.version>
   </properties>
 
   <dependencies>
--- a/examples/src/main/java/io/dapr/examples/OpenTelemetryConfig.java
+++ b/examples/src/main/java/io/dapr/examples/OpenTelemetryConfig.java
@@ -6,13 +6,18 @@
 package io.dapr.examples;
 
 import io.dapr.examples.invoke.http.InvokeClient;
+import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
-import io.opentelemetry.api.trace.propagation.HttpTraceContext;
-import io.opentelemetry.context.propagation.DefaultContextPropagators;
+import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.context.propagation.TextMapPropagator;
 import io.opentelemetry.exporter.zipkin.ZipkinSpanExporter;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration;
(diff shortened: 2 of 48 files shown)


examples/pom.xml
```diff
@@ -24,7 +24,7 @@
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <maven.deploy.skip>true</maven.deploy.skip>
     <spotbugs.fail>false</spotbugs.fail>
-    <opentelemetry.version>0.10.0</opentelemetry.version>
+    <opentelemetry.version>0.14.0</opentelemetry.version>
```
artursouza, line 27: I am keeping opentelemetry for example only. I am also upgrading it to the latest version.

examples/src/main/java/io/dapr/examples/OpenTelemetryConfig.java
```diff
@@ -57,13 +62,45 @@ public static Tracer createTracer(String instrumentationName) {
...
+  public static reactor.util.context.Context getReactorContext() {
+    return getReactorContext(Context.current());
+  }
+
+  /**
+   * Converts given OpenTelemetry's context into Reactor's context.
+   * @param context OpenTelemetry's context.
+   * @return Reactor's context.
+   */
+  public static reactor.util.context.Context getReactorContext(Context context) {
```
artursouza, line 93: This logic can potentially be exported into a separate package to be published: sdk-opentelemetry. It would contain utility classes to use opentelemetry with dapr. We can do that once opentelemetry is stable.

sdk/pom.xml
```diff
@@ -52,16 +52,6 @@
       <artifactId>okhttp</artifactId>
       <version>4.9.0</version>
     </dependency>
-    <dependency>
-      <groupId>io.opentelemetry</groupId>
-      <artifactId>opentelemetry-api</artifactId>
-      <version>0.10.0</version>
```
artursouza, line 58: Yeah!

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -73,18 +62,6 @@
  */
 public class DaprClientGrpc extends AbstractDaprClient {
 
-  private static final TextMapPropagator.Setter<Map<String, String>> MAP_SETTER =
```
artursouza, line 76: I moved all the logic for tracing out of this class.

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -51,12 +49,6 @@
    */
   private static final String DEFAULT_HTTP_SCHEME = "http";
 
-  /**
-   * Sets the headers for OpenTelemetry SDK.
-   */
-  private static final HttpTraceContext.Setter<Request.Builder> OPENTELEMETRY_SETTER =
```
artursouza, line 57: Reactor's context is easier to work with.

sdk/src/main/java/io/dapr/internal/opencensus/BigendianEncoding.java
```diff
@@ -0,0 +1,156 @@
...
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.internal.opencensus;
+
+import java.util.Arrays;
+
+/**
+ * Code originally from https://github.com/census-instrumentation/opencensus-java/blob/
+ * 446e9bde9b1f6c0317e3f310644997e5d6d5eab2/api/src/main/java/io/opencensus/trace/BigendianEncoding.java
```
artursouza, line 12: Most of the files in this package are from opencensus. I copy and pasted the classes I needed, removed the methods that were unused and did some formatting. I also made almost all method to be package accessible. As per @yaron2, we should refer to the origin of the code but still use our license.

sdk/src/main/java/io/dapr/internal/opencensus/GrpcWrapper.java
```diff
@@ -0,0 +1,113 @@
...
+
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+/**
+ * Wraps a Dapr gRPC stub with telemetry interceptor.
+ */
+public final class GrpcWrapper {
```
artursouza, line 27: This is the only class in this package that is 100% written by us. I kept it here because it allows me to restrict access to most methods in this package.

wcs1only (approved):
I must state for the record, that I hate deep copying, but if taking a dependency is untenable, so be it.

artursouza:
> I must state for the record, that I hate deep copying, but if taking a dependency is untenable, so be it.

This is a common technique in the Java world to avoid conflicting dependencies with consuming apps. In this case, an unstable dependency is much worse because minor upgrades can be breaking changes.

wcs1only (approved):

PR #464, opened 2021-01-30
Setup auto validation of java examples

mukundansundar:
# Description

Added auto validation based on manual trigger for now. 

Current Actors example code is not deterministic enough for auto validation.
Also tracing is also not deterministic for validation.

Example successful validation build: https://github.com/mukundansundar/java-sdk/runs/1796635852?check_suite_focus=true

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/validate.yml
```diff
@@ -0,0 +1,123 @@
+# ------------------------------------------------------------
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# ------------------------------------------------------------
+
+name: Auto Validate Examples
+
+on:
+  workflow_dispatch:
```
mukundansundar, line 9: For now manual trigger. Can change later.

wcs1only: I have it set to run on every push for python-sdk, because why not?

wcs1only (approved):
Love it!

.github/workflows/validate.yml
artursouza: Add to pull requests and pushes. This is not different from our ITs.

artursouza: Let me fix that.

artursouza: Done.

artursouza (approved):

PR #485, opened 2021-02-10
Support many query strings in HTTP and binary data in Cloud Event for PubSub.

artursouza:
# Description

## Issue reference

Please reference the issue this PR will close: #484 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/domain/HttpExtension.java
+++ b/sdk/src/main/java/io/dapr/client/domain/HttpExtension.java
@@ -6,9 +6,12 @@
 package io.dapr.client.domain;
 
 import io.dapr.client.DaprHttp;
+import okhttp3.HttpUrl;
 
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * HTTP Extension class.
@@ -60,9 +63,9 @@ public final class HttpExtension {
   private DaprHttp.HttpMethods method;
 
   /**
-   * HTTP querystring.
+   * HTTP query params.
    */
-  private Map<String, String> queryString;
+  private Map<String, List<String>> queryParams;
 
   /**
    * HTTP headers.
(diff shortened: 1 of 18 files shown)


sdk/src/main/java/io/dapr/client/domain/HttpExtension.java
```diff
@@ -101,11 +106,31 @@ public HttpExtension(DaprHttp.HttpMethods method) {
...
+   * @return Encoded HTTP query string.
+   */
+  public String encodeQueryString() {
+    if ((this.queryParams == null) || (this.queryParams.isEmpty())) {
+      return "";
+    }
+
+    HttpUrl.Builder urlBuilder = new HttpUrl.Builder();
+    // Setting required values but we only need query params in the end.
+    urlBuilder.scheme("http").host("localhost");
```
mukundansundar, line 128: should this be `Properties. SIDECAR_IP` ?

sdk/src/test/java/io/dapr/client/CloudEventTest.java
```diff
@@ -149,4 +149,26 @@ public void deserializeBytes() throws Exception {
     assertEquals("AQI=", cloudEvent.getData());
     assertArrayEquals(expected, OBJECT_MAPPER.convertValue(cloudEvent.getData(), byte[].class));
   }
+
```
mukundansundar, line 152: with the change for bytes as data_base64 line 142 should never happen right? can the test be removed?

mukundansundar (commented):
added some comments.

sdk/src/main/java/io/dapr/client/domain/HttpExtension.java
artursouza: See comment right above this line. It is unused, just required by HttpUrl.Builder()

sdk/src/test/java/io/dapr/client/CloudEventTest.java
artursouza: Actually, it can happen. As long as the application sends data as JSON. It is perfectly fine. But app will need to deserialize the string on the way back.

mukundansundar: ok good to know.

mukundansundar (approved):

Issue #491, opened 2021-02-15
Add support to publish CloudEvent payload


artursouza:
Publishing a CloudEvent payload requires `application/cloudevents+json` content-type so dapr sidecar does not wrap it in another cloud event. This issue is to allow the content-type to be set at the publishing time, so this and other scenarios can be enabled where the content-type need to be modified at a per request basis.

RELEASE NOTE: **ADDED** support to publish CloudEvent payload (and others) with correct content-type per event.

PR #492, opened 2021-02-16
Allows override of content-type in pubsub publish.

artursouza:
# Description

## Issue reference

Please reference the issue this PR will close: #491

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


yaron2 (commented):
Do we need docs changes that make it clear which content types are accepted and their use case?

artursouza:
> Do we need docs changes that make it clear which content types are accepted and their use case?

We have changed the PubSub example for that.

yaron2 (approved):
lgtm

Issue #524, opened 2021-03-23
Use Mono.empty() instead of a State with null value (when a state does not exist)


xiazuojie:
### Non-existing value
Project reactor does not allow null values. It has the concept of an **empty** stream to represent a non-existing value. It has the benefit of being null-safe.

When a state does not exist, current state API does not return an empty Mono. Instead, it returns a State object with its value set to null. It is not null-safe. It requires the client to do null check as follows:
```java
        GetStateRequest request = new GetStateRequestBuilder("example-store", "test-key").build();
        Mono<State<String>> state = daprClient.getState(request, TypeRef.STRING).flatMap(
                it -> Mono.justOrEmpty(it.getValue()) // boilerplate code to check null
        );
```

I suggest ``getState()`` API returns ``Mono.empty()`` when a state does not exist.

artursouza:
Closing this issue as discussed in the PR. Thanks.

PR #525, opened 2021-03-23
Use Mono.empty() instead of a State with null value

xiazuojie:
# Description

Use Mono.empty() instead of a State with null value

## Issue reference

Please reference the issue this PR will close: #524 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
@@ -283,7 +283,7 @@ public class DaprClientGrpc extends AbstractDaprClient {
               this.<DaprProtos.GetStateResponse>createMono(
                   it -> intercept(context, asyncStub).getState(envelope, it)
               )
-      ).map(
+      ).flatMap(
           it -> {
             try {
               return buildStateKeyValue(it, key, options, type);
@@ -369,7 +369,7 @@ public class DaprClientGrpc extends AbstractDaprClient {
     return new State<>(key, value, etag, item.getMetadataMap(), null);
   }
 
-  private <T> State<T> buildStateKeyValue(
+  private <T> Mono<State<T>> buildStateKeyValue(
       DaprProtos.GetStateResponse response,
       String requestedKey,
       StateOptions stateOptions,
@@ -377,11 +377,14 @@ public class DaprClientGrpc extends AbstractDaprClient {
     ByteString payload = response.getData();
     byte[] data = payload == null ? null : payload.toByteArray();
     T value = stateSerializer.deserialize(data, type);
+    if (value == null) {
+      return Mono.empty();
+    }
     String etag = response.getEtag();
     if (etag.equals("")) {
       etag = null;
     }
-    return new State<>(requestedKey, value, etag, response.getMetadataMap(), stateOptions);
+    return Mono.just(new State<>(requestedKey, value, etag, response.getMetadataMap(), stateOptions));
   }
 
   /**
--- a/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
@@ -376,10 +376,10 @@ public class DaprClientGrpc extends AbstractDaprClient {
       TypeRef<T> type) throws IOException {
     ByteString payload = response.getData();
     byte[] data = payload == null ? null : payload.toByteArray();
-    T value = stateSerializer.deserialize(data, type);
-    if (value == null) {
+    if (data == null || data.length == 0) {
       return Mono.empty();
     }
+    T value = stateSerializer.deserialize(data, type);
     String etag = response.getEtag();
     if (etag.equals("")) {
       etag = null;
--- a/sdk/src/main/java/io/dapr/client/DaprClientHttp.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClientHttp.java
@@ -350,7 +350,7 @@ public class DaprClientHttp extends AbstractDaprClient {
           .invokeApi(DaprHttp.HttpMethods.GET.name(), pathSegments, queryParams, null, context)
       ).flatMap(s -> {
         try {
-          return Mono.justOrEmpty(buildState(s, key, options, type));
+          return buildState(s, key, options, type);
         } catch (Exception ex) {
           return DaprException.wrapMono(ex);
         }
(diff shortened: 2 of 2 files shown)


sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -369,19 +369,22 @@
...
+  private <T> Mono<State<T>> buildStateKeyValue(
       DaprProtos.GetStateResponse response,
       String requestedKey,
       StateOptions stateOptions,
       TypeRef<T> type) throws IOException {
     ByteString payload = response.getData();
     byte[] data = payload == null ? null : payload.toByteArray();
     T value = stateSerializer.deserialize(data, type);
+    if (value == null) {
+      return Mono.empty();
```
artursouza, line 381: Well, it depends on the scenario. Having null here can mean the value exists, is null, has an etag and (why not) metadata. The value being null is not the same as "record not found".

I am not sure if we should return empty mono even for "record not found". This probably goes all the way into how each component handles key not found. Can we guarantee that no "relevant" metadata is returned when "key not found" in the component implementation? This new behavior would drop any extra information. I think it is responsibility of the app to check if "value is null" is enough to consider "record not found" for their scenario.

artursouza (changes requested):

artursouza:
@xiazuojie Thanks for your contribution. Please, see my comment regarding this change.

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
xiazuojie: "record not found" and value "null" is not the same when the state value is properly serialized as JSON.
"record not found": emtpy byte[].
"null value": "null".getBytes(). It's a 4 byte array.

Because fields in protobuf is normally not nullable, the only way to represent "record not found" in current GetStateResponse in a empty byte[].
```
message GetStateResponse {
  // The byte array data
  bytes data = 1;

  // The entity tag which represents the specific version of data.
  // ETag format is defined by the corresponding data store.
  string etag = 2;

  // The metadata which will be sent to app.
  map<string,string> metadata = 3;
}
```
I believe it's safe to assume byte[] means "record not found".

artursouza: I agree that handling "record not found" in a unique way belongs to the API design. Now, we need to design the SDK based on the actual API we have. This is where my comments are coming from. We can discuss changes to the dapr API but that should be in another issue IMO.

I am not worried about string with "null" content. I am concerned about the case where the value is null (not "null") in the database and contains an etag. With this change, the etag is dropped and user will assume it is the case for key not found.

Checking for null payload or empty payload is not enough.

This is why I think the app should understand what is the behavior for the state store he is using and do the check. If I am using state stores that have etag, for example, the right thing to do is check for null etag on the response and not the payload. Our redis state store actually implements etag.

On the other hand, not every state store supports etag, so for those the check will be different.

This is why I am saying that (with the current Dapr API), the app must know how the state store being used handles "record not found". This change would create limitations.

xiazuojie: I think it comes down to different opinions about what's more important.

- being null-safe, or
- support the corner cases of null values with etag

protobuf and golang (which means current Dapr API) do not support null values. They are specifically designed to not permit null values. Being null-safe is more important to them.

I do not think Java SDK somehow can "support" null values (when current Dapr API does not). 
In my opinion, current Java SDK implementation is not "supporting" null values. It's treating an empty byte[] as value null, which is misinterpreting values.
When null is serialized to byte[], it's not null. It's usually a byte array indicating a null value. For JSON, it's 4 bytes array - "null".

I do not think My PR break the support of null values since they are not supported in the first place. My RP is adding a convention of treating empty byte[] as "record not found" in order to be null-safe.

artursouza: > I think it comes down to different opinions about what's more important.
> 
> * being null-safe, or
> * support the corner cases of null values with etag

I agree with this statement. It all goes down to what it means to be null-safe. In this API, it means that users will get a State object that is never null. A state with null value is a legit possibility (semantically, since it is byte[] in Dapr API), returning empty Mono does not allow scenarios where app wants to save value as null because the response for "key not found" and "null value" would be the same. I've written a new test to showcase this and is in this PR (I've confirmed in the gRPC client that value is an empty byte array for "key not found" and "null value" scenarios): https://github.com/dapr/java-sdk/pull/528

> 
> protobuf and golang (which means current Dapr API) do not support null values. They are specifically designed to not permit null values. Being null-safe is more important to them.

I agree. I am not advocating for "null value" to be what is actually saved in the database. What I am saying is the value semantically mapping to null (empty byte[]) is returned in both cases: "key not found" and "value is null".


> 
> I do not think Java SDK somehow can "support" null values (when current Dapr API does not).
> In my opinion, current Java SDK implementation is not "supporting" null values. It's treating an empty byte[] as null, which is misinterpreting values.

I see this differently. The interpretation of empty byte[] to an object type is a key functionality in the SDK. How do you deserialize empty byte[] to MyClass? Our default serializer chose to convert it to null. Again, the custom serializer comes handy here: the app decides how to solve this per domain class.

> 
> When null is serialized to byte[], it's not null. It's usually a byte array indicating a null value. For JSON, it's a 4 bytes array - "null".getBytes().
> In short:
> 
> * Deserializing "null" as null: ok. Means the state exists, its value is null.
> * Deserializing empty byte[] as null: wrong. Should means "record not found" by convention.
>   Current Java SDK implementation interprets both "null".getBytes() and empty byte[] as null. It's losing the differentiation.

This is done by the serializer. We expect apps to use their own deserializer to solve the ambiguity. We cannot expect the convention proposed above to be true for every possible serializer. It can be true for JSON serializer but we are designing to support other serializers too.

> 
> I do not think My PR breaks the support of null values since they are not supported in the first place. My RP is suggesting a sensible convention of treating empty byte[] as "record not found".

We cannot adopt this convention because empty byte[] is also what is saved when user saves "null object". This behavior is beyond just the SDK like I said in the comment above.

> 
> I do not suggest that Dapr API be updated to support null values.

I agree. But I think it should handle "key not found" in a way that is not ambiguous with null value.


I really appreciate this PR and the healthy discussion here. These types of discussions is what will help shape the SDKs but (more importantly) the Dapr runtime APIs.

artursouza:
I have discussed this change with another maintainer in Dapr. Dapr cannot determine semantically if "null value" is semantically the same as "value not found". Components can implement those scenarios differently and even return metadata, so returning Mono.empty() will drop information to the app. If apps wants to have a Facade on top of the SDK to handle this, it is OK. We believe Dapr should not be opinionated about this particular scenario.


I am closing this PR for now. Thanks a lot for your contribution as this raised valuable discussions.

Issue #425, opened 2020-12-19
Fix flaky reminderRecoveryTest


mukundansundar:
## Describe the proposal
Fix flaky reminderRecoveryTest. This test has been flaky multiple times and causes build failures. 
One example is below.
https://github.com/dapr/java-sdk/runs/1578906076

RELEASE NOTE:N/A

PR #526, opened 2021-03-24
Add pause between stopping/starting runtime

halspang:
# Description
ActorReminderRecoveryIT and ActorTimerRecoveryIT both rely on reminders/timers being persisted when dapr restarts. On some restarts, it appears that the placement service shares its placement table during on ongoing update. This causes the dapr runtime to use the wrong address for the actor.

Placement service logs:
```
INFO[0322] Entering removeMember()                       instance=Hals-MacBook-Pro.local scope=dapr.placement.raft type=log ver=edge
INFO[0322] Loading: 192.168.0.219:50394, 0, 0, ActorReminderRecoveryIT  instance=Hals-MacBook-Pro.local scope=dapr.placement.raft type=log ver=edge
INFO[0322] Loading: 192.168.0.219:50489, 0, 0, ActorReminderRecoveryIT  instance=Hals-MacBook-Pro.local scope=dapr.placement.raft type=log ver=edge
INFO[0322] Start desseminating tables. memberUpdateCount: 1, streams: 2, targets: 2, table generation: 10  instance=Hals-MacBook-Pro.local scope=dapr.placement type=log ver=edge
INFO[0322] Removing 192.168.0.219:50394                  instance=Hals-MacBook-Pro.local scope=dapr.placement.raft type=log ver=edge
INFO[0322] Completed dessemination. memberUpdateCount: 1, streams: 2, targets: 2, table generation: 10  instance=Hals-MacBook-Pro.local scope=dapr.placement type=log ver=edge
```

```
java.lang.RuntimeException: io.dapr.exceptions.DaprException: INTERNAL: error invoke actor method: failed to invoke target 192.168.0.219:50394 after 3 retries
```

When this interleaving happens, the error message received references the actor with the port that was being removed. This implies that the table isn't being updated appropriately and the runtime gets the old port. When the actor is receiving continuous calls, the placement service does not update the location as it believes the actor is busy.

## Issue reference
Closes: #425

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/actors/ActorReminderRecoveryIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/actors/ActorReminderRecoveryIT.java
@@ -12,7 +12,6 @@ import io.dapr.it.AppRun;
 import io.dapr.it.BaseIT;
 import io.dapr.it.DaprRun;
 import io.dapr.it.actors.app.MyActorService;
-import io.dapr.it.state.GRPCStateClientIT;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.junit.After;
 import org.junit.Before;
@@ -20,7 +19,7 @@ import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


nobodyiam:
why is it 10 seconds? will it vary in different environments?
I would suggest we take a look at the [awaitility](https://github.com/awaitility/awaitility) project, it may help.

halspang:
> why is it 10 seconds? will it vary in different environments?
> I would suggest we take a look at the [awaitility](https://github.com/awaitility/awaitility) project, it may help.

Admittedly it's fairly arbitrary. The issue here is that what we're waiting to propagate is information that exists inside the runtime itself. This information isn't available to the Java SDK (nor should it be). Since we can't access the required state, I don't think there's anything we could really gate on. The library looks useful though. 

For context I did at first try and do something similar where I would wait and ensure the app started responding again. The issue with this was that in this state, the continued requests to the actor would cause it to be held in memory longer so there'd have to be some arbitrary wait time regardless.

artursouza:
> > why is it 10 seconds? will it vary in different environments?
> > I would suggest we take a look at the [awaitility](https://github.com/awaitility/awaitility) project, it may help.
> 
> Admittedly it's fairly arbitrary. The issue here is that what we're waiting to propagate is information that exists inside the runtime itself. This information isn't available to the Java SDK (nor should it be). Since we can't access the required state, I don't think there's anything we could really gate on. The library looks useful though.
> 
> For context I did at first try and do something similar where I would wait and ensure the app started responding again. The issue with this was that in this state, the continued requests to the actor would cause it to be held in memory longer so there'd have to be some arbitrary wait time regardless.

I like the awaitility project - we should use this in our test project here. Like Hal said, this is to avoid a race condition in the runtime where the state is not exposed. @halspang please, create an issue in dapr/dapr - we should not have to wait to restart.

artursouza (approved):

mukundansundar (approved):

PR #528, opened 2021-03-26
Return empty metadata for key not found + add IT test.

artursouza:
# Description

Return empty metadata for key not found + add IT test.

## Issue reference

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk-tests/src/test/java/io/dapr/it/state/AbstractStateClientIT.java
```diff
@@ -249,6 +250,52 @@ public void saveUpdateAndGetStateWithEtag() {
...
+    State<MyData> stateFound = daprClient.getState(STATE_STORE_NAME, stateKey, MyData.class).block();
+    assertEquals(stateKey, stateFound.getKey());
+    assertNotNull(stateFound.getValue());
+    assertEquals("data in property A", stateFound.getValue().getPropertyA());
+    assertEquals("data in property B", stateFound.getValue().getPropertyB());
+    assertNotNull(stateFound.getEtag());
+    assertFalse(stateFound.getEtag().isEmpty());
+    assertNull(stateFound.getOptions());
+    assertNull(stateFound.getError());
+    assertEquals(0, stateFound.getMetadata().size());
```
mukundansundar, line 284: in this case whenever a metadata field is returned it is always empty?

halspang (commented):
Couple things, some potentially may not be related to this PR:

- Should we change the other State constructors to return an empty map instead of null for metadata? Is there a case for null metadata?
- Based on the build state method, is there a time when we actually assign the metadata? Looking through the http client, it doesn't seem like it's assigned during any gets. Is it saved as part of the value?
- Only the GRPC client has tests for metadata, is this intended?
- Is this a breaking change? Do we think customers rely on the null metadata?

sdk-tests/src/test/java/io/dapr/it/state/AbstractStateClientIT.java
artursouza: It is possible. Today, Redis (the state store we use for ITs) does not do that.

artursouza:
> Couple things, some potentially may not be related to this PR:
> 
> * Should we change the other State constructors to return an empty map instead of null for metadata? Is there a case for null metadata?

I thought about that. We would need to see where it is being used and see if it makes sense.

> * Based on the build state method, is there a time when we actually assign the metadata? Looking through the http client, it doesn't seem like it's assigned during any gets. Is it saved as part of the value?

We don't have that in http yet, you are right. We would need to see how the API returns metadata in Dapr API.

> * Only the GRPC client has tests for metadata, is this intended?

We need to see how Dapr API handles metadata response in HTTP - it might not be supported.

> * Is this a breaking change? Do we think customers rely on the null metadata?

Well, it can be depending on the logic the customer has. On the other hand, it is odd that HTTP has a different behavior than GRPC. Since GRPC is the default client, it looks like a bug in the HTTP client.

artursouza:
HTTP API does not return metadata, it only returns ETag: https://github.com/dapr/dapr/blob/39b076c273126791a72bc952f4a3805841b987fe/pkg/http/responses.go#L45

If HTTP were to return metadata too, it would need to be handled with the "metadata." prefix in the header, similar to how it handles incoming metadata: https://github.com/dapr/dapr/blob/39b076c273126791a72bc952f4a3805841b987fe/pkg/http/api.go#L1356

I suggest we merge this PR because the behavior is aligned to how the runtime is implemented. Also, it makes the HTTP implementation behave like the gRPC implementation for this example and the test confirms both are working the same way.

halspang (approved):

mukundansundar:
I think for the key not found scenario, the empty metadata makes sense ... I think HTTP API in runtime should be enhanced to return metadata also and the implementations in components-contrib need to be enhanced to return any metadata or empty map.

mukundansundar (approved):

artursouza:
Issue to follow up in the runtime: https://github.com/dapr/dapr/issues/3172

Issue #503, opened 2021-03-03
[RECALLED] Wrong value in state store (redis) when use custom json serializer  


abogdanov37:
I use dapr v1.0.0. I wrote the actor use Java SDK. To use the right OffsetDateTime class serialization I wrote a custom Json serializer `class ISO8601DaprObjectSerializer extends ObjectSerializer implements DaprObjectSerializer`. 
## Expected Behavior
When I use this serializer I want to see JSON in state store
## Actual Behavior
Now in state store I see base64 encoded string. In this case, I should use a special deserialized object constructor with string parameter.
## Steps to Reproduce the Problem
When register actor add custom serializer
```
import com.fasterxml.jackson.annotation.JsonInclude;
import com.fasterxml.jackson.core.JsonGenerator;
import com.fasterxml.jackson.databind.*;
import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
import io.dapr.client.ObjectSerializer;
import io.dapr.client.domain.CloudEvent;
import io.dapr.serializer.DaprObjectSerializer;
import io.dapr.utils.TypeRef;

import java.io.IOException;
import java.time.OffsetDateTime;
import java.time.ZoneId;
import java.time.format.DateTimeFormatter;
import java.util.regex.Pattern;

public class ISO8601DaprObjectSerializer extends ObjectSerializer implements DaprObjectSerializer {

    private static final Pattern PATTERN_ISO8601 = Pattern.compile("^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z|[+-](?:2[0-3]|[01][0-9]):[0-5][0-9])?$");
    private static final DateTimeFormatter ISO8601 = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSX").withZone(ZoneId.of("UTC"));

    private static final JavaTimeModule TIME_MODULE = new JavaTimeModule();
    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
            .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true)
[...]

Issue #509, opened 2021-03-11
[Proposal] New project to handle JSON serialization


artursouza:
The core SDK does provide solutions for serialization of application domain objects, it only offers an interface. In order to enable applications to opt-in to specific serialization solutions, it might be better to expose as an optional project (similar to how it is done in Spring Boot).

This proposal is to create a new project for JSON serialization that apps can opt in and be extensible, so it can be configured to handle specific scenarios unique for each app. This package should be specific per vendor, which means a different JSON serializer should be done in a separate project.

We can start with `dapr-jackson` project, where it uses the Jackson serializer.

abogdanov37:
It's an interesting idea. But I think that it should be a group of projects. For example, a common project `dapr-serialization` with common modules. And then `dapr-serialization-jackson`, `dapr-serialization-jsoniter` and other.

artursouza:
@abogdanov37 What would be in `dapr-serialization`? I wonder if it is something already present in `dapr-sdk`.

stliu:
totally agree, we have run into many dependencies issues caused by the hard dependency of jackson, you can find more here https://github.com/Azure/azure-sdk-for-java/issues?q=is%3Aissue+jackson, and azure sdk is planning to remove it https://github.com/Azure/azure-sdk-for-java/issues/24480

meanwhile, you may also consider this https://github.com/Azure/azure-sdk-for-java/issues/24530

PR #539, opened 2021-05-04
Fix issue where custom JSON serializer generates Base64 encoded data in state store

abogdanov37:
# Description

Refactor ObjectSerializer class to add ability set custom object mapper. 
Refactor DaprStateAsyncprovider class to use custom serializers based on ObjectSerializer class as Json serializer. 
Add unit tests.

## Issue reference

#503 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk-actors/src/main/java/io/dapr/actors/ActorId.java
```diff
@@ -20,7 +20,7 @@
   /**
    * An error message for an invalid constructor arg.
    */
-  private final String errorMsg = "actor needs to be initialized with an id!";
+  private static final String errorMsg = "actor needs to be initialized with an id!";
```
halspang, line 23: Static final variables should be capitalized: ERROR_MSG

sdk-actors/src/test/java/io/dapr/actors/runtime/DaprStateAsyncProviderTest.java
```diff
@@ -74,6 +73,28 @@ public int hashCode() {
 
   }
 
+  class CustomJsonSerializer implements DaprObjectSerializer{
```
halspang, line 76: Missing a space before `{`

halspang (commented):
Code seems fine in general, just some formatting callouts. My concern here is how this fits into the larger picture. Especially in regards to the conversation that happened on the issue and how it relates to https://github.com/dapr/java-sdk/issues/509.

We also have an IT failure which should be looked into.

@artursouza What are your thoughts?

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
```diff
@@ -69,7 +73,7 @@
         }
 
         T response = this.stateSerializer.deserialize(s, type);
-        if (this.isStateSerializerDefault && (response instanceof byte[])) {
+        if (this.isStateSerializerJson && (response instanceof byte[])) {
```
artursouza, line 76: Changing this logic will make this not backwards compatible - it means that values saving with the current version of the code will not be able to be read from a newer version.

After thinking about this same problem in another PR, I think this should be done by a configuration. When registering the ActorType, a config can be set to determine is the data is written as Text or binary (Base64). The default behavior is Base64. This way, existing code does not break.

artursouza (changes requested):
@abogdanov37 Thanks for your contribution.

My main concern is backwards compatibility. This time, I think I found a way out, please, see my comment.

@halspang Yes, we need to be mindful about the vision to move default serializer out of this package. This improvement (IMO) does not invalidate that.

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
abogdanov37: Hi! I understood your point. But we have a unit test that checks that binary processing is OK. See https://github.com/dapr/java-sdk/blob/master/sdk-actors/src/test/java/io/dapr/actors/runtime/DaprStateAsyncProviderTest.java#L166
I add this test for a custom serializer and it works well. 
I can add a method to the serializer interface with the name `useMimeType` or `checkMimeType` and change behavior use this method

abogdanov37: @artursouza since Java 8 we have a `default` method. Because of that, it won't be breaking changes. I still can't imagine a case in which current changes will be breaking. Did you find this case with integration tests? May be beta tester give this information to you? 
I have a real system with a custom serializer and many states stored in binary format. I try changed SDK for binary format. All works as expected.

artursouza: @abogdanov37 There are 2 things here:

1. Adding to the serializer interface might be done without breaking by using default - OK, but I would still be concerned over adding more to that interface. I would like to keep it as simple as possible. So, I would be OK if this setting was outside the serializer.

2. This is a breaking change because data saved with the old code cannot be read with the new code - we don't have test for that yet, I am inferring this by looking at the code. If you can prove that data saved with the old code is read with the new, that would close this concern.


If not, I plan to address this in the next milestone. I am not 100% satisfied with our current serializer approach.

abogdanov37: Thanks for replying. I can prove it. It's not a problem. Current tests for load base64 encoded string works fine. 
I implement my idea with extend interface. If you want to look at I can provide new mr. Let me now if you want.

abogdanov37: Yep. I wrote about it. If you want I double check it.

artursouza: @abogdanov37 Just confirming. Thanks. I will check that later before merging the PR. If that confirms, then this PR can be merged without any extra config.

abogdanov37: @artursouza I double-check PR changes. It has full backward compatibility.

artursouza: I validated that this PR is a breaking change, like I thought originally. This is the error I get:

```txt
== APP == com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `int` from String "MjE=": not a valid Integer value

== APP ==  at [Source: (byte[])""MjE=""; line: 1, column: 1]

== APP ==       at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:67)

== APP ==       at com.fasterxml.jackson.databind.DeserializationContext.weirdStringException(DeserializationContext.java:1702)

== APP ==       at com.fasterxml.jackson.databind.DeserializationContext.handleWeirdStringValue(DeserializationContext.java:947)

```

How to reproduce:

1. `git checkout master`
2. `mvn clean install`
3. Apply patch attached in this comment (unzip first, of course): `git apply modify_actor_example.patch`
4. Run on a separate terminal: `cd examples; dapr run --components-path ./components/actors --app-id demoactorservice --app-port 3000  -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.actors.DemoActorService -p 3000`
5. Run the client: `cd examples; dapr run --components-path ./components/actors --app-id demoactorclient -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.actors.DemoActorClient`
6. Let it run for a few seconds (10 or so). Then, stop it with Ctrl+C or something.
7. `cd ..` to go back to project's main directoy.
8. `git stash` to stash the changes from step 3.
9. `gh pr checkout 539` to apply the changes for this PR.
10. `git stash pop` to reapply changes from step 3.
11. `mvn clean install`
12. Repeat step 4.
13. Repeat step 5.
14. Now go to terminal for step 12 and see exceptions like the one above.

[modify_actor_example.zip](https://github.com/dapr/java-sdk/files/6645320/modify_actor_example.zip)

abogdanov37: Hi! Thanks for example. Sorry for my bad testing. I check only for objects but not for primitives (((

abogdanov37: I found my place in my code where I convert Base64. I can reproduce this behavior in tests. Take a time to update code to avoid breaking changes. Thank you @artursouza!

artursouza: @abogdanov37 You might want to enable the new behavior via a System property, this way, apps can opt in when they know it will not break them (new app, for example).

abogdanov37:
```
        T response;
        try {
          response = this.stateSerializer.deserialize(s, type);
        } catch (IOException ex) {
          byte[] data = Arrays.copyOfRange(s, 1, s.length - 1);
          response = this.stateSerializer.deserialize(Base64.getDecoder().decode(data), type);
        }
```
this solution works for all types exclude String. Strings pass through as is (base64 encoded)
Feature flag is a great idea!
By System property you mean JVM property?

abogdanov37: @artursouza  I push another solution based on "fair" conversion. 
If we look there https://github.com/dapr/java-sdk/blob/0e5e3112de9604b874b1832111e5a453b797eaeb/sdk-actors/src/main/java/io/dapr/actors/runtime/DaprHttpClient.java#L98 we can see that we have a transformation to Base64
I add code to transform Base64 from store to origin byte array which can be processed.

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
```diff
@@ -68,6 +68,19 @@
           return Mono.empty();
         }
 
+        if (!this.isStateSerializerDefault) {
+          if (s.length == 0) {
```
artursouza, line 72: Why is this needed only in the deserialize path?

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
```diff
@@ -68,6 +68,19 @@
...
         }
 
+        if (!this.isStateSerializerDefault) {
+          if (s.length == 0) {
+            return Mono.empty();
+          }
+          s = OBJECT_MAPPER.readValue(s, byte[].class);
+          if (type.getType().equals(String.class)) {
+            byte[] out = new byte[s.length + 2];
+            out[0] = '"';
```
artursouza, line 78: This does not cover the scenarios where the string contains a quote already since it would not be escaped.

artursouza (changes requested):

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
abogdanov37: In another way we'll get an exception

sdk-actors/src/main/java/io/dapr/actors/runtime/DaprStateAsyncProvider.java
abogdanov37: Quotes will be added in the serialization process. Give me please an example I'll add it in tests

artursouza: `my "string" with quotes`

abogdanov37: image
It'll be quoted in the serialization process.

abogdanov37: I add two cases for quoted strings. As I saw in debug additional quotes had been escaped with serializer.

artursouza: I am still concerned about this logic. This should not be needed. I will need more time to try this PR on my desktop first.

cicoyle:
gentle ping - @abogdanov37 mind resolving the conflicts and addressing the ^ comments?

cicoyle:
gentle ping - @abogdanov37

salaboy:
closing due iniactivity @cicoyle

PR #545, opened 2021-05-14
[DRAFT] Add support for Actor Reentrancy

halspang:
# Description
This adds the new Reentrancy field to the actor runtime config
and perserves the Dapr-Reentrancy-Id header which is required for
all reentrant calls.

## Reasons for Draft
- Needs updated/new tests
- Some existing tests are failing, needs to be investigated
- Only functions in HTTP calls
- Some built in protections around actors in the SDK had to be bypassed, need to be reconsidered under reentrant case

## Issue reference

Please reference the issue this PR will close: #533

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


sdk-actors/src/main/java/io/dapr/actors/client/ActorProxyImpl.java
```diff
@@ -77,6 +81,20 @@ public String getActorType() {
    */
   @Override
   public <T> Mono<T> invokeMethod(String methodName, Object data, TypeRef<T> type) {
+    final Object reentrancyId = ActorRuntime.getInstance().getActorManager(actorType)
+        .getReentrancyId(actorId.toString(), actorType).block();
+
+    if (reentrancyId instanceof Optional) {
+      final Optional rawOptional = (Optional) reentrancyId;
+      if (rawOptional.isPresent() && rawOptional.get() instanceof String) {
+        System.out.println("Attempting to utilize ~~REENTRANCY~~");
```
artursouza, line 90: reminder: remove println

sdk-actors/src/main/java/io/dapr/actors/client/DaprClient.java
```diff
@@ -23,4 +25,9 @@
    */
   Mono<byte[]> invoke(String actorType, String actorId, String methodName, byte[] jsonPayload);
 
+  Mono<byte[]> invoke(String actorType,
+                      String actorId,
+                      String methodName,
+                      byte[] jsonPayload,
+                      Map<String, String> headers);
```
artursouza, line 32: same

sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
```diff
@@ -38,6 +42,15 @@
...
   public Mono<byte[]> invoke(String actorType, String actorId, String methodName, byte[] jsonPayload) {
+    return invoke(actorType, actorId, methodName, jsonPayload, Collections.emptyMap());
+  }
+
+  @Override
+  public Mono<byte[]> invoke(String actorType,
+                             String actorId,
+                             String methodName,
+                             byte[] jsonPayload,
+                             Map<String, String> headers) {
```
artursouza, line 53: I assume the gRPC implementation is still in progress for this.

sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java
```diff
@@ -244,9 +244,9 @@ protected ActorStateManager getActorStateManager() {
    * Resets the cached state of this Actor.
    */
   void rollback() {
-    if (!this.started) {
-      throw new IllegalStateException("Cannot reset state before starting call.");
-    }
+    //    if (!this.started) {
+    //      throw new IllegalStateException("Cannot reset state before starting call.");
```
artursouza, line 248: Why is this needed?

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
```diff
@@ -14,12 +15,13 @@
...
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 
 /**
  * Manages actors of a specific type.
  */
-class ActorManager<T extends AbstractActor> {
+public class ActorManager<T extends AbstractActor> {
```
artursouza, line 24: Why does this need to be made public? Making this public is a big deal since it increases the API surface.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
```diff
@@ -50,6 +57,7 @@
     this.runtimeContext = runtimeContext;
     this.actorMethods = new ActorMethodInfoMap(runtimeContext.getActorTypeInformation().getInterfaces());
     this.activeActors = Collections.synchronizedMap(new HashMap<>());
+    this.actorReentrancyContext = new ActorReentrancyContext();
```
artursouza, line 60: The reentrancy context is per method, not?

Still, I don't think this will work because it would need a way to garbage collect. Imagine that the call stack failed, there would need to be a way to understand that the call stack is over and release the id.

I like the idea of providing the context via an argument to the Actor's method is Reentrancy annotation is present. This way, it remains a stateless solution.

The preferred solution would still be to have this context passed through via Mono. I think we should move this to the next release since it will require more investigation.

artursouza (changes requested):

sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
```diff
@@ -65,7 +68,25 @@ private ActorClient(DaprApiProtocol apiProtocol, ManagedChannel grpcManagedChann
...
+   * @param methodName  Method name to invoke.
+   * @param jsonPayload Serialized body.
+   * @param headers     Request headers.
+   * @return Asynchronous result with the Actor's response.
+   */
+  Mono<byte[]> invoke(String actorType,
+                      String actorId,
+                      String methodName,
+                      byte[] jsonPayload,
+                      Map<String, String> headers) {
```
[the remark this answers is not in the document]
halspang: It's http specific currently, but I thought that grpc Metadata was also a Map<String, String> hence the raw map here. I can rename it/make it into a type if you'd prefer though.

sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
halspang: Yes, the problem here is that the protobuf doesn't appear to expose Metadata to us, so I think this change might require a protobuf update.

sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java
halspang: This was just a quick hack to get a PoC working so I could publish the PR and get early feedback. If left in the way it is though, it explicitly forbids reentrancy.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
halspang: It doesn't, technically. At this point I was just working towards a PoC but locally I now have the underlying ID (reason why I exposed it) provided by the already exposed ActorRuntime.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
halspang: > The reentrancy context is per method, not?

The context is per actor id/type combo.

> I like the idea of providing the context via an argument to the Actor's method is Reentrancy annotation is present. This way, it remains a stateless solution.

I avoided passing it through via a method because we originally wanted to have the passthrough happen without downstream actor's having to touch it. I don't think we can achieve that just by passing it through so we'd have to make that call about experience.

> Still, I don't think this will work because it would need a way to garbage collect. Imagine that the call stack failed, there would need to be a way to understand that the call stack is over and release the id.

This is actually one of the few places where Reactor was able to help us. On the original call into the actor we can setup a terminal event (currently we just have onSuccess) that clears the ID. This comes in both success and error flavors. If the actual Java runtime crashes, the data is cleared anyway, so I think this is fine.

> The preferred solution would still be to have this context passed through via Mono. I think we should move this to the next release since it will require more investigation.

As for passing the context through via Mono's SubscriberContext, I don't think we're actually using Reactor in a way that this is possible. The SubscriberContext as we need it is available in a single call chain. Since our chain involves a method that can call into other methods (thus creating a nested mono), we lose access to the original pipeline and context in the nested monos.

I agree this change has gotten a lot more complex than anticipated and we shouldn't try and rush it into the release on the day of code freeze.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorInvocationContext.java
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) Microsoft Corporation and Dapr Contributors.
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors.runtime;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ActorInvocationContext {
```
artursouza, line 11: This looks a lot like the Context from Mono.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
```diff
@@ -177,6 +185,20 @@
     }).thenReturn(true);
   }
 
+  Mono<Optional<String>> getReentrancyId(String actorId, String actorType) {
```
artursouza, line 188: No need to use Mono if values are "static".

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorManager.java
```diff
@@ -298,13 +320,18 @@
...
+      }
+
       return actor.onPreActorMethodInternal(context)
           .then((Mono<Object>) func.apply(actor))
           .switchIfEmpty(
               actor.onPostActorMethodInternal(context))
           .flatMap(r -> actor.onPostActorMethodInternal(context).thenReturn(r))
           .onErrorMap(throwable -> {
-            actor.rollback();
+            actor.rollback(context);
```
artursouza, line 334: Why is this needed? Does it break telemetry?

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
```diff
@@ -249,6 +250,32 @@ public ActorRuntimeConfig getConfig() {
...
+   * language invocation.
+   *
+   * @param actorTypeName   Actor type name to invoke the method for.
+   * @param actorId         Actor id for the actor for which method will be invoked.
+   * @param actorMethodName Method name on actor type which will be invoked.
+   * @param payload         RAW payload for the actor method.
+   * @param reentrancyId    Id used to match reentrant requests.
+   * @return Response for the actor method.
+   */
+  public Mono<byte[]> invoke(String actorTypeName,
```
artursouza, line 264: Again, another public method that will force us to keep supporting this, while passing context as a param instead of using Mono's context.

artursouza (changes requested):
I saw some key points that we need to address before proceeding with this change.

@skyao @xiazuojie can we get some help from Alibaba in using Mono's context here?

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorRuntime.java
danp11: Hi all, 

This PR is a beutiful one 👍 and is coming with some super nice additions to the actors.

However, it looks like it is a bit stale atm. @halspang @skyao @xiazuojie , any chance that you 
can look into Arthurs comments. 

Thank you and have a nice weekend!

From a Dapr Java-SDK user :-)

/Dan

johnou:
ping @halspang

pravinpushkar:
@halspang If you are not planning to work on this right now then feel free to assign it to me. I would like to continue on this.

halspang:
This hasn't been a priority for a while. It was discussed offline, but never here that others would take over this work. I'm not sure where that ended up or if they need this PR open anymore. Given how old this is, it's probably best to close the PR and start a new one with whoever decided to take this over.

Issue #562, opened 2021-06-08
Blocking call detected by BlockHound when invoke Actor methods, and access actorStateManager methods


xiaoweil:
Hi there,

We are experimenting with the dapr actor service, and have seen some blocking calls detected by BlockHound. Wondering if these are false positive or we did something wrong with the actor usage. Please see below for example:

Create a simple actor call Test

```
@ActorType(name = "Test")
public interface Test {
  @ActorMethod(name = "testHello")
  Mono<Void> testHello(String name);

  @ActorMethod(name = "testName")
  Mono<String> testName(String name);
}
```
```
public class TestImpl extends AbstractActor implements Test {
  private static final String NAME = "name_";

  public TestImpl(ActorRuntimeContext runtimeContext, ActorId id) {
    super(runtimeContext, id);
  }

  @Override
  public Mono<Void> testHello(String name) {
    return super.getActorStateManager().set(NAME + name, name);
  }

  @Override
  public Mono<String> testName(String name) {
    return super.getActorStateManager().get(NAME + name, TypeRef.STRING);
  }
}
```

Create testing endpoints with spring webflux:

```
  @PostMapping(path = "/testHello")
  public Mono<Void> testHello(@RequestBody(required = false) byte[] body) {
    return Mono.just(new String(body))
        .map(id -> new ActorProxyBuilder<Test>(Test.class, actorClient).build(new ActorId(id)))
        .flatMap(test -> test.testHello(new String(body)));
  }

  @PostMapping(path = "/testName")
  public Mono<String> testName(@RequestBody(required = false) byte[] body) {
    return Mono.just(new String(body))
        .map(id -> new ActorProxyBuilder<Test>(Test.class, actorClient).build(new ActorId(id)))
        .flatMap(test -> test.testName(new String(body)));
  }
```

Register the actor:
`ActorRuntime.getInstance().registerActor(TestImpl.class);`

Now make a remote call to the dapr service:
`http://localhost:3500/v1.0/invoke/testapp/method/testName`

We got the following BlockHound warning on DaprGrpcClient.invoke:
```
reactor.blockhound.BlockingOperationError: Blocking call! jdk.internal.misc.Unsafe#park
	at java.base/jdk.internal.misc.Unsafe.park(Unsafe.java) ~[na:na]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.MonoCallable] :
[...]

PR #564, opened 2021-06-11
Change Actor gRPC client to async.

artursouza:
# Description

Change Actor gRPC client to async.

## Issue reference

Please reference the issue this PR will close: #562 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
@@ -5,12 +5,26 @@
 
 package io.dapr.actors.client;
 
-import com.google.common.util.concurrent.ListenableFuture;
 import com.google.protobuf.ByteString;
+import io.dapr.config.Properties;
 import io.dapr.exceptions.DaprException;
+import io.dapr.internal.opencensus.GrpcWrapper;
 import io.dapr.v1.DaprGrpc;
 import io.dapr.v1.DaprProtos;
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.stub.StreamObserver;
 import reactor.core.publisher.Mono;
+import reactor.core.publisher.MonoSink;
+import reactor.util.context.Context;
+
+import java.util.concurrent.ExecutionException;
+import java.util.function.Consumer;
 
 /**
  * A DaprClient over GRPC for Actor.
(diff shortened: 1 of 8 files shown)


sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
```diff
@@ -5,59 +5,126 @@
...
+          MethodDescriptor<ReqT, RespT> methodDescriptor,
+          CallOptions callOptions,
+          Channel channel) {
+        ClientCall<ReqT, RespT> clientCall = channel.newCall(methodDescriptor, callOptions);
+        return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(clientCall) {
+          @Override
+          public void start(final Listener<RespT> responseListener, final Metadata metadata) {
+            String daprApiToken = Properties.API_TOKEN.get();
+            if (daprApiToken != null) {
+              metadata.put(Metadata.Key.of("dapr-api-token", Metadata.ASCII_STRING_MARSHALLER), daprApiToken);
```
halspang, line 86: Is there a negative case if the token is empty? Or is this related to the feature that has to be enabled and if it's not we hit the null (expected) case.

halspang (commented):
Overall looks good to me. Just a quick question on how the token works.

sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
artursouza: We hit null if not set. If it is set as empty, it is OK to pass it through.

sdk-actors/src/test/java/io/dapr/actors/client/DaprGrpcClientTest.java
```diff
@@ -5,114 +5,127 @@
...
-      return true;
-    }))).thenReturn(settableFuture);
-    Mono<byte[]> result = client.invoke(ACTOR_TYPE, ACTOR_ID, methodName, null);
+    Mono<byte[]> result = client.invoke(ACTOR_TYPE, ACTOR_ID_EXCEPTION, METHOD_NAME, null);
 
     assertThrowsDaprException(
         ExecutionException.class,
         "UNKNOWN",
-        "UNKNOWN: java.lang.ArithmeticException",
+        "UNKNOWN: ",
```
artursouza, line 122: This has to do with how Exceptions are handled in the grpc library, it does not show the exception as it would in the blocking path.

sdk-tests/src/test/java/io/dapr/it/actors/ActorExceptionIT.java
```diff
@@ -0,0 +1,58 @@
...
+        new ActorProxyBuilder("MyActorTest", MyActor.class, newActorClient());
+    logger.debug("Creating actorId");
+    ActorId actorId1 = new ActorId("1");
+    logger.debug("Building proxy");
+    MyActor proxy = proxyBuilder.build(actorId1);
+
+    callWithRetry(() -> {
+      assertThrowsDaprExceptionSubstring(
+          "INTERNAL",
+          "INTERNAL: error invoke actor method: error from actor service",
```
artursouza, line 51: When testing as an integration test, we got the error message.

artursouza: This behavior is observed prior to this change as well.

halspang: Seems like a fine error to me. Just curious, would the actual error be displayed anywhere? For instance, in the dapr runtime?

halspang (approved):

sdk-tests/src/test/java/io/dapr/it/actors/ActorExceptionIT.java
artursouza: It should be logged in Dapr runtime. The whole error handling for dapr is something that is not proper designed in the APIs today.

wcs1only (approved):
LGTM

PR #499, opened 2021-02-24
enable Dependabot v2

sullis:
# Description

https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

Issue #348, opened 2020-10-06
Add dependabot and fix High and Medium vulnerabilities


artursouza:
## Expected Behavior

Dependabot scan should report zero High or Medium vulnerabilities in SDK's dependencies.

## Actual Behavior

No dependabot scan.

## Steps to Reproduce the Problem

Run Dependabot scan

mukundansundar:
Potential dependabot.yml file:
```yaml
version: 2
updates:
  - package-ecosystem: "maven"
    directory: "/"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
  - package-ecosystem: "maven"
    directory: "/sdk"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
  - package-ecosystem: "maven"
    directory: "/sdk-actors"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
  - package-ecosystem: "maven"
    directory: "/sdk-autogen"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
  - package-ecosystem: "maven"
    directory: "/sdk-springboot"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
  - package-ecosystem: "maven"
    directory: "/sdk-tests"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 5
```

The dependabot yaml changes will be only applied module by module independently. See [here](https://github.com/mukundansundar/java-sdk/pulls).

Need to specify the [directory](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#directory) where pom.xml is located for each module.

This will cause build failures if the target-branch is `master` and if the updates are merged one by one to `master` branch. 


We might possibly use  a separate `develop` branch with no build actions but as a [target-branch](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#target-branch) for dependabot updates, merge all the corresponding updates to versions in all modules, then manually create a PR from develop to master branch once all the security updates for related dependencies in all modules are done. @artursouza I can probably continue to look at this if that is the way we want to move forward.

artursouza:
Let's use the branch idea that you proposed (maybe calling it `dependabot`) and have a triggered workflow on that branch where DaprBot will automatically create a PR if unit tests passes.

berndverst:
@mukundansundar can we make this P1 and add it to the Java SDK 1.2 milestone?

PR #570, opened 2021-06-29
Enable dependabot for all packages.

berndverst:
# Description

Replaces stale PR #499 to enable Dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,26 +4,32 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    target-branch: "dependabot"
   - package-ecosystem: "maven"
     directory: "/sdk"
     schedule:
       interval: "weekly"
+    target-branch: "dependabot"
   - package-ecosystem: "maven"
     directory: "/sdk-actors"
     schedule:
       interval: "weekly"
+    target-branch: "dependabot"
   - package-ecosystem: "maven"
     directory: "/sdk-autogen"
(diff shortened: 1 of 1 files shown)


berndverst:
Please take a look: @mukundansundar @halspang @artursouza

halspang (approved):
Looks good to me. I'm curious to see how often things will actually update after the first set.

berndverst:
@halspang apparently I need to pause this PR at the moment to update this for the requirements in #348:

@mukundansundar @artursouza 

It seems we want:
- Dependabot making PRs into the `dependabot` branch (to be created)
- Tests to run in the `dependabot` branch and upon success to create a PR against the master branch.

The problem I see here is... 
Doesn't this mean the `dependabot` branch needs to be synced against master regularly?

halspang:
> The problem I see here is...
> Doesn't this mean the dependabot branch needs to be synced against master regularly?

Yes, but that could also be part of a workflow. Just merge master into it at some regular cadence. Or even as part of the workflow it could create, PR, merge, and then delete its own branch so it's fresh every time. That's what I'd go for personally but I think that relies on us being able to trigger dependabot instead of it going on its own cadence.

artursouza (approved):

berndverst:
Given that this PR was merged we will need to be careful not to approve every PR dependabot creates. This PR was never updated to use the special branch concept. However, doing so would also be unnecessarily complex and also error prone.

PR #573, opened 2021-07-06
Update GRPC libraries for security updates

berndverst:
# Description

Updating GRPC libraries to the latest version for upcoming 1.3 release

* [X] Code compiles correctly

### DIFF
--- a/sdk-actors/pom.xml
+++ b/sdk-actors/pom.xml
@@ -29,6 +29,7 @@
 
   <properties>
     <maven.deploy.skip>false</maven.deploy.skip>
+    <grpc.version>1.39.0</grpc.version>
   </properties>
 
   <dependencies>
--- a/sdk-actors/pom.xml
+++ b/sdk-actors/pom.xml
@@ -74,7 +74,7 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-testing</artifactId>
-      <version>1.39.0</version>
+      <version>${grpc.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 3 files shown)


halspang (approved):

sdk-actors/pom.xml
```diff
@@ -74,6 +74,7 @@
     <dependency>
       <groupId>io.grpc</groupId>
       <artifactId>grpc-testing</artifactId>
+      <version>1.39.0</version>
```
delvh, line 77: Is there a reason why this call is not simply
```xml
<version>${grpc.version}</version>
```
?
This variable should be passed down from the parent project, which is the root pom.
Same for all the lines changed below…

artursouza: I agree. We should avoid duplicating this here and in the other lines too.

berndverst: Going back to this conversation -- this doesn't work because this particular POM file doesn't define the GRPC version parameter. That makes sense because this POM file doesn't use any other GRPC libraries, whereas the other POM files have multiple GRPC libraries.

I can extract it as a variable but that kind of doesn't make sense...

delvh: According to my understanding of Maven, the GRPC version still does not need to be defined.
The reason behind that are the following lines:
```xml
<parent>
<groupId>io.dapr</groupId>
<artifactId>dapr-sdk-parent</artifactId>
<version>1.2.0-SNAPSHOT</version>
</parent>
```
Because Maven can then internally pass the variables declared inside the parent pom down to the children. And the parent pom already defines
```xml
 <properties>
    …
    <grpc.version>1.39.0</grpc.version>
    …
</properties>
```
Hence, that variable can still be used here.

artursouza (approved):

Issue #575, opened 2021-07-07
Pub/sub binding for Spring Boot ignores @PostMapping value


mthmulders:
## Expected Behavior

In my Spring Boot application, I have a `@RestController` annotated class with one method:

```java
    @Topic(name = "speedingviolations", pubsubName = "pubsub")
    @PostMapping("/collectfine")
    public ResponseEntity<Void> registerViolation(@RequestBody final CloudEvent event) {
        // omitted for brevity
    }
```

I have added the **io.dapr:dapr-sdk-springboot** dependency, so I expect that this method will be invoked by the sidecar when a message is published on the "speedingviolations" topic.

## Actual Behavior

The Dapr sidecar logs the following error:

```
ERRO[0031] non-retriable error returned from app while processing pub/sub event b411c8c6-e191-4853-b678-930c1183e4ab: {"timestamp":"2021-07-07T18:10:45.006+00:00","status":404,"error":"Not Found","path":"/speedingviolations"}. status code returned: 404  app_id=finecollectionservice instance=edsger scope=dapr.runtime type=log ver=1.2.2
```

The path "/speedingviolations" is not where the method is mapped to; the pub/sub message should be sent to "/collectfine".

## Steps to Reproduce the Problem

See above. Crucial thing is that the `@PostMapping` annotation should use the `value` element rather than the `path` element.

## Release Note

RELEASE NOTE: **FIX** Pub/sub binding for Spring Boot properly reads @PostMapping value

mthmulders:
PS. If you feel this is indeed worth fixing, I am willing to give it a shot.

danp11:
Hi @artursouza 

I/we know you are super busy with the Dapr core development first and foremost. 
And maybe it is vacation times in the US aswell, so please don't get me wrong.

But I do believe that for the Java SDK to be successful, there should me more maintainers 
of this SDK so we can have a quite active discussion on the issues. Just my two cents and I hope
you can forward this to "the ones that decides about resources in the Dapr team at Microsoft" :-) 

Thank you for your contribution and support and the other Java SDK maintainers. It is/has been great!

With hopes that you can get more help to maintain the Java SDK,

/Dan

PR #583, opened 2021-07-25
Add support for @PostMapping("/path") to pub/sub topic subscriptions

mthmulders:
# Description
I've updated the `DaprBeanPostProcessor` class so it supports `@PostMapping("/path")` like annotations. Also, I've added tests for the `DaprBeanPostProcessor` class.

## Issue reference
This PR will close issue #575.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk-springboot/src/test/java/io/dapr/springboot/DaprBeanPostProcessorIT.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.test.context.web.WebAppConfiguration;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@ContextConfiguration(
+        classes = { DaprAutoConfiguration.class }
+)
+@ExtendWith(SpringExtension.class)
+@WebAppConfiguration
+public class DaprBeanPostProcessorIT implements WithAssertions {
```
artursouza, line 19: Please, write integration tests in the sdk-tests project.

artursouza (changes requested):
Thanks a lot for this pull request. Please, move the integration test to the sdk-tests package - see how there is a way we designed to write those already. So, org.assertj dependency should be removed too.

sdk-springboot/src/test/java/io/dapr/springboot/DaprBeanPostProcessorIT.java
mthmulders: Done! In fact, I changed an existing test; all tests used `path = "..."`, and now one test does not.

Out of curiosity, why is that module not included in the parent POM?

mthmulders: Also, if there are any instructions on how to run the integration tests locally, I'd appreciate it. Now I'm hoping for the GH Actions to do their job, I couldn't get them to run locally...

artursouza: sdk-tests is not included in the main pom because it would add a circular dependency between the integration tests and the main SDK code. The integration tests require the SDK jar to be built.

To run the integration tests locally, you basically need to have dapr installed and the sdk installed via mvn install. Then you can do mvn test in the sdk-tests folder.

artursouza:
The integration test code is not in the PR yet.

mthmulders:
> The integration test code is not in the PR yet. I will give it another look afterwards. Thanks.

I've already updated **sdk-tests/src/test/java/io/dapr/it/pubsub/http/SubscriberController.java**: one out of the many `@PostMapping` annotated methods there no longer specifies the `path` argument. All tests that depend on that controller method are still green, which I would consider a proof that the changes have the intended effect.

If you feel that the changes need more tests, could you please elaborate on what kind of test you'd like to see?

artursouza (dismissed):

sdk-springboot/src/test/java/io/dapr/springboot/DaprBeanPostProcessorIT.java
mthmulders: Thanks, I was able to run them. I don't fully understand the circular dependency remark yet, will investigate further.

artursouza (approved):

PR #591, opened 2021-08-01
Declare CloudEvent generic to allow automatic payload deserialization using Jackson

### LINKED ISSUE #576: Automatically unwrap pub/sub CloudEvent
### Background
In the Dapr SDK for .NET, one can configure the SDK to automatically unwrap _CloudEvents_ that an application receives over pub/sub using `app.UseCloudEvents()`.

The Dapr SDK for Java, however, only lets you receive a `CloudEvent`. It provides access to the underlying message using `getData`, but that method returns an `Object`. According to the specs, this can be a JSON object or string. This means in practice it can be anything - for complex types, it will most likely be a `java.util.Map`.

### Proposal
It would be cool if the SDK for Java would allow to unwrap a `CloudEvent` in the same way as the SDK for .NET does.

### Alternatives
As an alternative, if the `CloudEvent` class would be a generic class with one type parameter that describes the type of the `data`, field, it would at least make it a lot easier to unwrap the `CloudEvent` by declaring it of the right type and then calling `getData()` on it. I did a quick & dirty PoC by copying the `CloudEvent` class into my application and making it generic, and Jackson (the default serializer) seems to handle perfectly well.

mthmulders:
# Description
I've made the `CloudEvent` class a generic class, meaning its instance variable `data` is now of a generic type `<T>`. This allows Spring Boot controller methods to accept an argument `CloudEvent<MyPayload>`, which will make the code inside that controller method a lot easier as the `MyPayload` object will automatically be populated from a JSON payload.

At runtime, this generic type `<T>` resorts to `Object`, which I believe makes this backward compatible. The existing tests are all green which supports this claim.

Also, I did a small refactoring in the Spring Boot _Rest Controller_ that the `PubSubIT` uses to make it easier to maintain. 

## Issue reference

This closes #576.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/examples/src/main/java/io/dapr/examples/pubsub/http/CloudEventPublisher.java
+++ b/examples/src/main/java/io/dapr/examples/pubsub/http/CloudEventPublisher.java
@@ -47,7 +47,7 @@ public class CloudEventPublisher {
     //Creating the DaprClient: Using the default builder client produces an HTTP Dapr Client
     try (DaprClient client = new DaprClientBuilder().build()) {
       for (int i = 0; i < NUM_MESSAGES; i++) {
-        CloudEvent cloudEvent = new CloudEvent();
+        CloudEvent<String> cloudEvent = new CloudEvent<>();
         cloudEvent.setId(UUID.randomUUID().toString());
         cloudEvent.setType("example");
         cloudEvent.setSpecversion("1");
--- a/daprdocs/content/en/java-sdk-docs/_index.md
+++ b/daprdocs/content/en/java-sdk-docs/_index.md
@@ -150,7 +150,7 @@ public class SubscriberController {
 
   @Topic(name = "testingtopic", pubsubName = "${myAppProperty:messagebus}")
   @PostMapping(path = "/testingtopic")
-  public Mono<Void> handleMessage(@RequestBody(required = false) CloudEvent cloudEvent) {
+  public Mono<Void> handleMessage(@RequestBody(required = false) CloudEvent<?> cloudEvent) {
     return Mono.fromRunnable(() -> {
       try {
         System.out.println("Subscriber got: " + cloudEvent.getData());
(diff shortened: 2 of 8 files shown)


examples/src/main/java/io/dapr/examples/pubsub/http/CloudEventPublisher.java
```diff
@@ -47,7 +47,7 @@ public static void main(String[] args) throws Exception {
     //Creating the DaprClient: Using the default builder client produces an HTTP Dapr Client
     try (DaprClient client = new DaprClientBuilder().build()) {
       for (int i = 0; i < NUM_MESSAGES; i++) {
-        CloudEvent cloudEvent = new CloudEvent();
```
artursouza, line 50: Is this required? Changes must be backwards compatible, which means that code without generics should still work as-is.

mthmulders: It is not required, in the sense that it will also work without specifying the generic type. You can see that in the `PubSubIT`, where even the published message is declared without a generic type. On the consuming side, it can be consumed as a `CloudEvent` or as a `CloudEvent<PubSubIT.MyObject>`. See `SubscriberController` for an example of that.

I thought it would make sense to update the documentation. Since this file is in the **examples/** folder, I consider it part of the documentation, and hence decided to update it.

artursouza (changes requested):
LGTM. I can merge this after ITs are passing. Please, take a look.

mthmulders:
> LGTM. I can merge this after ITs are passing. Please, take a look.

Welp, must've missed that. Apologies. I've updated the helper REST controller to always return a list, even if it's empty, rather than returning `null` when no messages are present for a topic.

artursouza (approved):

PR #596, opened 2021-08-16
deprecate builders and add setters for domain classes

### LINKED ISSUE #587: Package setters on SDK Request classes
It should be possible to create an instance of a request class, e.g. the `InvokeMethodRequest` class, set the parameters on it, and then call ```<T> Mono<T> invokeMethod(InvokeMethodRequest invokeMethodRequest, TypeRef<T> type);``` on `DaprClient`. However, the setters defined on `InvokeMethodRequest` are package scope, so they cannot be called. The setters should have public scope, or there should be a constructor that allows them to be set.

Same comment applies to `PublishEventRequest`, `InvokeBindingRequest`, `GetStateRequest`, `BulkStateRequest`, `DeleteStateRequest`, `GetSecretRequest`, `GetBulkSecretRequest`. `ExecuteStateTransactionRequest` has a constructor that sets the parameters, but could perhaps also benefit from setters.

I do see the Builder pattern is available (e.g. InvokeMethodRequestBuilder), so that can always be used instead. However, this is overkill for just setting parameters for the request. Also, the builder pattern is not quite right- there should be a static method on a builder class to instantiate a new builder, rather than having to instantiate a builder via new- typically, constructors on a builder have package / private access. And rather than setting the name of the service and method in a constructor on the builder, these should be additonal methods on the builder class to make it more explicit what is being set.

mukundansundar:
# Description

closes #587 
## Issue reference

Please reference the issue this PR will close: #587 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/domain/DeleteStateRequestBuilder.java
+++ b/sdk/src/main/java/io/dapr/client/domain/DeleteStateRequestBuilder.java
@@ -10,7 +10,10 @@ import java.util.Map;
 
 /**
  * Builds a request to delete a state by key.
+ * Deprecated in favor of @see{@link DeleteStateRequest}.
+ * Deprecated since SDK version 1.3.0, slated for Removal in SDK version 1.5.0.
  */
+@Deprecated
 public class DeleteStateRequestBuilder {
 
   private final String stateStoreName;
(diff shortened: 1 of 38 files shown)


sdk/src/main/java/io/dapr/client/domain/DeleteStateRequestBuilder.java
```diff
@@ -10,7 +10,10 @@
 
 /**
  * Builds a request to delete a state by key.
+ * Deprecated in favor of @see{@link DeleteStateRequest}.
+ * Deprecated since SDK version 1.3.0, slated for Removal in SDK version 1.5.0.
```
artursouza, line 14: nit: lowercase r for removal.

artursouza (approved):

PR #598, opened 2021-08-18
Prepare build for Java16

### LINKED ISSUE #594: Allow development with JDK 16 + add build with multiple JDK versions
## Describe the proposal

Running unit tests with JDK 16 fails. We should make it pass with JDK 16.

Action items:
+ Fix code to allow unit tests to run with JDK 16
+ Change Dapr Build to run against multiple JDKs: 11, 13, 15, 16

```text
Tests run: 362, Failures: 0, Errors: 228, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for dapr-sdk-parent 1.2.0-SNAPSHOT:
[INFO] 
[INFO] dapr-sdk-parent .................................... SUCCESS [  9.167 s]
[INFO] dapr-sdk-autogen ................................... SUCCESS [ 15.819 s]
[INFO] dapr-sdk ........................................... FAILURE [ 14.405 s]
[INFO] dapr-sdk-actors .................................... SKIPPED
[INFO] dapr-sdk-springboot ................................ SKIPPED
[INFO] dapr-sdk-examples .................................. SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  41.957 s
[INFO] Finished at: 2021-08-13T11:53:23-07:00
[INFO] ------------------------------------------------------------------------
```

mthmulders:
# Description
This pull request allows building the Dapr SDK for Java on Java 16. Things that have changed:

* Upgrade to latest version of Mockito. It has a more recent version of ByteBuddy, which in turn adds support for Java 16. This reduces the number of failed test cases in *dapr-sdk* from 228 to 11.
* Modify the GitHub Actions to run on Java 11 and Java 16. This is a change from the current jobs, which only run on Java 13. It uses a "build matrix" so we can run multiple versions in parallel.
* The remaining 11 failing test cases in *dapr-sdk* seem to be caused by the fact that the [System Rules JUnit rules](https://github.com/stefanbirkner/system-rules). These rules use some reflective code that no longer works on Java 16. To fix this, I have added an `--add-opens` statement to the Surefire configuration.

## Issue reference

This PR will close #594

## Checklist

* [X] Code compiles correctly
* [ ] ~~Created/updated tests~~
* [ ] ~~Extended the documentation~~

### DIFF
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -130,6 +130,14 @@
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.22.2</version>
+        <configuration>
+          <argLine>--add-opens java.base/java.util=ALL-UNNAMED</argLine>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>3.2.0</version>
         <executions>
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -30,6 +30,9 @@
   <properties>
     <maven.deploy.skip>false</maven.deploy.skip>
     <grpc.version>1.39.0</grpc.version>
+    <argLine>
+      --add-opens java.base/java.util=ALL-UNNAMED
+    </argLine>
   </properties>
 
   <dependencies>
@@ -127,14 +130,6 @@
             </goals>
           </execution>
         </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <version>2.22.2</version>
-        <configuration>
-          <argLine>--add-opens java.base/java.util=ALL-UNNAMED</argLine>
-        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -178,11 +178,12 @@
                    <limit>
                       <counter>LINE</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>80%</minimum>
+                      <!--TODO Switch back coverage on removing the deprecated builder code-->
+                      <minimum>79%</minimum>
                     </limit>
                   </limits>
                   <excludes>
-                    <exclude>io.dapr.utils.Constants</exclude>
+                    <exclude>**/io/dapr/client/domain/**/*Builder.java</exclude>
                   </excludes>
                 </rule>
               </rules>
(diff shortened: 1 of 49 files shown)


mthmulders:
I see that this change might also require a change in the GitHub project setup: it waits for the 'build' job to be completed, but there is no longer a job by that name.

artursouza:
> I see that this change might also require a change in the GitHub project setup: it waits for the 'build' job to be completed, but there is no longer a job by that name.


I can force merge and then the build step will not be required going forward. It is just that the 'build' job is present in master branch.

sdk/pom.xml
```diff
@@ -30,6 +30,9 @@
   <properties>
     <maven.deploy.skip>false</maven.deploy.skip>
     <grpc.version>1.39.0</grpc.version>
+    <argLine>
+      --add-opens java.base/java.util=ALL-UNNAMED
```
artursouza, line 34: Is this needed for tests only? If so, is it possible to only enable this for tests?

mthmulders: It is documented in the [surefire:test](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#argLine) documentation. The property belongs to Surefire and I would argue it is - as a consequence - restricted to test phase. I am not aware of any other plugin using it.

TomCools: This property should be set in surefire itself, not in the general <argLine>. Also see this migration guide for Java 17 (longer term, but might as wel fix it now ) regarding JEP 403 - Strongly Encapsulate JDK internals.

https://github.com/johanjanssen/JavaUpgrades#jep-403-strongly-encapsulate-jdk-internals

From that page:
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
        <argLine>
            --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED
        </argLine>
    </configuration>
</plugin>
```

mthmulders: > [...] who knows what other plugins might pick it up otherwise.

The JaCoCo plugin, for once, works with that property. The _shortest_ way to make Surefire and JaCoCo play nicely together is to define a property: `<argLine>-your -extra -arguments</argLine>`.

The JaCoCo plugin also shows a more verbose example, which is to add this to the Surefire plugin (as @TomCools and Johan Janssen suggest):

```xml
<argLine>
    ${argLine}
    --add-opens java.base/java.lang=ALL-UNNAMED
    --add-opens java.base/java.util=ALL-UNNAMED
</argLine>
```

Since we have no evidence that it _unintentionally_ influences other plugins, I would argue to stick with the `<argLine>--add-opens...</argLine>` approach.

TomCools: Fair enough, let's be pragmatic about this. Consider my remark closed. 😉

Issue #543, opened 2021-05-10
Provide MockDaprClient or MockActorClient for easy unit/integration testing


javageek79:
## Describe the proposal
For making unit testing and integration testing easier, it would be great, if dapr java sdk could ship with ready configured or easy configurable mock implementations for generally used items like DaprClient or ActorClient.

For ActorClient in particular, it would also be great, if the protocol http/gRpc could be configured during tests so it would be easier to use an in memory mock http webserver instead of gRPC.

leonids2005:
Will other SDK be supported? .NET in particular?

CodeMonkeyLeet:
@leonids2005 Consider creating an issue in dapr/dotnet-sdk for tracking a similar feature if you're interested, I believe these requests are being tracked on a per SDK basis as they have different maintainers.

artursouza:
These issues are per SDK based on what is idiomatic per language. @leonids2005 Please, create an issue in the .Net SDK is you think there is a need for the same there.

artursouza:
Hi @javageek79 ,

I have created a pull request to address your ask to make mocking easier for ActorClient. See PR how it can be used: https://github.com/dapr/java-sdk/pull/606/commits/4a47177005f7e93758f2f7f7d95c0db0f50e9876#diff-e59f8f865998bb2c375fe9e23fd64df4fb3b0c9b2b89e0e5926cbcb423e19b2a

The actor’s DaprClient interface does not need to be mocked since ActorClient should be the only class offering connectivity to Actors for the app and DaprClient is hidden from the app.

I also did not find a need to create a mocking class for the non-actor DaprClient interface since it is already an interface – which should be easy to mock already.

Please, let me know what you think.

Thanks,
Artur Souza

Issue #541, opened 2021-05-07
New project for app's integration testing with Dapr


artursouza:
## Describe the proposal

Publish a new Maven artifact that allows application developers to easily write integration and end-to-end testing with Dapr. Ideally, supporting both standalone and K8s environments.

POC for integration tests on standalone mode: https://github.com/artursouza/java-sdk/commit/a284a0ecbb93c09288b598b0b7a6b0e26d943523

PR #606, opened 2021-08-21
Creates sdk-testing and adds ActorClientStub class.

artursouza:
# Description

Creates a new project: sdk-testing.
Adds ActorClientStub class for easy mocking of ActorClient class.

## Issue reference

Please reference the issue this PR will close: #543 

Created new project (sdk-testing) to be used in #541 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


artursouza:
@mthmulders  @TomCools It would be great to have your feedback on this PR.

mthmulders:
> @mthmulders @TomCools It would be great to have your feedback on this PR.

I will go through it next week.

TomCools:
> > @mthmulders @TomCools It would be great to have your feedback on this PR.
> 
> I will go through it next week.

Idem Ditto 😄... weekend first!

sdk-testing/src/main/java/io/dapr/actors/client/ActorClientStub.java
```diff
@@ -0,0 +1,38 @@
...
+ * Licensed under the MIT License.
+ */
+
+package io.dapr.actors.client;
+
+import reactor.core.publisher.Mono;
+
+/**
+ * Stub implementation of ActorClient to enable unit testing in apps.
+ * <p>This class is needed because ActorClient is a class and not an interface.</p>
```
TomCools, line 12: This is an incorrect assumption. Mockito allows you to mock Classes as well... as long they aren't _final_.
The reason for that behavior is because **Mockito will create a subclass at runtime** (see screenshot, you can see in the debugger that the instance "mock" is actually an automatically generated subclass of what you are mocking (ActorClient in this case).. So the users of the SDK don't actually need this stub implementation to help them test.

image

TomCools (commented):
I think we more clearly need to define what kinds of features we would want in an "sdk-testing" module. Adding stubs like the one that was added now doesn't add much on top of Mockito. If we want to support other testing libraries... that's something else, but i'd argue that Mockito is pretty standard.

I do however see a lot of value in the following kind of test-utilities that might be placed in this module.

-  A Testing sidecar, and a way to spin up other "components + sidecar".
- These other components could be custom wrapper this module provides in the form of
   - For HTTP: "Sidecar + WireMock"
   - For Pubsub: "Sidecar + ?"

@mthmulders you have any more ideas on this?

sdk-testing/src/test/java/io/dapr/actors/client/ActorClientStubTest.java
```diff
@@ -0,0 +1,52 @@
...
+    // Changing this test can mean a breaking change since apps might mock based on some internal behavior of the SDK.
+    ActorClientStub mock = Mockito.mock(ActorClientStub.class);
+    Mockito.when(
+        mock.invoke("MyActor", "100", "getName", null))
+        .thenReturn(Mono.just("\"mock result 100\"".getBytes(StandardCharsets.UTF_8)));
+    Mockito.when(
+            mock.invoke("MyActor", "100", "setName", "\"My Name\"".getBytes(StandardCharsets.UTF_8)))
+        .thenReturn(Mono.just("".getBytes(StandardCharsets.UTF_8)));
+
+    ActorProxyBuilder<MyActor> builder = new ActorProxyBuilder(MyActor.class, mock);
```
mthmulders, line 42: Passing a diamond operator makes it clear for the compiler that you're using the generic type. You then leave it to the compiler to infer that generic type.

```suggestion
    ActorProxyBuilder<MyActor> builder = new ActorProxyBuilder<>(MyActor.class, mock);
```

sdk-testing/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+  <description>Testing utilities for applications using Dapr SDK.</description>
+
+  <repositories>
+    <repository>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>jcenter</id>
+      <name>jcenter</name>
+      <url>https://jcenter.bintray.com/</url>
```
mthmulders, line 26: JCenter has been [closed down](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/) I believe. I would not expect new references to be introduced (and I would even expect old references to disappear someday - but that should be a separate ticket, probably?)

mthmulders (commented):
I am not yet familiar with Actors in Dapr, so I'm having a hard time understanding what the added value of these changes is.

In general, I agree with @TomCools, it would probably be a lot more useful if we could offer "ready to roll" sidecars for testing. I'm thinking something like [Test Containers](https://www.testcontainers.org/). We might even use them to test the Spring Boot SDK.

sdk-testing/src/main/java/io/dapr/actors/client/ActorClientStub.java
artursouza: Awesome! I will update the comment. On the other hand, we still need to make the method public here. I was keeping it package protected to make sure the App does not invoke any method in ActorClient and only used internally.

sdk-testing/pom.xml
artursouza: Yes, that needs to be addressed. Too bad. We had to migrate to jcenter due to another mirror being closed down before. It seems to be a thing now for public Maven repositories.

artursouza: Tracking here: https://github.com/dapr/java-sdk/issues/608

artursouza:
> I think we more clearly need to define what kinds of features we would want in an "sdk-testing" module. Adding stubs like the one that was added now doesn't add much on top of Mockito. If we want to support other testing libraries... that's something else, but i'd argue that Mockito is pretty standard.
> 
> I do however see a lot of value in the following kind of test-utilities that might be placed in this module.
> 
> * A Testing sidecar, and a way to spin up other "components + sidecar".
> * These other components could be custom wrapper this module provides in the form of
>   
>   * For HTTP: "Sidecar + WireMock"
>   * For Pubsub: "Sidecar + ?"
> 
> @mthmulders you have any more ideas on this?



We have an issue to create this "testing SDK" project: https://github.com/dapr/java-sdk/issues/541

yes, the idea is that it can help apps to write integration tests with actual Dapr sidecar running. We have that code already in our internal testing package (sdk-tests), so the idea is to create an user friendly API.

mthmulders:
A second line of thought on this: I typically try to avoid mocking "things I don't own". If I were building applications with Dapr, that would mean I would avoid mocking Dapr internals. I'd rather not have my test depend on the internals of an SDK I use.

I'd prefer a test that proofs the behaviour of my production code over a test that proofs I invoke the right methods of a third-party class. The latter may change as that SDK evolves. The former should hardly ever change.

By providing convenient mocks of our own internals, I feel we're encouraging consumers to do the former.

artursouza:
> I like the idea of this as well, but it'd be good to see it in action. We have the actor example in the Java SDK already. Do you think we could see how this would add a unit test/integration test to that as an example? I think that would help show what value this adds, or, show us that we're OK with just Mockito for now.


The unit test in this code shows that.

artursouza:
> A second line of thought on this: I typically try to avoid mocking "things I don't own". If I were building applications with Dapr, that would mean I would avoid mocking Dapr internals. I'd rather not have my test depend on the internals of an SDK I use.
> 
> I'd prefer a test that proofs the behaviour of my production code over a test that proofs I invoke the right methods of a third-party class. The latter may change as that SDK evolves. The former should hardly ever change.
> 
> By providing convenient mocks of our own internals, I feel we're encouraging consumers to do the former.


I've seen this approach before and it probably means creating Facade interfaces to abstract 3rd party interfaces. This is not  necessarily a universal practice. We already encourage mocking of Dapr client by providing it as an interface. If a developer wants to avoid that, he then should come up with his/her own interfaces to avoid exposing any direct interaction with Dapr.

To a greater extend, Dapr is that abstraction already - avoids applications to implement against specific interfaces of their cloud components by providing a common interface.

artursouza:
/ping @halspang @TomCools @mthmulders 

Code freeze is today, so I assume we will not include this in this release unless we get some consensus on this. As a maintainer, I seek consensus as much as possible - so will give everyone some time to agree on a path forward here.

halspang (approved):
Looks good to me!

artursouza:
I talked to @rynowak about this. I will try a solution where the user provides a mock implementation of their actor interface and we offer an ActorClient implementation that calls into the in-memory implementation. So users just mock their actor interface and not the method from ActorClient. I will try that in the next release, since code freeze is today. Thanks for playing.

mukundansundar (approved):
lgtm

Issue #638, opened 2021-10-04
Spring security context added as additional http header to http request via service invocation 


javageek79:
## Expected Behavior

During service invocation, we are affected by [#3408](https://github.com/dapr/dapr/issues/3408#issuecomment-930357810). This is mainly because the entire spring security context is added as a header to the request. In our case, this exceeds the default allowed header size for fasthttp requests.
As the service beeing called has no secured endpoint in this case, the  security context is of no use during this workflow.
I would expect, that per default a tranfer of security context information via service invocation is disabled, an that it can be added, if needed.

## Actual Behavior

service invocation in context of a spring security application breaks during runtime with a too large http header. see [#3408](https://github.com/dapr/dapr/issues/3408#issuecomment-930357810)

## Steps to Reproduce the Problem

Have a spring security context containing many roles and scopes received e.g. by an oauth2 jwt token.

## Release Note

RELEASE NOTE:

stefanJ-hub:
is there actually a reason why the spring security context is added as header to the service invocation? 
(i do not see a reason here)

as @javageek79  mentioned - if this can not be "simply" dropped; would be good if we can at least toggle it off.

if you know some mediation approaches (that can be applied just now) - to have the security context NOT passed into the service invocation - we would be happy to hear ;-)

artursouza:
Is the Spring security context coming from the frontend calling into the backend and then Dapr sidecar? I would like to understand the flow.

javageek79:
Hi @artursouza, yes, the flow is as follows
"public" REST API -- spring security -> Service A -- dapr service invocation -> Service B

artursouza:
I am working on an integration test to repro this. Thanks.

artursouza:
@javageek79 PR is merged, please try using the SDK version 1.4.0-SNAPSHOT

PR #642, opened 2021-10-11
Filter out headers not required for W3C tracing in service invocation.

artursouza:
# Description

Filter out headers not required for W3C tracing in service invocation.

## Issue reference

Please reference the issue this PR will close: #638 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
```diff
@@ -388,6 +390,28 @@ public void invokeServiceNoHotMono() {
     // No exception should be thrown because did not call block() on mono above.
   }
 
+  @Test
+  public void invokeServiceWithContext() {
```
halspang, line 394: Does this test actually check that the header is filtered out? I'm not sure if the `mockInterceptor` rule will only be matched if it has just those headers.

artursouza:
Great point. The intent for this is to handle W3C headers and nothing else. If we want to extend the use, we can by expanding the list.

artursouza:
So, adding everything to the header is a bug, causing the large header issue too.

Issue #218, opened 2020-02-06
The sdk should let the user choose between ipv4 loopback and ipv6 loopback for daprd communication.


LMWF:
## Expected Behavior

Today the sdk will choose the ipv4 loopback address to contact darpd: 127.0.0.1 (this was changed from "localhost" today).  Some users may prefer ipv6 or may operate in an ipv6 only environment.  The sdk should let the user choose between ipv4 loopback and ipv6 loopback.

## Actual Behavior

Today the sdk will always use the ipv4 loopback address: 127.0.0.1 to connect to daprd.

yaron2:
Might be better to detect the environment automatically and set the address accordingly. This is a detail the user shouldn't be aware of.

amanbha:
While reducing the friction because of issues faced by localhost is good. Asking user to provide this ipv4(or ipv6)address is not good solution, 
I agree it should be auto discovered for user.

artursouza:
@amanbha @yaron2 I agree for auto-detection. It would be interesting to see how the auto-detection can be done without causing delay (waiting for IPV6 connection to fail, for example).

yaron2:
A delay on startup is a non-issue.

artursouza:
@yaron2 If it is "small", I agree.

yaron2:
Looks like Java has this built in: https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getLoopbackAddress%28%29

LMWF:
Doesn't it have to use the one daprd is using?  Can't daprd be using the ipv6 loopback, and that api returns the ipv4 loopback?

artursouza:
I thought about it too. If both are present, client must use what dapr is using - which can only be known by trying to connect.

willtsai:
I have made an investigation pass into this issue and have the following findings:
* A constant called SIDECAR_IP has since been added as a part of https://github.com/dapr/java-sdk/pull/333/, which is implemented to detect the sidecar's IP address by inspecting the system properties via a call to Properties.SIDECAR_IP.get() that looks for the sidecar's IP using this logic: first try to get system property named "dapr.sidecar.ip", if not available try get environment variable named "DAPR_SIDECAR_IP", and if neither is available fallback to "DEFAULT_SIDECAR_IP" which is hardcoded to "127.0.0.1"
* Given the above, I think that these points that were raised are now already addressed: 
[...]

PR #649, opened 2021-11-04
Change "127.0.0.1" to SIDECAR_IP to dynamically account for IPv4 vs. IPv6 environments

willtsai:
# Description

I've made some changes to move away from hardcoded "127.0.0.1" IP addresses (representing `localhost`) to make the SDK and Tests compatible with exclusively IPv6 environments:
- Replaced hardcoded "127.0.0.1" in various URL strings with the SIDECAR_IP property constant
- Changed the DEFAULT_SIDECAR_IP property constant's value from the hardcoded "127.0.0.1" to instead use the Java API getLoopbackAddress(). This would make the IP default value more dynamic and account for situations where users operate in an exclusively ipv6 environment and the logic ends up falling back to the default value. 

## Issue reference

Please reference the issue this PR will close: #218 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation (N/A)


mukundansundar:
@artursouza Wasn't  there was a particular reason why 127.0.0.1 was used instead of localhost?

willtsai:
> @artursouza Wasn't there was a particular reason why 127.0.0.1 was used instead of localhost?

@mukundansundar according to https://github.com/dapr/java-sdk/pull/213, it was changed from localhost to 127.0.0.1 "to avoid local OS specific DNS lookups that might hinder performance". My changes here don't reintroduce the use of `localhost`, but rather uses the dynamically sourced `SIDECAR_IP` constant from the system.

sdk/src/main/java/io/dapr/config/Properties.java
```diff
@@ -18,7 +19,7 @@
   /**
    * Dapr's default IP for HTTP and gRPC communication.
    */
-  private static final String DEFAULT_SIDECAR_IP = "127.0.0.1";
+  private static final String DEFAULT_SIDECAR_IP = InetAddress.getLoopbackAddress().getHostAddress();
```
artursouza, line 22: This logic might not work. For example, the loopback address might be ::1 but Dapr is listening on 127.0.0.1 only. I think this needs to be a logic to auto-detect where Dapr sidecar is running instead. So, this property would change to something like "" or "auto" and add a logic to auto detect where the sidecar is running.

artursouza (changes requested):

sdk/src/main/java/io/dapr/config/Properties.java
willtsai: Good point on why loopback address might not work. However, I'm not entirely sure that we need to auto detection logic to detect where the sidecar is running for this, as `DEFAULT_SIDECAR_IP` is only used as a fallback if the `dapr.sidecar.ip` or `DAPR_SIDECAR_IP` System properties are not available which as I understand by definition tells us where the sidecar is running. Thus, I think we just need something reasonable to use for the fallback `DEFAULT_SIDECAR_IP`, so what if we use `InetAddress.getLocalHost()` as the fallback default instead of the hardcoded "127.0.0.1"? Is it reasonable to assume that the Dapr sidecar is running on the same localhost as the application if the sidecar ip properties are not defined?

willtsai:
Work in progress, please do not merge

sdk/src/main/java/io/dapr/config/Properties.java
artursouza: DAPR_SIDECAR_IP env is only used as a workaround for when the default behavior does not work. We want to have the default to be a very good guess to avoid forcing users to set DAPR_SIDECAR_IP for their Java App.

artursouza: I've just looked at the current code in master for runtime and this is the default behavior of the sidecar:
+ Standalone mode: listen to all addresses
+ K8s mode: listen to [::1],127.0.0.1

So, this approach is fine as of now. Let's keep it. User can override this setting if he/she decided to override the default listen ports.

willtsai: @artursouza - awesome, thanks for looking into this. I'd like to confirm here before proceeding:
1. Are you saying that the approach of setting DEFAULT_SIDECAR_IP to the loopback address via `InetAddress.getLoopbackAddress()` is fine?
2. Or are you thinking we need to set DEFAULT_SIDECAR_IP to the localhost address via `InetAddress.getLocalHost()`?

Given the default behavior of the sidecar you've determined, I believe that Option 1 would be the correct approach given that it listens to the loopback localhost addresses in K8s mode. Please **let me know if you also agree with Option 1**, thanks!

willtsai:
@mukundansundar @artursouza -- this is ready for review now, can you please take a look and if ok can remove the `do-not-merge` label?

mukundansundar:
@willtsai The integration tests are failing due to compilation error ...

willtsai:
> @willtsai The integration tests are failing due to compilation error ...

@mukundansundar - weird, I was able to compile and pass tests locally...looks like it's one specific failure `package Properties does not exist` - I'll investigate it. Probably a package dependency issue in the integration test class.

sdk/src/main/java/io/dapr/config/Properties.java
johnou: could have a little bit of logic here depending on the system properties `java.net.preferIPv4Stack` and `java.net.preferIPv6Addresses`

willtsai: @johnou - thanks for the suggestion, though I do wonder if this extra logic is really needed for `DEFAULT_SIDECAR_IP` since the user can override the default IP address for their sidecar through the `dapr.sidecar.ip` property or `DAPR_SIDECAR_IP` env variable, which can be either IPv4 or IPv6 based on their preference. Does this make sense, or am I interpreting your feedback incorrectly?

willtsai:
> > @willtsai The integration tests are failing due to compilation error ...
> 
> @mukundansundar - weird, I was able to compile and pass tests locally...looks like it's one specific failure `package Properties does not exist` - I'll investigate it. Probably a dependency issue due to missing import statement in the integration test class.

@mukundansundar - looks like it's passed integration tests and compiling successfully now. Also, thanks for restarting the validator jobs, looks like the second runs succeeded. I think this PR is ready for review now, you can remove the `do-not-merge` label if appropriate.

sdk/src/main/java/io/dapr/config/Properties.java
johnou: makes sense, I'd keep it simple given they can override the address.

sdk-actors/src/test/java/io/dapr/actors/TestUtils.java
```diff
@@ -54,4 +55,12 @@ public static void assertThrowsDaprException(
     Assertions.assertEquals(expectedErrorCode, daprException.getErrorCode());
     Assertions.assertEquals(expectedErrorMessage, daprException.getMessage());
   }
+
+  public static String getSidecarIpForHttpUrl(final String sidecarIp) {
+    String retSidecarIp = sidecarIp;
+    if(InetAddressValidator.getInstance().isValidInet6Address(sidecarIp)) {
+      retSidecarIp = "[" + sidecarIp + "]";
```
mukundansundar, line 62: @willtsai Can you add a comment in this method which points to the format expected for IP v6?

mukundansundar (commented):
added some comments. overall lgtm.

sdk-actors/src/test/java/io/dapr/actors/TestUtils.java
willtsai: @mukundansundar - done, let me know if it looks good.

mukundansundar (commented):
lgtm. @artursouza Please review ...

sdk-actors/pom.xml
```diff
@@ -67,6 +67,11 @@
       <version>${grpc.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-validator</groupId>
```
artursouza, line 71: We cannot add 3rd party dependencies. These can conflict with app dependencies. If this is needed for tests only, update the scope.

sdk/pom.xml
```diff
@@ -103,6 +103,11 @@
       <version>${grpc.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-validator</groupId>
+      <artifactId>commons-validator</artifactId>
```
artursouza, line 108: Same

artursouza (changes requested):
I will give another pass tomorrow. Thanks.

sdk/src/main/java/io/dapr/utils/NetworkUtils.java
```diff
@@ -47,4 +49,12 @@ public static void waitForSocket(String host, int port, int timeoutInMillisecond
       }
     }, timeoutInMilliseconds);
   }
+
+  /**
+   * Retrieve loopback address for the host.
+   * @return The loopback address String
+   */
+  public static String getHostLoopbackAddress() {
+    return InetAddress.getLoopbackAddress().getHostAddress();
```
artursouza, line 58: Can this method automatically return the Ipv6 between "[]"?

artursouza (changes requested):
Just a minor change.

sdk/src/main/java/io/dapr/utils/NetworkUtils.java
willtsai: @artursouza  Sorry for the delay, finally had some time to look into this. I changed some unit tests locally to validate IPv6 sidecar IP constants contained within "[]" and it seems like your suggestion is viable - for both URL endpoint invocation formats and Dapr client instantiation. Great recommendation!

However, before I proceed, I need to validate an implementation detail with you. To test for whether an address is IPv6, it would be convenient to leverage the Apache Commons Validator, though you have recommended against bringing in 3P libraries beyond the test scope. Is it possible to get an exception for this case, or do I have to write a custom IPv6 validator within NetworkUtils?

mukundansundar: @willtsai Is it complex to write a validator for IPv6 ? both standard pattern and hex compressed pattern along with IPv4 validation? 
I am wary of adding a new dependency specifically for one validation ....

willtsai: Understood - should not be too complex but I wanted to see if the work (and subsequent unit tests) could be avoided. Seems like adding a new dependency is not worth it here, so I'll work on the custom validator.

cicoyle:
@willtsai mind resolving the conflicts?

willtsai:
> @willtsai mind resolving the conflicts?

done, though now i need to investigate the test failures...

willtsai:
@cicoyle i've addressed all merge conflicts and test failures, can you please take another look? (i think the build tests just need to be re-run)

artursouza (approved):

cicoyle (approved):
lgtm - nice to see this coming to a closing 🎉

artursouza (approved):

willtsai:
woo! thanks @cicoyle @artursouza 🥳

Issue #694, opened 2022-02-24
spring-boot: Using @Topic annotation reads the path from methods' @PostMapping, ignoring class-level @RequestMapping


ittays:
## Expected Behavior

When annotating a controller with `@RequestMapping` and a method with `@PostMapping`  actual HTTP path is the combination of both.

So, when annotating the method with `@Topic` I expect it to map the topic with the full path.

## Actual Behavior

The topic is mapped with a truncated path, which is only the `@PostMapping` value.
Publishing to the topic constantly fail, as path not found.

## Steps to Reproduce the Problem

Create a controller and start the app:

```java
@Controller
@RequestMapping(value = "/v1")
public class MyController {
    @Topic(pubsubName = "pubsub", name = "my-topic")
    @PostMapping("foo")
    public void foo(@RequestBody(required = false) CloudEvent<String> cloudEvent) {
        log.trace("{}", cloudEvent.getData());
    }
}
```

Publish with `dapr publish --publish-app-id myapp --pubsub pubsub --topic my-topic --data hello-world`.

Observe sidecar error: `"non-retriable error returned from app while processing pub/sub event 9d8b9f70-9985-4665-8ce8-1d63a7935977: {\"timestamp\":1645712855614,\"status\":404,\"error\":\"Not Found\",\"message\":\"No message available\",\"path\":\"/foo\"}. status code returned: 404"`.

Error referrs to `"path": "/foo"`, but actual path is `/v1/foo`.

The code that can be improved is on https://github.com/dapr/java-sdk/blob/master/sdk-springboot/src/main/java/io/dapr/springboot/DaprBeanPostProcessor.java 

## Release Note

RELEASE NOTE: **FIX** Topic annotation handles class-level @RequestMapping

artursouza:
Thanks for reporting this bug with such level of detail.

artursouza:
Agreed.

DeepanshuA:
Assigning this to myself. Please feel free to re-assign if someone has already started work on this.

ittays:
Well done! 🌟

PR #700, opened 2022-03-09
Catering to all possible routes for Post Request

DeepanshuA:
Signed-off-by: deepanshuagarwal <deepanshu.agarwal1984@gmail.com>

# Description

A topic is getting subscribed to in Dapr with a truncated path, in case a class level RequestMapping is also provided with PostMapping path. This change aims at fixing this by taking into account all paths possible for POST.

## Issue reference

Please reference the issue this PR will close: #694 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mukundansundar:
@DeepanshuA Please see to the indentation ... Viewing via GitHub is hard ... please check if the checkstyle passes locally ...

mukundansundar:
@DeepanshuA Please see to the indentation ... Viewing via GitHub is hard ... please check if the checkstyle passes locally ...

mukundansundar (commented):
lgtm. @DeepanshuA Can you look into adding Unit Tests for the same?

mukundansundar:
@DeepanshuA 
Can you fix DCO ? Also is the PR complete or WIP ?

DeepanshuA:
> @DeepanshuA Please see to the indentation ... Viewing via GitHub is hard ... please check if the checkstyle passes locally ...

yeah @mukundansundar, fixed indentation

DeepanshuA:
> @DeepanshuA Please see to the indentation ... Viewing via GitHub is hard ... please check if the checkstyle passes locally ...

fixed indentation

DeepanshuA:
> lgtm. @DeepanshuA Can you look into adding Unit Tests for the same?

yups, added, was in WIP actually before this

DeepanshuA:
> @DeepanshuA Can you fix DCO ? Also is the PR complete or WIP ?

yeah @mukundansundar, missed out on sign-off part, let me fix it; yeah, will change the title - it will be PR complete after DCO change.

sdk-springboot/src/main/java/io/dapr/springboot/DaprBeanPostProcessor.java
```diff
@@ -80,27 +85,25 @@ private static void subscribeToTopics(Class clazz, EmbeddedValueResolver embedde
...
-
       String topicName = embeddedValueResolver.resolveStringValue(topic.name());
       String pubSubName = embeddedValueResolver.resolveStringValue(topic.pubsubName());
       if ((topicName != null) && (topicName.length() > 0) && pubSubName != null && pubSubName.length() > 0) {
         try {
           TypeReference<HashMap<String, String>> typeRef
                   = new TypeReference<HashMap<String, String>>() {};
           Map<String, String> metadata = MAPPER.readValue(topic.metadata(), typeRef);
-          DaprRuntime.getInstance().addSubscribedTopic(pubSubName, topicName, route, metadata);
+          List<String> routes = SpringProcessorHelper.getAllCompleteRoutesForPost(clazz, method, topic.name());
```
mukundansundar, line 95:
```suggestion
          List<String> routes = SpringProcessorHelper.getAllCompleteRoutesForPost(clazz, method, topicName);
```
Always use the resolved name.

mukundansundar (changes requested):
small change

sdk-springboot/src/main/java/io/dapr/springboot/DaprBeanPostProcessor.java
DeepanshuA: Agreed.

mukundansundar (approved):

mukundansundar:
cc @artursouza

sdk-springboot/src/main/java/io/dapr/springboot/util/SpringProcessorHelper.java
```diff
@@ -0,0 +1,96 @@
...
+  /**
+   * Method to provide all possible complete routes list fos this post method present in this controller class,
+   * for mentioned topic.
+   *
+   * @param clazz     Controller class
+   * @param method    Declared method for posting data
+   * @param topicName Associated topic name
+   * @return All possible routes for post mapping for this class and post method
+   */
+  public static List<String> getAllCompleteRoutesForPost(Class clazz, Method method, String topicName) {
```
artursouza, line 38: Please, move this as a private method in DaprBeanPostProcessor. The reason is that this design will make this method public and part of our contract - in other words, users can depend on this method. Marking it as private makes it clear that is internal use only. Unfortunately, we don't have an internal visibility, only protected, public, private and default.

sdk-springboot/src/test/java/io/dapr/springboot/util/SpringProcessorHelperTest.java
```diff
@@ -0,0 +1,60 @@
...
+        {MockControllerNoClazzAnnotation.class, "testMethod3", new String[] {"foo"}, true},
+        {MockControllerNoClazzAnnotation.class, "testMethod4", new String[] {"foo1", "foo2"}, true},
+        {MockControllerNoClazzAnnotation.class, "testMethod5", new String[] {TOPIC_NAME}, true}
+    });
+  }
+
+
+  @Test
+  public void testAllPostRoutesGeneration() throws NoSuchMethodException {
+    List<String> routesArrayTestMethod1 = SpringProcessorHelper.getAllCompleteRoutesForPost(clazzToBeTested,
```
artursouza, line 50: After this method is made private, you can still test it via reflection.

artursouza (changes requested):

mukundansundar:
@DeepanshuA Do you know why codecov is giving potentially incorrect results ?

mukundansundar (approved):
lgtm

sdk-springboot/src/main/java/io/dapr/springboot/util/SpringProcessorHelper.java
DeepanshuA: Agreed that this will expose it without need of exposing. Though, just thinking that at least taking out some methods with default access (package private)/ protected may be required, if refactoring a huge class OR using a required design pattern may force us making some extra methods/interfaces public sometimes. BUT, that's not the case with DaprBeanPostProcessor as of now. So, making the suggested change.

sdk-springboot/src/test/java/io/dapr/springboot/util/SpringProcessorHelperTest.java
DeepanshuA: Done.

artursouza (approved):

PR #697, opened 2022-02-28
Query State Preview API implementation.

mukundansundar:
Signed-off-by: Mukundan Sundararajan <msundar.ms@outlook.com>

# Description

changes made for query state API. 

The dependency updates are needed for Java SDK to successfully compile using Rosetta.

Have used MongoDb for querying stored state. 

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #653 
closes #685 
closes #702 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation

PR #705, opened 2022-03-15
Add IntelliJ IT docs

mukundansundar:
Signed-off-by: Mukundan Sundararajan <msundar.ms@outlook.com>

# Description

Added docs in README to explain on how to run Integration tests. To be merged after #697 since this PR refers to a file `local-test-mongo.yml` created in that PR. 

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


README.md
```diff
@@ -207,3 +207,69 @@ Change the properties below in [pom.xml](./pom.xml) to point to the desired refe
...
+### Running Integration Tests
+
+#### Pre-Requisites for ITs
+Along with the pre-requisites for [SDK](#pre-requisites) the following are needed. 
+
+* Docker installed 
+  * [Docker Compose](https://docs.docker.com/compose/install/) 
+  * [Docker Desktop](https://www.docker.com/products/docker-desktop)
+* Bash shell
+  * In Windows use [WSL](https://docs.microsoft.com/en-us/windows/wsl/install)
```
DeepanshuA, line 220: It may be better to recommend for WSL2, as most of the folks here in team are using WSL2.
https://howto.lintel.in/wsl-vs-wsl-2-performance/
https://docs.microsoft.com/en-us/windows/wsl/compare-versions

mukundansundar: In the docs it is given 
>New Linux installations will be set to WSL 2 by default, but the wsl --set-version command can be used to downgrade from WSL 2 to WSL 1 as well. To see whether your Linux distribution is set to WSL 1 or WSL 2, use the command: wsl -l -v.


I think the default installation is wsl2 ?

DeepanshuA: Yeah, as per docs that seems to be the case. I don't remember exact commands, but my default implementation was WSL1 when I had installed a month back. But, anyways, we have already used WSL2 in the documentation now.

artursouza (approved):

PR #708, opened 2022-03-18
set maxRequestsPerHost of okhttp to support slow response requests and high TPS

skyao:
# Description

In dapr java sdk, there are some default settings in okhttp which we don't set and use the default value:

- maxRequests: default to 64
- maxRequestsPerHost: default to 5! 
- maxIdleConnections

The default value  5 of maxRequestsPerHost is totally unacceptable, it will lead to very serious problems of very low throughput as discribed in issue dapr/java-sdk#709.

And the default value 64 of maxRequests is not enough for the user case of high TPS. For example, if the response time is 100 ms, 64 connections leads to at most 640 TPS. 

## Issue reference

This PR is to fix this issue.

dapr/java-sdk#709

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
+++ b/sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
@@ -14,9 +14,12 @@ limitations under the License.
 package io.dapr.client;
 
 import io.dapr.config.Properties;
+import okhttp3.ConnectionPool;
+import okhttp3.Dispatcher;
 import okhttp3.OkHttpClient;
 
 import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 
 /**
  * A builder for the DaprHttp.
@@ -32,6 +35,13 @@ public class DaprHttpBuilder {
    * Static lock object.
    */
   private static final Object LOCK = new Object();
+
+  /**
+   * HTTP keep alive duration in seconds.
+   *
+   * <p>Just hard code to a reasonable value.
+   */
+  private static final int KEEP_ALIVE_DURATION = 30;
 
   /**
    * Build an instance of the Http client based on the provided setup.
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
```diff
@@ -55,6 +65,19 @@ private DaprHttp buildDaprHttp() {
           OkHttpClient.Builder builder = new OkHttpClient.Builder();
           Duration readTimeout = Duration.ofSeconds(Properties.HTTP_CLIENT_READ_TIMEOUT_SECONDS.get());
           builder.readTimeout(readTimeout);
+
+          Dispatcher dispatcher = new Dispatcher();
+          dispatcher.setMaxRequests(Properties.HTTP_CLIENT_MAX_REQUESTS.get());
+          // The maximum number of requests for each host to execute concurrently.
+          // Default value is 5 in okjava which is totally UNACCEPTABLE!
```
mukundansundar, line 72:
```suggestion
          // Default value is 5 in okHttp which is totally UNACCEPTABLE!
```

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
```diff
@@ -33,6 +36,13 @@
    */
   private static final Object LOCK = new Object();
 
+  /**
+   * HTTP keep alive duration in seconds.
+   *
+   * <p>Just hard code to a reasonable value.
+   */
+  private static final int KEEP_ALIVE_DURATION = 30;
```
mukundansundar, line 44: Make this also a configurable parameter?

mukundansundar (commented):
@skyao Keep in mind that in future we will be moving away from OkHttp and move to native Java Client. But I believe these parameters will still add value there.

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
skyao: I think this value can be safe to hard code. 

I have to set this because it's required by the constructor:

ConnectionPool pool = new ConnectionPool(Properties.HTTP_CLIENT_MAX_IDLE_CONNECTIONS.get(),
                  KEEP_ALIVE_DURATION, TimeUnit.SECONDS);

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
skyao: Correct.

skyao:
> @skyao Keep in mind that in future we will be moving away from OkHttp and move to native Java Client. But I believe these parameters will still add value there.

OK. 

Before we changed to native java client, we still need OkHttp. The fix for default maxRequestPerHost is very important.

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
artursouza: We will remove OKHTTP dependency, so it is OK to hardcode this one for now.

artursouza (dismissed):

artursouza:
@skyao Please, sign off the commits. This is a new requirement in our repos since the CNCF donation.

mukundansundar (dismissed):

sdk/src/main/java/io/dapr/config/Properties.java
```diff
@@ -56,7 +56,24 @@
...
    * Dapr's default timeout in seconds for HTTP client reads.
    */
-  private static final Integer DEFAULT_HTTP_CLIENT_READTIMEOUTSECONDS = 60;
+  private static final Integer DEFAULT_HTTP_CLIENT_READ_TIMEOUT_SECONDS = 60;
+
+  /**
+   *   Dapr's default maximum number of requests for HTTP client to execute concurrently.
+   *
+   *   <p>Above this requests queue in memory, waiting for the running calls to complete.
+   *   Default is 64 in okjava which is OK for most case, but for some special case
```
mukundansundar, line 65: nit :
```suggestion
   *   Default is 64 in okhttp which is OK for most case, but for some special case

```

skyao: typo, fixed.

mukundansundar (approved):
lgtm

Issue #701, opened 2022-03-14
Run build and validations with Java 17


mukundansundar:
## Describe the proposal
With Java 17 being released, change `build.yaml` and `validate.yaml` to include validations against [Java 17](https://jdk.java.net/17/).

PR #710, opened 2022-03-21
Run Build and Validate against java 17

tanvigour:
Signed-off-by: tanvigour <tanvi.gour@gmail.com>

# Description

## Issue reference

Please reference the issue this PR will close: #701

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


mukundansundar (commented):
Can we remove 13, 15 ,16 ? They are out of support ... 
cc @artursouza thoughts ?

mukundansundar:
@tanvigour I think we need to probably move to j11 and also might need to update some of the dependencies for the build to succeed in j17 ...

artursouza (changes requested):
FindBugs is failing. You might need to update this dependency too.

mukundansundar:
> > Can we remove 13, 15 ,16 ? They are out of support ... cc @artursouza thoughts ?
> 
> Let's keep those for now since some devs might still be using them, also there is not really any cost in running those.

The cost is there is still some flakiness in our tests and when running them in parallel across different version, sometimes they fail in one of them and causes all the other executions to cancel.

pom.xml
```diff
@@ -26,7 +26,7 @@
     <maven.deploy.skip>true</maven.deploy.skip>
 
     <gpg.skip>true</gpg.skip>
-    <spotbugs.version>4.0.0-RC1</spotbugs.version>
+    <spotbugs.version>4.6.0</spotbugs.version>
     <spotbugs.fail>true</spotbugs.fail>
```
tanvigour, line 30: @mukundansundar , @artursouza : It looks like we have enabled spotbugs to fail on error and interrupt the build. Do we want it to block the build?  In newer version it has converted all the warnings to errors.

artursouza: If we upgrade, then we need to fix the spotbugs errors.

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ 11, 13, 15, 16 ]
```
artursouza, line 21: Please, keep 13 and 15 since we are not winning anything by removing those.

mukundansundar: We are ... in the sense we are removing some flakiness in the test suite ... not necessarily the tests themselves ...

Issue #712, opened 2022-03-23
Actor DurationUtils parsing of duration throws an exception


mukundansundar:
## Expected Behavior

Duration is parsed correctly and no exception is thrown

## Actual Behavior

Following Actor API reference in docs : https://docs.dapr.io/reference/api/actors_api/#examples-3

if a reminder is created in the format 
```sh
curl http://localhost:3500/v1.0/actors/stormtrooper/50/reminders/checkRebels \
  -H "Content-Type: application/json"
-d '{
      "data": "someData",
      "dueTime": "1m",
      "period": "20s"
    }'
```

And used in a Java with an Actor Type names stormtrooper and the Actor class implements Remindable<String> data ... 
```java
public Mono<Void> receiveReminder(String reminderName, String state, Duration dueTime, Duration period) {
```
When receiveReminder is called by the actor runtime, the dueTime field is parsed and converted to Java Duration using the class 
DurationUtils.java

```java
 public static Duration convertDurationFromDaprFormat(String valueString) {
    // Convert the format returned by the Dapr runtime into Duration
    // An example of the format is: 4h15m50s60ms. It does not include days.
    int hourIndex = valueString.indexOf('h');
    int minuteIndex = valueString.indexOf('m');
    int secondIndex = valueString.indexOf('s');
    int milliIndex = valueString.indexOf("ms");

    String hoursSpan = valueString.substring(0, hourIndex);

    int hours = Integer.parseInt(hoursSpan);
    int days = hours / 24;
    hours = hours % 24;

    String minutesSpan = valueString.substring(hourIndex + 1, minuteIndex);
    int minutes = Integer.parseInt(minutesSpan);

    String secondsSpan = valueString.substring(minuteIndex + 1, secondIndex);
    int seconds = Integer.parseInt(secondsSpan);

    String millisecondsSpan = valueString.substring(secondIndex + 1, milliIndex);
    int milliseconds = Integer.parseInt(millisecondsSpan);

    return Duration.ZERO
        .plusDays(days)
        .plusHours(hours)
        .plusMinutes(minutes)
        .plusSeconds(seconds)
        .plusMillis(milliseconds);
  }
```

In this an exception is thrown as such: 
```
[...]

PR #714, opened 2022-03-24
fix duration parsing

mukundansundar:
# Description

Fix the Duration parsing. When reminder dueTime is stored in DB as "30m" or "10s" etc ... Java SDK fails to load the reminder with IndexOutOfBoundException. 

## Issue reference

Please reference the issue this PR will close: #712 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/utils/DurationUtils.java
+++ b/sdk/src/main/java/io/dapr/utils/DurationUtils.java
@@ -26,32 +26,56 @@ public class DurationUtils {
   public static Duration convertDurationFromDaprFormat(String valueString) {
     // Convert the format returned by the Dapr runtime into Duration
     // An example of the format is: 4h15m50s60ms. It does not include days.
+    if (valueString == null) {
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/dapr/utils/DurationUtils.java
```diff
@@ -65,8 +89,7 @@ public static String convertDurationToDaprFormat(Duration value) {
 
     // return empty string for anything negative, it'll only happen for reminder "periods", not dueTimes.  A
     // negative "period" means fire once only.
-    if (value == Duration.ZERO
-        || (value.compareTo(Duration.ZERO) == 1)) {
```
mukundansundar, line 69: `compareTo` can produce any value greater than 1, by contract any positive value means the value is greater. This failed the test when only milliseconds were given ...

artursouza (approved):

Issue #667, opened 2021-12-09
Remove Hashicorp Vault from Secret Store example


artursouza:
## Describe the proposal

There was a change in Hashicorp Vault output that broke the example validation. Instead, use the local file secret store so there is no need to install Hashicorp Vault anymore.

This work needs to make sure the instructions in the README.md file for the example is reworded to explain how the secret is retrieved. In addition, it should point to the documentation in Dapr docs website where other secret store components can be used.

Change should also remove Vault installation from the validate.yaml workflow file.

PR #719, opened 2022-03-30
replace vault to local

dntbeSris:
Signed-off-by: zjx244729 <zjx244729@alibaba-inc.com>

# Description

Replace the Secret Store from Hashicorp Vault to Local File

## Issue reference

Please reference the issue this PR will close: #667 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/build.yml
```diff
@@ -87,10 +87,6 @@ jobs:
       run: |
         docker-compose -f ./sdk-tests/deploy/local-test-kafka.yml up -d
         docker ps
-    - name: Install Local Hashicorp Vault using docker-compose
```
mukundansundar, line 90: have the vault install in the build.yaml and remove it from the validate.yaml file ....

mukundansundar (changes requested):
@dntbeSris One change ... The vault install is removed from the validate.yaml file. 
Because it was removed from build.yaml it is breaking the build.

.github/workflows/build.yml
dntbeSris: ok, and the Integration Test was fail because the vault secret test case. should i fix it ?

mukundansundar: Yes please @dntbeSris 
cc @artursouza

dntbeSris: done

mukundansundar (changes requested):
in the integration tests why have the secrets file be created on the file. I think the test local secrets file can be committed as a file in the components directory itself ... 
This is currently causing an issue with other tests also since the component in being loaded in other tests also and it is not finding the secrets file which is only created on the fly for this test ... so dapr initialising is failing because the component is not being loaded ... 

Changing it to a file that exists in the components directory from start is fine ...

dntbeSris:
> 

yep! i create and delete the secret file just in secret test program at first.but it made the other tests failed . because it always load all components . so i created the file earlier. this change in commit [9f21553](https://github.com/dapr/java-sdk/pull/719/commits/9f215534d19dd0a2e34dd5fe8ba158624ab8aac0)

dntbeSris:
> in the integration tests why have the secrets file be created on the file. I think the test local secrets file can be committed as a file in the components directory itself ... This is currently causing an issue with other tests also since the component in being loaded in other tests also and it is not finding the secrets file which is only created on the fly for this test ... so dapr initialising is failing because the component is not being loaded ...
> 
> Changing it to a file that exists in the components directory from start is fine for integration tests ...

what should i do next ？

Issue #698, opened 2022-02-28
Add Unsubscribe API for Configuration building block


mukundansundar:
https://github.com/dapr/dapr/pull/4130 changes proto files to add unsubscribe API as well as additional fields to existing configuration API response. 

Implement the unsubscribe API as well as add in the additional fields necessary for the change to include ID field in subscribe API to track for stopping subscription. 

## Release Note

RELEASE NOTE:  Add unsubscribe API.

mukundansundar:
@artursouza

pravinpushkar:
@mukundansundar Verified this with dapr master and proto file, it works fine. There is no change needed at Java-sdk level.

We can close this issue.

mukundansundar:
@pravinpushkar Interesting ... We are not using the full SubscribeConfigurationResponse anywhere ... Which is kind of expected since it only contained List<ConfigurationItem> till the above PR. 
Now it has an ID field too ... which is used for unsubscribe API.

I will keep this issue open and change the title to implement unsubscribe API. Then we would need the proto changes that were made in dapr/dapr. https://github.com/dapr/dapr/pull/4130/files#diff-da511bac0105b1e517ff78c6351b68bda0d5862cb0daae89127f6c0250551265R528

mukundansundar:
I have added a breaking changes label since I believe the exposed API in the preview client will change to include the SubscribeConfigurationResponse.

artursouza:
Are we planning this for 1.5 still?

PR #740, opened 2022-05-06
Config API : HTTP APIs and Unsubscribe gRPC API changes

pravinpushkar:
Signed-off-by: pravinpushkar <ppushkar@microsoft.com>

# Description

Added HTTP API suport for get, subscribe and unsubscribe config APIs
Added unsusbcribe gRPC API support
Added domain class for subscribe response
Added domain classes for unsubscribe request and response

## Issue reference

Please reference the issue this PR will close: #698 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [] Extended the documentation


tanvigour (dismissed):
LGTM

pravinpushkar:
Ping @artursouza @mukundansundar

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -106,35 +111,83 @@ public static void getConfigurations(DaprPreviewClient client) {
...
+    // To ensure main thread does not die before outFlux subscribe gets called
+    inducingSleepTime(5000);
+  }
+
+  /**
+   * Unsubscribe using subscription id.
+   *
+   * @param client DaprPreviewClient object
+   */
+  public static void unsubscribeToConfigurationItems(DaprPreviewClient client) {
```
mukundansundar, line 150:
```suggestion
  public static void unsubscribeConfigurationItems(DaprPreviewClient client) {

```

and add code snippet to readme also.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -106,35 +111,83 @@ public static void getConfigurations(DaprPreviewClient client) {
...
+
+    // To ensure key starts getting updated
+    inducingSleepTime(1000);
+
+    UnsubscribeConfigurationResponse res = client.unsubscribeToConfiguration(
+        subscriptionId.get(),
+        CONFIG_STORE_NAME
+    ).block();
+    assert res != null;
+    System.out.println("IsUnsubscribed : " + res.getIsUnsubscribed());
```
mukundansundar, line 185:
```suggestion
    System.out.println("Unsubscribed from : " + res.getIsUnsubscribed());

```

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -106,35 +111,83 @@ public static void getConfigurations(DaprPreviewClient client) {
...
+    new Thread(updateKeys).start();
+
+    // To ensure key starts getting updated
+    inducingSleepTime(1000);
+
+    UnsubscribeConfigurationResponse res = client.unsubscribeToConfiguration(
+        subscriptionId.get(),
+        CONFIG_STORE_NAME
+    ).block();
+    assert res != null;
```
mukundansundar, line 184: check if (res != null) { sysout }
else {sys out not successful}
Do not use assert in any examples.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -106,35 +111,83 @@ public static void getConfigurations(DaprPreviewClient client) {
...
+    System.out.println("IsUnsubscribed : " + res.getIsUnsubscribed());
+  }
+
+  private static void inducingSleepTime(int timeInMillis) {
     try {
-      // To ensure main thread does not die before outFlux subscribe gets called
-      Thread.sleep(10000);
-      disposableAtomicReference.get().dispose();
```
mukundansundar, line 137: why is this line removed?

sdk-tests/src/test/java/io/dapr/it/configuration/grpc/ConfigurationClientIT.java
```diff
@@ -142,6 +144,64 @@ public void subscribeToConfiguration() {
         assertFalse(updatedValues.contains("update_myconfigvalue3"));
     }
 
+    @Test
+    public void unsubscribeToConfigurationItems() {
```
mukundansundar, line 148:
```suggestion
    public void unsubscribeFromConfigurationItems() {

```

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
```diff
@@ -560,6 +563,16 @@ public Flux<List<ConfigurationItem>> subscribeToConfiguration(
     return this.subscribeToConfiguration(request);
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  public Mono<UnsubscribeConfigurationResponse> unsubscribeToConfiguration(
```
mukundansundar, line 569:
```suggestion
  public Mono<UnsubscribeConfigurationResponse> unsubscribeFromConfiguration(
```

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -831,15 +834,49 @@ public Flux<List<ConfigurationItem>> subscribeToConfiguration(SubscribeConfigura
...
     } catch (Exception ex) {
       return DaprException.wrapFlux(ex);
     }
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Mono<UnsubscribeConfigurationResponse> unsubscribeToConfiguration(UnsubscribeConfigurationRequest request) {
```
mukundansundar, line 852:
```suggestion
  public Mono<UnsubscribeConfigurationResponse> unsubscribeFromConfiguration(UnsubscribeConfigurationRequest request) {
```

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
```diff
@@ -97,16 +100,33 @@ public interface DaprPreviewClient extends AutoCloseable {
...
-  Flux<List<ConfigurationItem>> subscribeToConfiguration(String storeName, List<String> keys,
-                                                         Map<String, String> metadata);
+  Flux<SubscribeConfigurationResponse> subscribeToConfiguration(String storeName, List<String> keys,
+                                                                Map<String, String> metadata);
 
   /**
    * Subscribe to the keys for any change.
    *
    * @param request request for subscribing to any change for the given keys in request
    * @return Flux of List of configuration items
```
mukundansundar, line 110: Change and link to the new class

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
```diff
@@ -97,16 +100,33 @@ public interface DaprPreviewClient extends AutoCloseable {
...
+  Flux<SubscribeConfigurationResponse> subscribeToConfiguration(SubscribeConfigurationRequest request);
+
+  /**
+   * Subscribe to the keys for any change.
+   *
+   * @param id subscription id returned by subscribeToConfiguration API.
+   * @param storeName Name of the configuration store.
+   * @return Mono of UnsubscribeConfigurationResponse.
+   */
+  Mono<UnsubscribeConfigurationResponse> unsubscribeToConfiguration(String id, String storeName);
```
mukundansundar, line 121: rename

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
```diff
@@ -97,16 +100,33 @@ public interface DaprPreviewClient extends AutoCloseable {
...
    */
-  Flux<List<ConfigurationItem>> subscribeToConfiguration(SubscribeConfigurationRequest request);
+  Flux<SubscribeConfigurationResponse> subscribeToConfiguration(SubscribeConfigurationRequest request);
+
+  /**
+   * Subscribe to the keys for any change.
+   *
+   * @param id subscription id returned by subscribeToConfiguration API.
+   * @param storeName Name of the configuration store.
+   * @return Mono of UnsubscribeConfigurationResponse.
```
mukundansundar, line 119: link to that class

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
```diff
@@ -97,16 +100,33 @@ public interface DaprPreviewClient extends AutoCloseable {
...
+   */
+  Mono<UnsubscribeConfigurationResponse> unsubscribeToConfiguration(String id, String storeName);
+
+  /**
+   * Subscribe to the keys for any change.
+   *
+   * @param request request for unsubscribing to any change for the given subscription id in request
+   * @return Mono of UnsubscribeConfigurationResponse
```
mukundansundar, line 127: link to that class

sdk/src/main/java/io/dapr/client/domain/SubscribeConfigurationResponse.java
```diff
@@ -0,0 +1,43 @@
...
+
+package io.dapr.client.domain;
+
+import java.util.List;
+
+/**
+ * Domain object for response from subscribeToConfiguration API.
+ */
+public class SubscribeConfigurationResponse {
+  private final String id;
```
mukundansundar, line 22: add java doc for the fields. Should this field be `subscriptionId` to make it explicit that it is the ID for the subscription and not necessary for individual configuration items\?

sdk/src/main/java/io/dapr/client/domain/SubscribeConfigurationResponse.java
```diff
@@ -0,0 +1,43 @@
...
+  private final List<ConfigurationItem> items;
```
mukundansundar, line 23: make field immutable.

sdk/src/main/java/io/dapr/client/domain/UnsubscribeConfigurationRequest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Request to unsubscribe to one or more configuration items using subscription id.
+ */
+public class UnsubscribeConfigurationRequest {
+  private final String storeName;
```
mukundansundar, line 24: add java doc for fields.

sdk/src/main/java/io/dapr/client/domain/UnsubscribeConfigurationResponse.java
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+*/
+
+package io.dapr.client.domain;
+
+/**
+ * Domain object for unsubscribe response.
+ */
+public class UnsubscribeConfigurationResponse {
+  private final boolean isUnsubscribed;
```
mukundansundar, line 20: add java doc

sdk/src/test/java/io/dapr/client/DaprPreviewClientGrpcTest.java
```diff
@@ -194,9 +198,11 @@ public void subscribeConfigurationTest() {
 			return null;
 		}).when(daprStub).subscribeConfigurationAlpha1(any(DaprProtos.SubscribeConfigurationRequest.class), any());
 
-		Iterator<List<ConfigurationItem>> itr = previewClient.subscribeToConfiguration(CONFIG_STORE_NAME, "configkey1").toIterable().iterator();
+		Iterator<SubscribeConfigurationResponse> itr = previewClient.subscribeToConfiguration(CONFIG_STORE_NAME, "configkey1").toIterable().iterator();
 		assertTrue(itr.hasNext());
-		assertEquals("configkey1", itr.next().get(0).getKey());
+		SubscribeConfigurationResponse res = itr.next();
```
mukundansundar, line 203: best to check it has next and then calling next.

mukundansundar (commented):
@artursouza Instead of `subscribeToConfiguration` and `unsubscribeFromConfiguration` should it just follow the Proto naming and go with `subscribeConfiguration` and `unsubscribeConfiguration` ? Thoughts? 
@pravinpushkar Added comments on refactoring the code.

sdk/src/test/java/io/dapr/client/DaprPreviewClientGrpcTest.java
pravinpushkar: Actually the test is based on this logic that the next item has to exist. See line just above this `assertTrue(itr.hasNext());`. If the next item is not there the test should fail.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
pravinpushkar: Actually this was not needed. As soon as the application will terminate the flux will also be disposed. Earlier I was disposing the flux before actually finishing the application.

sdk-tests/src/test/java/io/dapr/it/configuration/grpc/ConfigurationClientIT.java
```diff
@@ -154,4 +214,22 @@ private static void executeDockerCommand(String[] command) {
             e.printStackTrace();
         }
     }
+
+    private static void executeDockerCommand(String key, int postfix) {
```
mukundansundar, line 218: almost repetition of method `executeDockerCommand(String[] command)`. Please change to use only one format.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -770,10 +773,18 @@ public Mono<List<ConfigurationItem>> getConfiguration(GetConfigurationRequest re
    * {@inheritDoc}
    */
   @Override
-  public Flux<List<ConfigurationItem>> subscribeToConfiguration(SubscribeConfigurationRequest request) {
+  public Flux<SubscribeConfigurationResponse> subscribeConfiguration(SubscribeConfigurationRequest request) {
     return DaprException.wrapFlux(new UnsupportedOperationException());
```
mukundansundar, line 777: There is a HTTP version added since 1.8. https://docs.dapr.io/reference/api/configuration_api/

mukundansundar (changes requested):
please add the http API also ...

examples/src/main/java/io/dapr/examples/configuration/http/ConfigSubscriberController.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  /**
+   * Api mapping for subscribe configuration.
+   * @param pathVarsMap Path variables for post call
+   * @param obj request Body
+   * @return Returns void
+   */
+  @PostMapping(path = "/configuration/{configStore}/{key}", produces = MediaType.APPLICATION_JSON_VALUE)
+  public Mono<Void> handleConfigUpdate(@PathVariable Map<String, String> pathVarsMap, @RequestBody JsonNode obj) {
```
mukundansundar, line 42: I think this should be SubscriptionConfigurationReponse object?

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -763,15 +766,134 @@ private <T> QueryStateResponse<T> buildQueryStateResponse(DaprHttp.Response resp
...
+        try {
+          JsonNode root = INTERNAL_SERIALIZER.parseNode(response.getBody());
+          String subscriptionId = root.path("id").asText();
+          return new SubscribeConfigurationResponse(subscriptionId, new ArrayList<>());
+        } catch (IOException e) {
+          throw new RuntimeException(e);
+        }
+      }).block();
+      if (res != null) {
+        return Flux.just(res);
```
pravinpushkar, line 849: Here I could not think of better way of handelling this.
Reason -  We have common interface for http and gRPC. The gRPC api is implemented using stream data type in runtime. So, here we need to have Flux as return type but HTTP API is implemented differently in runtime which just returns a subscription id and then dapr makes a callback to app endpoint for getting the subscribed changes. So here Mono return type is appropriate. Thus the above logic of returning just one item as Flux.

pravinpushkar:
I will rebase this PR again when list to map changes are merged with following PR -  https://github.com/dapr/java-sdk/pull/769

examples/src/main/java/io/dapr/examples/configuration/http/ConfigSubscriberController.java
mukundansundar: @pravinpushkar thoughts?

pravinpushkar: @mukundansundar It should be `SubscriptionConfigurationReponse`. I am facing some issue in deserialising the req body. I am looking into it.

sdk-tests/src/test/java/io/dapr/it/configuration/http/ConfigSubscriberController.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  /**
+   * Api mapping for subscribe configuration.
+   * @param pathVarsMap Path variables for post call
+   * @param obj request Body
+   * @return Returns void
+   */
+  @PostMapping(path = "/configuration/{configStore}/{key}", produces = MediaType.APPLICATION_JSON_VALUE)
+  public Mono<Void> handleMessage(@PathVariable Map<String, String> pathVarsMap, @RequestBody JsonNode obj) {
```
mukundansundar, line 42: @pravinpushkar This should be `SubscribeConfigurationResponse` as well

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -768,15 +772,140 @@ private <T> QueryStateResponse<T> buildQueryStateResponse(DaprHttp.Response resp
...
+      if ((configurationStoreName == null) || (configurationStoreName.trim().isEmpty())) {
+        throw new IllegalArgumentException("Configuration Store Name cannot be null or empty.");
+      }
+      if (keys.isEmpty()) {
+        throw new IllegalArgumentException("Keys can not be empty or null");
+      }
+
+      String[] pathSegments = new String[] {DaprHttp.ALPHA_1_API_VERSION, "configuration", configurationStoreName };
+      Map<String, List<String>> queryParams = new HashMap<>();
+      queryParams.put("key", Collections.unmodifiableList(keys));
```
mukundansundar, line 787: @pravinpushkar Metadata is got from query args as well ... so that needs to be populated here ...
See https://github.com/dapr/dapr/blob/master/pkg/http/api.go#L1036

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -768,15 +772,140 @@ private <T> QueryStateResponse<T> buildQueryStateResponse(DaprHttp.Response resp
...
+        throw new IllegalArgumentException("Configuration Store Name can not be null or empty.");
+      }
+      if (keys.isEmpty()) {
+        throw new IllegalArgumentException("Keys can not be null or empty.");
+      }
+
+      String[] pathSegments =
+              new String[] { DaprHttp.ALPHA_1_API_VERSION, "configuration", configurationStoreName, "subscribe" };
+      Map<String, List<String>> queryParams = new HashMap<>();
+      queryParams.put("key", Collections.unmodifiableList(keys));
```
mukundansundar, line 844: metadata is got from query args and must be populated here

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
mukundansundar: looks good ... @artursouza I think this is good enough for now ... any concerns?

sdk/src/main/java/io/dapr/client/domain/ConfigurationItem.java
```diff
@@ -20,10 +20,17 @@
  * A configuration item from Dapr's configuration store.
  */
 public class ConfigurationItem {
-  private final String key;
-  private final String value;
-  private final String version;
-  private final Map<String, String> metadata;
+  private String key;
```
mukundansundar, line 23: why is the final modifier removed?

sdk/src/main/java/io/dapr/client/domain/ConfigurationItem.java
```diff
@@ -20,10 +20,17 @@
...
-  private final Map<String, String> metadata;
+  private String key;
+  private String value;
+  private String version;
+  private Map<String, String> metadata;
+
+  /**
+   * Default constructor for ConfigurationItem.
+   */
+  public ConfigurationItem() {
```
mukundansundar, line 31: why is a default constructor needed?

mukundansundar (dismissed):
metadata changes are needed and the gRPC IT should also have the proper response object.

sdk/src/main/java/io/dapr/client/domain/ConfigurationItem.java
pravinpushkar: This is to make jackson library able to parse the Request body to `SubscribeConfigurationResponse ` sent by dapr to app endpoint in sdk/examples. Jackson needed the `no Arg constructor` and then since we are not actually doing any setting/ initialisation, it was complaining to initialise the final variables.

.github/workflows/build.yml
```diff
@@ -25,8 +25,8 @@ jobs:
       GOARCH: amd64
       GOPROXY: https://proxy.golang.org
       JDK_VER: ${{ matrix.java }}
-      DAPR_CLI_VER: 1.8.0-rc.1
-      DAPR_RUNTIME_VER: 1.8.0-rc.3
+      DAPR_CLI_VER: 1.8.1
+      DAPR_RUNTIME_VER: 1.8.4
```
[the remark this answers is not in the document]
pravinpushkar: @mukundansundar  yes thats correct but should we not update this as well to latest release cli and runtime ?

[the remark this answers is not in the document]
artursouza: This is fine. We will change to next RC soon.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -145,11 +201,11 @@ private static void print(ConfigurationItem item, String key) {
     System.out.println(item.getValue() + " : key ->" + key);
   }
 
-  private static void executeDockerCommand(int postfix) {
+  private static void executeDockerCommand(String key, int postfix) {
```
artursouza, line 204: Please, move this `docker exec` logic to README.md. This is not a good example to follow for a real app.

examples/src/main/java/io/dapr/examples/configuration/http/ConfigSubscriberController.java
```diff
@@ -0,0 +1,60 @@
...
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  /**
+   * Api mapping for subscribe configuration.
+   * @param pathVarsMap Path variables for post call
+   * @param obj request Body
+   * @return Returns void
+   */
+  @PostMapping(path = "/configuration/{configStore}/{key}", produces = MediaType.ALL_VALUE)
+  public Mono<Void> handleConfigUpdate(
```
artursouza, line 44: Is there a way to move this to our SpringBoot SDK? We do that for Actors already so Spring Boot Apps don't need to implement this by themselves.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -106,36 +110,88 @@ public static void getConfigurations(DaprPreviewClient client) {
...
-    } catch (InterruptedException e) {
-      e.printStackTrace();
-    }
+
+    // To ensure that subscribeThread gets scheduled
+    inducingSleepTime(0);
+
     Runnable updateKeys = () -> {
       int i = 1;
       while (i <= 3) {
```
artursouza, line 134: Move this logic to README.md.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
```diff
@@ -106,36 +110,88 @@ public static void getConfigurations(DaprPreviewClient client) {
...
-    System.out.println("*****Subscribing to keys using subscribe method: " + keys.toString() + " *****");
+  public static void subscribeConfigurationRequest(DaprPreviewClient client) {
+    System.out.println("Subscribing to key: myconfig1");
     AtomicReference<Disposable> disposableAtomicReference = new AtomicReference<>();
-    SubscribeConfigurationRequest req = new SubscribeConfigurationRequest(CONFIG_STORE_NAME, keys);
+    SubscribeConfigurationRequest req = new SubscribeConfigurationRequest(
+        CONFIG_STORE_NAME, Collections.singletonList("myconfig1"));
     Runnable subscribeTask = () -> {
-      Flux<Map<String, ConfigurationItem>> outFlux = client.subscribeToConfiguration(req);
+      Flux<SubscribeConfigurationResponse> outFlux = client.subscribeConfiguration(req);
```
artursouza, line 119: I think this is too complicated to use this feature in the SDK. We should follow up with a solution that wraps the update of the key into a wrapper class: DaprSubscribedConfig or something. Then, the app just needs to get the value and it is automatically updated behind the scenes. The app can also register a callback function in that instance to receive a callback when it is updated.

examples/src/main/java/io/dapr/examples/configuration/http/ConfigurationClient.java
```diff
@@ -0,0 +1,159 @@
...
+import reactor.core.publisher.Mono;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ConfigurationClient {
```
artursouza, line 35: This example is too complex. We don't need to cover every single use case in an example.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
artursouza: I am fine with this.

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -27,6 +27,7 @@
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
 import org.jetbrains.annotations.NotNull;
+import reactor.core.publisher.Flux;
```
artursouza, line 30: Remove?

artursouza (changes requested):

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
pravinpushkar: @artursouza In reality the useful code for subscribing is only this much. 

```
Flux<SubscribeConfigurationResponse> outFlux = client.subscribeConfiguration(req);
outFlux.subscribe(
        cis -> {
          cis.getItems().forEach((k,v) -> print(v,k));
        }
 ));
``` 
rest other things are are for demonstrating how this is going to work when there is some config update going on behind the scene.
I have simplified the both http and gRPC examples now.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
pravinpushkar: removed this

examples/src/main/java/io/dapr/examples/configuration/http/ConfigSubscriberController.java
pravinpushkar: Actually user's app is supposed to handle the response here for what they want to do with it. We can have the route in spring boot sdk but after that it needs the user's logic to run. Is this possible today with spring boot sdk ?

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
pravinpushkar: removed

pravinpushkar:
@artursouza I will submit a new commit with fixes.

examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java
artursouza: Yes. I think we can make it even more convenient. I will propose something for 1.10

pravinpushkar:
@artursouza Updated the commit with suggested changes. There is minor diff in code cov in PR checks.

sdk/src/main/java/io/dapr/client/domain/ConfigurationItem.java
mukundansundar: Another option is here to use `@JsonCreator` annotation ...

sdk/src/main/java/io/dapr/client/BaseSubscribeConfigHandler.java
```diff
@@ -0,0 +1,86 @@
...
+package io.dapr.client;
+
+import io.dapr.client.domain.SubscribeConfigurationResponse;
+import java.util.logging.Logger;
+
+/**
+ * Base class to handle configuration subscribe response.
+ * Application needs to extend this calls to have better control on response handelling
+ */
+public class BaseSubscribeConfigHandler {
```
artursouza, line 23: Move the logic to DaprRuntime so we avoid spreading the number of singletons in the SDK.

You might just need to add the following to DaprRuntime:
```java
  private final Map<String, BiConsumer<String, String>> configurationChangeHandlers = Collections.synchronizedMap(new HashMap<>());

  public void registerConfigurationChangeHandler(String store, BiConsumer<String, String> handler) {
    this.configurationChangeHandlers.put(store, handler);
  }

  public void handleConfigurationChange(String store, String key, String value) {
    BiConsumer<String, String> handler = this.configurationChangeHandlers.get(store);
    handler.accept(key, value);
  }
```

sdk/src/main/java/io/dapr/client/domain/ConfigurationItem.java
artursouza: I agree with Mukundan. This also should make use of the constructor we have.

artursouza (approved):

artursouza (changes requested):
Forgot to mark as "request changes"

sdk/src/main/java/io/dapr/client/domain/ConfigurationItem.java
pravinpushkar: done

sdk/src/main/java/io/dapr/client/BaseSubscribeConfigHandler.java
pravinpushkar: Did this change. The catch is `DaprRuntime.java` has default modifier so, we can not access the `registerConfigurationChangeHandler ` from examples.

I made `DaprRuntime` to 'public final` hope this should be fine.

mukundansundar (commented):
@artursouza Do you think making `DaprRuntime` public is the right step here?

sdk-springboot/src/main/java/io/dapr/springboot/DaprRuntime.java
```diff
@@ -14,17 +14,19 @@
...
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
 /**
  * Internal Singleton to handle Dapr configuration.
```
mukundansundar, line 27: nitpick not an internal class anymore

sdk/src/main/java/io/dapr/client/BaseSubscribeConfigHandler.java
artursouza: Sorry, this side effect was unexpected and undesired. Let me find a solution for this right now to merge this for 1.9.

artursouza: Let's remove it and keep the SpringBoot example without this layer of convenience. I have explored a few options and I think we will need further design discussions. So, I will make the changes now to simplify it until we create clarity for that.

artursouza (approved):

Issue #759, opened 2022-07-11
Fix the guid in the README  file


RcXu:
## Actual Behavior

As the guid in the README, the version of JDK is 1.5.0, but there is something wrong when use the package io.dapr.Rule;
and it's used in example file also。
when configure the 1.5.0 version, can't find the package but 1.6.0 is ok.

RELEASE NOTE: **FIX** README guid content

mukundansundar:
If the Rule is for pubusb routing, that is only there from 1.6.

mukundansundar:
Can you check master branch now? I think that it was fixed with release of 1.6 and updating README in master branch.

RcXu:
yes, it's already fixed in the master

PR #760, opened 2022-07-11
Fix the guid content in the README file

RcXu:
Signed-off-by: RcXu <rongcai.xu@ivydad.com>

# Description

As the guid in the README, the version of JDK is 1.5.0, but there is something wrong when use the package io.dapr.Rule;
and it's used in example file also。
when configure the 1.5.0 version, can't find the package but 1.6.0 is ok.

Please reference the issue this PR will close: #759


mukundansundar:
As there are zero changes, closing PR.

PR #764, opened 2022-07-25
Support distributed lock API

CrazyHZM:
# Description

## Issue reference

Please reference the issue this PR will close: #747

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


sdk/src/main/java/io/dapr/client/domain/UnLockRequest.java
```diff
@@ -0,0 +1,74 @@
...
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.client.domain;
+
+/**
+ * A request to unlock.
+ */
+public class UnLockRequest {
```
mukundansundar, line 19:
```suggestion
public class UnlockRequest {
```

sdk/src/main/java/io/dapr/client/domain/LockRequest.java
```diff
@@ -0,0 +1,86 @@
...
+limitations under the License.
+*/
+
+package io.dapr.client.domain;
+
+
+/**
+ * A request to lock.
+ */
+public class LockRequest {
```
mukundansundar, line 20: Make this class json serializable with only the required properties ie resourcrId etc being in the json form ie except storeName.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -657,6 +660,110 @@ public Mono<Map<String, Map<String, String>>> getBulkSecret(GetBulkSecretRequest
...
+      }
+      if (resourceId == null || resourceId.isEmpty()) {
+        throw new IllegalArgumentException("ResourceId cannot be null or empty.");
+      }
+      if (lockOwner == null || lockOwner.isEmpty()) {
+        throw new IllegalArgumentException("LockOwner cannot be null or empty.");
+      }
+
+
+      Map<String, Object> jsonMap = new HashMap<>();
```
mukundansundar, line 737: Make the UnlockRequest itself a json serializable object and have only required fields in json form using Jackson annotations?

sdk/src/main/java/io/dapr/client/DaprClient.java
```diff
@@ -559,6 +562,39 @@ Mono<Void> executeStateTransaction(String storeName,
...
 
+  /**
+   * Tries to get a lock with an expiry.
+   * @param storeName Name of the store
+   * @param resourceId Lock key
+   * @param lockOwner The identifier of lock owner
+   * @param expiryInSeconds The time before expiry
+   * @return Whether the lock is successful
+   */
+  Mono<Boolean> tryLock(String storeName, String resourceId, String lockOwner, Integer expiryInSeconds);
```
mukundansundar, line 573: Move all the new method definitions into the Preview Client interface since the distributed lock api is still in alpha. Only stable api methods go in this DaprClient interface....

mukundansundar (changes requested):
@CrazyHZM 
Thanks for this PR!! Added some comments. Please add E2E, example and it's validation for this api with both http and grpc. See the configuration api implementation for the Preview Client change if needed.

sdk/src/main/java/io/dapr/client/domain/LockRequest.java
CrazyHZM: I think this is a class open to users. If `storeName` is excluded, the meaning of LockRequest will be incomplete. It can be compared to `GetBulkStateRequest`, `DeleteStateRequest`, etc. What do you think?

sdk/src/main/java/io/dapr/client/DaprClient.java
CrazyHZM: My mistake, I will adjust this.

sdk/src/main/java/io/dapr/client/domain/LockRequest.java
mukundansundar: I think it is ok to have `storeName` excluded in the JSON format with correct documentation. @artursouza WDYT?

mukundansundar:
@CrazyHZM Please fix conflicts in the PR ...

examples/src/main/java/io/dapr/examples/lock/grpc/DistributedLockGrpcClient.java
```diff
@@ -0,0 +1,75 @@
...
+import io.dapr.client.DaprPreviewClient;
+import io.dapr.client.domain.LockRequest;
+import io.dapr.client.domain.UnlockRequest;
+import io.dapr.client.domain.UnlockResponseStatus;
+import reactor.core.publisher.Mono;
+
+/**
+ * DistributedLockGrpcClient.
+ */
+public class DistributedLockGrpcClient {
```
artursouza, line 27: We don't need an example for both HTTP and gRPC since the SDK abstract that.
It would be great if the example was using the APIs to make something more useful, like leader election, for example.

sdk/src/main/java/io/dapr/client/domain/LockRequest.java
artursouza: I agree. We can annotate to make it not part of the output.

sdk/src/test/java/io/dapr/client/DaprPreviewClientGrpcTest.java
```diff
@@ -63,6 +64,8 @@ public class DaprPreviewClientGrpcTest {
 	private static final ObjectMapper MAPPER = new ObjectMapper();
 	private static final String CONFIG_STORE_NAME = "MyConfigStore";
 	private static final String QUERY_STORE_NAME = "testQueryStore";
+    private static final String LOCK_STORE_NAME = "MyLockStore";
```
artursouza, line 67: nit: format

artursouza (changes requested):
Pretty close to be merged. Thanks.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -662,6 +665,110 @@ public Mono<Map<String, Map<String, String>>> getBulkSecret(GetBulkSecretRequest
...
+      }
+      if (expiryInSeconds < 0) {
+        throw new IllegalArgumentException("ExpiryInSeconds cannot be negative.");
+      }
+
+
+      Map<String, Object> jsonMap = new HashMap<>();
+      jsonMap.put("resourceId", resourceId);
+      jsonMap.put("lockOwner", lockOwner);
+      jsonMap.put("expiryInSeconds", expiryInSeconds);
```
johnewart, line 696: Would you mind making these keys constants? Magic strings / numbers can cause bugs due to typos. I think @mukundansundar's suggestion below would also help with this.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
johnewart: Also magic strings (see above request)

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -662,6 +665,110 @@ public Mono<Map<String, Map<String, String>>> getBulkSecret(GetBulkSecretRequest
...
+              context -> this.client
+                      .invokeApi(DaprHttp.HttpMethods.POST.name(), pathSegments, null, requestBody, null, context)
+      ).flatMap(response -> {
+        try {
+          Map m = INTERNAL_SERIALIZER.deserialize(response.getBody(), Map.class);
+          if (m == null) {
+            return Mono.just(UnlockResponseStatus.INTERNAL_ERROR);
+          }
+
+          Integer statusCode = (Integer) m.get("status");
```
johnewart, line 760: Same here - can we make `"status"` a constant so it's guaranteed to be consistent anywhere it gets used?

johnewart (commented):
Just small requests re: magic strings but not a blocker.

sdk/src/main/java/io/dapr/client/domain/LockRequest.java
CrazyHZM: Done.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
CrazyHZM: For the time being, only the lock api is in use, and if there is a scene in the future, it will be extracted

examples/src/main/java/io/dapr/examples/lock/grpc/DistributedLockGrpcClient.java
CrazyHZM: It has not been deleted for the time being, and new examples can be added in the future.

msfussell:
@mukundansundar @artursouza - Is this ready for review?

sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
```diff
@@ -2392,6 +2375,7 @@ public void shutdownTest() {
     result.block();
   }
 
+
   private <T> DaprProtos.GetStateResponse buildFutureGetStateEnvelop(T value, String etag) throws IOException {
     return buildGetStateResponse(value, etag);
   }
```
artursouza, line 2381: Need to bring back the code from previous commit.

artursouza: I confirmed that nothing should change (apart from import organizer) since the new API is in the preview client.

artursouza (approved):

PR #789, opened 2022-09-26
initial bulk publish impl for java

### LINKED ISSUE #778: Bulk PubSub implementation
## Describe the proposal

Add SDK support for the new bulk Pub/Sub API runtime issue: https://github.com/dapr/dapr/issues/2218

mukundansundar:
Signed-off-by: Mukundan Sundararajan <65565396+mukundansundar@users.noreply.github.com>

# Description

initial bulk publish implementation for Java SDK

## Issue reference

Since bulk subscribe is already implemented, this PR closes #778 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -182,6 +189,82 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
...
+      envelopeBuilder.setTopic(topic);
+      envelopeBuilder.setPubsubName(pubsubName);
+
+      for (BulkPublishRequestEntry<?> entry: request.getEntries()) {
+        Object event = entry.getEvent();
+        byte[] data;
+        String contentType = entry.getContentType();
+
+        // Serialize event into bytes
+        if (!Strings.isNullOrEmpty(contentType) && objectSerializer instanceof DefaultObjectSerializer) {
```
artursouza, line 210: This is a problem we need to fix in other places in the SDK. Using the default serializer should not trigger different behavior. It should be based on the content-type only.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -174,6 +183,85 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
...
+
+      List<BulkPublishRequestEntry<?>> entries = new ArrayList<>();
+
+      for (BulkPublishRequestEntry<?> entry: request.getEntries()) {
+        Object event = entry.getEvent();
+        byte[] data;
+        String contentType = entry.getContentType();
+
+        // Serialize event into bytes
+        if (!Strings.isNullOrEmpty(contentType) && objectSerializer instanceof DefaultObjectSerializer) {
```
artursouza, line 211: No different behavior because of serializer's class, just content-type.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -174,6 +183,85 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
...
+          data = objectSerializer.serialize(event);
+          if (Strings.isNullOrEmpty(contentType)) {
+            // Only override content type if not given in input by user
+            contentType = objectSerializer.getContentType();
+          }
+        } else {
+          // this is scenario where content-type is empty
+          throw new IllegalArgumentException("content type expected for data");
+        }
+        BulkPublishRequestEntry<?> bulkPublishRequestEntry = new BulkPublishRequestEntry<>(entry.getEntryID(),data,
```
artursouza, line 233: Maybe `data` should be an object instead? This way, if the content-type for the serializer is JSON, you can pass it through here? You might be able to pass as a JSONObject directly?

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -174,6 +183,85 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
...
+
+      Map<String, List<String>> queryArgs = metadataToQueryArgs(request.getMetadata());
+
+      return Mono.subscriberContext().flatMap(
+          context -> this.client.invokeApi(
+              DaprHttp.HttpMethods.POST.name(), pathSegments, queryArgs, serializedRequest, requestHeaders, context
+          )
+      ).map(
+          it -> {
+            BulkPublishResponse response;
```
johnewart, line 251: This should be able to be simplified to `try { return INTERNAL_SERIALIZER... } catch (..) { }`, right?

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
```diff
@@ -224,4 +227,45 @@ <T> Mono<QueryStateResponse<T>> queryState(String storeName, Query query,
...
+   * @param pubsubName the pubsub name we will publish the event to.
+   * @param topicName the topicName where the event will be published.
+   * @param events the {@link List} of events to be published.
+   * @param contentType the content type of the event. Use Mime based types.
+   * @return the {@link BulkPublishResponse} containing publish status of each event.
+   *     The "entryID" field in {@link BulkPublishResponseEntry} in {@link BulkPublishResponse} will be
+   *     generated based on the order of events in the {@link List}.
+   * @param <T> The type of the events to publish in the call.
+   */
+  <T> Mono<BulkPublishResponse> publishEvents(String pubsubName, String topicName, List<T> events, String contentType);
```
johnewart, line 252: There may be a good reason why we do not already do this, but would it be possible to use `Optional` parameters instead of overriding methods and passing `null` values around?

sdk/src/main/java/io/dapr/client/domain/BulkPublishRequest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * Constructor for BulkPublishRequest.
+   * @param pubsubName Name of the pubsub to publish to.
+   * @param topic      Name of the topic to publish to.
+   */
+  public BulkPublishRequest(String pubsubName, String topic) {
+    this.pubsubName = pubsubName;
+    this.topic = topic;
+    this.entries = new ArrayList<>();
```
johnewart, line 57: See previous question about optional values, but also why is this mutable when the list in the constructor below is wrapped as immutable?

sdk/src/main/java/io/dapr/client/domain/BulkPublishRequest.java
```diff
@@ -0,0 +1,102 @@
...
+  public String getTopic() {
+    return topic;
+  }
+
+  public Map<String, String> getMetadata() {
+    return metadata;
+  }
+
+  public BulkPublishRequest<T> setMetadata(Map<String, String> metadata) {
+    this.metadata = metadata == null ? null : Collections.unmodifiableMap(metadata);
```
johnewart, line 90: Same thing about optional values but if we can't use them, can we use an empty map instead of null in places like this?

johnewart (commented):
Just a few questions / observations, mostly for my own edification - you don't need to block on them

sdk/src/main/java/io/dapr/client/domain/BulkPublishRequest.java
mukundansundar: fixed

mukundansundar:
The build failures here are wrt Config API changes which is handled in the PR #740.

Locally also verified that the PubsubIT passes as expected .... 
cc @dapr/maintainers-java-sdk

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
mukundansundar: moved to using non null but default values ....

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
mukundansundar: this is a change that needs to be done SDK wide and not only for one api ... That might cause serialized events from this API to become incompatible with the other normal publish API ...

sdk/src/main/java/io/dapr/client/domain/BulkPublishResponse.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Class representing the response returned on bulk publishing events.
+ */
+public class BulkPublishResponse {
+
+  private String errorCode;
```
yash-nisar, line 25: Do you think we should remove this field since we have changed the response here: https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L336 ?

yash-nisar: Morever, I don't think we use that field at all: https://github.com/dapr/java-sdk/blob/48b3a6c8d6dad74bdd90edc25f7d5e168a5c1538/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java#L258

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -185,6 +192,78 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
...
+
+      return Mono.subscriberContext().flatMap(
+          context ->
+              this.<DaprProtos.BulkPublishResponse>createMono(
+                  it -> intercept(context, asyncStub).bulkPublishEventAlpha1(envelopeBuilder.build(), it)
+              )
+      ).map(
+          it -> {
+            BulkPublishResponse response = new BulkPublishResponse();
+            List<BulkPublishResponseFailedEntry> entries = new ArrayList<>();
```
yash-nisar, line 251: nit: Shouldn't we name it as `failedEntries` instead of `entries` since it explicitly contains only the failed entries ?

examples/src/main/java/io/dapr/examples/pubsub/bulk/BulkPublisher.java
```diff
@@ -0,0 +1,100 @@
...
+          System.out.println("Going to publish message : " + message);
+        }
+        BulkPublishResponse res = client.publishEvents(PUBSUB_NAME, TOPIC_NAME, messages, "text/plain")
+            .subscriberContext(getReactorContext()).block();
+        System.out.println("Published the set of messages in a single call to Dapr");
+        if (res != null) {
+          if (res.getFailedEntries().size() > 0) {
+            // Ideally this condition will not happen in examples
+            System.out.println("Some events failed to be published");
+            for (BulkPublishResponseFailedEntry entry : res.getFailedEntries()) {
```
yash-nisar, line 80: nit: Shouldn't we name it as failedEntries instead of entries since it explicitly contains only the failed entries ?

sdk-tests/src/test/java/io/dapr/it/methodinvoke/grpc/MethodInvokeIT.java
```diff
@@ -138,10 +139,12 @@ public void testInvokeException() throws Exception {
...
                 client.invokeMethod(daprRun.getAppName(), "sleep", req.toByteArray(), HttpExtension.POST).block());
 
-            assertEquals("UNKNOWN", exception.getErrorCode());
             if (this.useGrpc) {
-                assertEquals("UNKNOWN: ", exception.getMessage());
+                assertEquals("INTERNAL", exception.getErrorCode());
+                assertNotNull(exception.getMessage());
+                assertTrue(exception.getMessage().contains("INTERNAL:"));
             } else {
+                assertEquals("UNKNOWN", exception.getErrorCode());
```
mukundansundar, line 147: related to https://github.com/dapr/dapr/pull/5648/files#diff-ce3f68305f0eb28b6b8911c281186536c80d8cac9dff3cece2c8bdf3dd955315R465-R475
now we get in gRPC (`io.dapr.exceptions.DaprException: INTERNAL: fail to invoke, id: methodinvokeit-methodinvokeservice, err: message is nil`)

mukundansundar:
As discussed offline, this PR only contains the gRPC implementation for Bulk Publish

examples/components/pubsub/kafka.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: kafka-pubsub
+spec:
+  type: pubsub.kafka
+  version: v1
+  metadata:
+    - name: brokers # Required. Kafka broker connection setting
+      value: "localhost:9092"
+    - name: "authType"
+      value: "none"
```
artursouza, line 12: One thing I would like to improve in these ITs, is to scope components only to applicable apps. Can you add scope here, so this component is not used by other tests?

examples/src/main/java/io/dapr/examples/pubsub/bulk/KafkaSubscriber.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2021 The Dapr Authors
```
artursouza, line 2: nit: 2023 everywhere.

examples/src/main/java/io/dapr/examples/pubsub/bulk/README.md
```diff
@@ -0,0 +1,314 @@
...
+== APP == Subscriber got: This is message #2
+== APP == Subscriber got: {"id":"2976f254-7859-449e-b66c-57fab4a72aef","source":"bulk-publisher","type":"com.dapr.event.sent","specversion":"1.0","datacontenttype":"text/plain","data":"This is message #2","data_base64":null}
+== APP == Subscriber got: This is message #3
+== APP == Subscriber got: {"id":"f21ff2b5-4842-481d-9a96-e4c299d1c463","source":"bulk-publisher","type":"com.dapr.event.sent","specversion":"1.0","datacontenttype":"text/plain","data":"This is message #3","data_base64":null}
+== APP == Subscriber got: This is message #4
+== APP == Subscriber got: {"id":"4bf50438-e576-4f5f-bb40-bd31c716ad02","source":"bulk-publisher","type":"com.dapr.event.sent","specversion":"1.0","datacontenttype":"text/plain","data":"This is message #4","data_base64":null}
+== APP == Subscriber got: This is message #5
+== APP == Subscriber got: {"id":"07bad175-4be4-4beb-a983-4def2eba5768","source":"bulk-publisher","type":"com.dapr.event.sent","specversion":"1.0","datacontenttype":"text/plain","data":"This is message #5","data_base64":null}
```
artursouza, line 279: You cannot use UUID since it varies on every run. Unless you use a generator that is deterministic.

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
```diff
@@ -392,6 +396,34 @@ public <T> Mono<QueryStateResponse<T>> queryState(QueryStateRequest request, Cla
     return this.queryState(request, TypeRef.get(clazz));
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public <T> Mono<BulkPublishResponse<T>> publishEvents(String pubsubName, String topicName, List<T> events,
```
artursouza, line 403: nit: instead, can you move content type prior to events? It allows a overloaded method using `T... events` as an array.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -179,6 +186,14 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
     }
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public <T> Mono<BulkPublishResponse<T>> publishEvents(BulkPublishRequest<T> request) {
+    return DaprException.wrapMono(new UnsupportedOperationException());
```
artursouza, line 194: I assume this will be implemented next, right?

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
```diff
@@ -244,4 +248,46 @@ <T> Mono<QueryStateResponse<T>> queryState(String storeName, Query query,
...
+   * @param pubsubName the pubsub name we will publish the event to.
+   * @param topicName the topicName where the event will be published.
+   * @param events the {@link List} of events to be published.
+   * @param contentType the content type of the event. Use Mime based types.
+   * @return the {@link BulkPublishResponse} containing publish status of each event.
+   *     The "entryID" field in {@link BulkPublishEntry} in {@link BulkPublishResponseFailedEntry} will be
+   *     generated based on the order of events in the {@link List}.
+   * @param <T> The type of the events to publish in the call.
+   */
+  <T> Mono<BulkPublishResponse<T>> publishEvents(String pubsubName, String topicName, List<T> events,
```
artursouza, line 273: Move events as the last one and add an overloaded method for `T... events`.

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
```diff
@@ -392,6 +396,34 @@ public <T> Mono<QueryStateResponse<T>> queryState(QueryStateRequest request, Cla
...
+                                                     String contentType, Map<String, String> requestMetadata) {
+    BulkPublishRequest<T> request = new BulkPublishRequest<>(pubsubName, topicName);
+    if (events == null || events.size() == 0) {
+      throw new IllegalArgumentException("list of events cannot be null or empty");
+    }
+    List<BulkPublishEntry<T>> entries = new ArrayList<>();
+    for (int i = 0; i < events.size(); i++) {
+      // entryID field is generated based on order of events in the request
+      entries.add(new BulkPublishEntry<>("" + (i + 1), events.get(i), contentType, null));
```
artursouza, line 421: I think starting from zero is more idiomatic to Java developers.

sdk/src/test/java/io/dapr/utils/TypeRefTest.java
```diff
@@ -0,0 +1,21 @@
...
+  @Test
+  public void testTypeRefIsPrimitive() {
+    Assert.assertTrue("expected this to be true as boolean is primitive", TypeRef.isPrimitive(TypeRef.BOOLEAN));
+    Assert.assertTrue("expected this to be true as short is primitive", TypeRef.isPrimitive(TypeRef.SHORT));
+    Assert.assertTrue("expected this to be true as float is primitive", TypeRef.isPrimitive(TypeRef.FLOAT));
+    Assert.assertTrue("expected this to be true as double is primitive", TypeRef.isPrimitive(TypeRef.DOUBLE));
+    Assert.assertTrue("expected this to be true as integer is primitive", TypeRef.isPrimitive(TypeRef.INT));
+
+    Assert.assertFalse("expected this to be false as string is not primitive",
+        TypeRef.isPrimitive(TypeRef.STRING));
```
artursouza, line 17: nit: alignment

artursouza (changes requested):

examples/components/pubsub/kafka.yaml
mukundansundar: done

examples/src/main/java/io/dapr/examples/pubsub/bulk/KafkaSubscriber.java
mukundansundar: done

examples/src/main/java/io/dapr/examples/pubsub/bulk/README.md
mukundansundar: Here I am not validating that it got a particular message, I am only validating that it got the message i.e. `Subscriber got: This is message #5`. I am not validating the ID.

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
mukundansundar: done, changed the other func with reqMeta also.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
mukundansundar: @artursouza as we discussed previously, we were saying that HTTP impl is not needed, and that HTTP Client is going to be deprecated. https://github.com/dapr/java-sdk/pull/790

sdk/src/main/java/io/dapr/client/DaprPreviewClient.java
mukundansundar: done

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
mukundansundar: changed

sdk/src/test/java/io/dapr/utils/TypeRefTest.java
mukundansundar: alignment is following the previous line, checkstyle is fine here.

mukundansundar:
@dapr/maintainers-java-sdk see the latest checks for this PR itself. 

For some reason the Kafka subscription is flaky. 

Additionally I am seeing flakiness in other validations as well. Please see the logs

mukundansundar:
@artursouza any updates on this PR?

examples/src/main/java/io/dapr/examples/pubsub/bulk/README.md
artursouza: OK.

examples/components/pubsub/kafka.yaml
artursouza: Oh. I mean to use the actual scope feature - not move to a different folder.

sdk/src/main/java/io/dapr/client/domain/BulkPublishResponse.java
```diff
@@ -0,0 +1,64 @@
...
+    this.errorCode = errorCode;
+    this.failedEntries = failedEntries == null ? Collections.unmodifiableList(new ArrayList<>()) :
+        Collections.unmodifiableList(failedEntries);
+  }
+
+  public String getErrorCode() {
+    return errorCode;
+  }
+
+  public BulkPublishResponse<T> setErrorCode(String errorCode) {
```
artursouza, line 51: Can this class be immutable?

sdk/src/main/java/io/dapr/client/domain/BulkPublishRequest.java
```diff
@@ -0,0 +1,104 @@
...
+    return this;
+  }
+
+  public List<BulkPublishEntry<T>> getEntries() {
+    return entries;
+  }
+
+  public BulkPublishRequest<T> setEntries(List<BulkPublishEntry<T>> entries) {
```
artursouza, line 100: Immutable class.

artursouza (changes requested):

examples/components/pubsub/kafka.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: kafka-pubsub
```
artursouza, line 4: Why not use Redis for this too. The README file will not work without additional instructions to setup Kafka.

artursouza (changes requested):

examples/components/pubsub/kafka.yaml
mukundansundar: done.

sdk/src/main/java/io/dapr/client/domain/BulkPublishResponse.java
mukundansundar: done.

sdk/src/main/java/io/dapr/client/domain/BulkPublishRequest.java
mukundansundar: done.

mukundansundar:
> @dapr/maintainers-java-sdk see the latest checks for this PR itself. It required only 2 runs. The older one https://github.com/dapr/java-sdk/actions/runs/3871888965 required 4 attempts. (Before adding doc content)
> 
> For some reason the Kafka subscription is flaky.
> 
> Additionally I am seeing flakiness in other validations as well. Please see the logs

Changed the examples to use Redis, that removes the flakiness.

mukundansundar:
@dapr/maintainers-java-sdk Had a discussion with @shubham1172 on whether to merge BulkSubscribeMessageEntry and BulkPublishEntry classes into a single class or not. As of now what we feel is that it is clearer to have two different set of classes for BulkPublish and BulkSubscribe and not have a common class in-between. There is some duplication, but since the APIs are orthogonal they can be used independently of each other and it will be clearer if we have two sets of classes and not have a common `BulkMessageEntry` class which will be used by BulkPublish and BulkSubscribe APIs.

artursouza (approved):

Issue #763, opened 2022-07-18
Replace okhttp with azure  core http


tanvigour:
Replace this to fix the dependency issue with okhttp3 https://github.com/dapr/java-sdk/issues/687

mukundansundar:
@tanvigour why are we replacing one dependency with another?

artursouza:
The Azure dependency is the same as the OKHTTP but in a different namespace to avoid conflict. There is a well known issue of conflicting OKHTTP version with the app when importing the Dapr Java SDK. It is a major pain point to get started. People have to pick one version to resolve conflict and hope they are compatible. This will stop this conflict.

We also had plans to remove OKHTTP completely and use Java's native HTTP client but that requires Java 11 as minimum support. Right now, we need to keep Java 8 support because of Spring Cloud requirements.

/cc @skyao

tanvigour:
> @tanvigour why are we replacing one dependency with another?

To fix the version compatibility issue as mentioned in this : https://github.com/dapr/java-sdk/issues/687

mukundansundar:
@artursouza What is the Spring Cloud dependency that you mention? 
cc @skyao

skyao:
> @artursouza What is the Spring Cloud dependency that you mention? cc @skyao

To integrate dapr with spring cloud, since current main version of spring is supporting jdk 1.8 like: 

- spring cloud v3.x
- spring boot v2.x
- spring framework v5.x 

And for next main version of spring/springboot is required jdk17. 

I think most spring users will stay in current spring for a long time because jdk17 and wait for spring new version to be stable. 

So, upgrading  jdk version of dapr java sdk to jdk11 will bring a risk that if the users are using jdk1.8, then we will loss these users ---- I don't think the user will update to jdk11 just because of dapr.

The best strategy of dapr java sdk is that we can support all of them and let the developer choose one of them:

- for jdk1.8, use okhttp3 
- for jdk11, use httpclient in jdk
- customized by the developers if they use other http client in their applications and don't want to introduce a new one.

johnewart:
@artursouza and I were talking about this earlier and I opened a PR to deprecate the HTTP-transport bits; we could get rid of some code it seems that it could make transitioning easier. 

Additionally, we could consider splitting off the HTTP client into its own package/artifact that could be used to keep HTTP support until a project has made the transition away from HTTP (it would also be easier to guarantee you have transitioned your project because removing the `-http` artifact as a dependency would have no effect)

PR #790, opened 2022-09-28
Mark HTTP support as deprecated to be removed in a future release.

johnewart:
# Description

In conversation with @artursouza earlier today regarding issue #763 we came to the conclusion that it might just make sense to simply deprecate the HTTP transport and let people implement their own HTTP client in the cases where they don't want to use gRPC. This would eliminate the split-dependency issue and (I believe) also make the JVM transition easier. 

This PR marks the HTTP-related client constructors, constants and references as deprecated (at least I think I got all the places we instantiate HTTP transport related bits) 

## Issue reference

This is related to (and should be able to close) issue #763 -- also related to proposal #794

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [NA] Extended the documentation (Javadoc generation should be automatic)


sdk/src/main/java/io/dapr/config/Properties.java
```diff
@@ -45,6 +45,8 @@ public class Properties {
 
   /**
    * Dapr's default use of gRPC or HTTP for Dapr's method invocation APIs.
+   *
+   * @deprecated HTTP support is being deprecated; use gRPC instead. This will be removed in a future release.
```
mukundansundar, line 49: @johnewart had a discussion with @artursouza .... Let's not have this deprecated as yet. Only for service invocation... We need to decide on the route of that

sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
```diff
@@ -172,6 +175,8 @@ private DaprClient buildDaprClientGrpc() {
    * Creates and instance of DaprClient over HTTP.
    *
    * @return DaprClient over HTTP.
+   *
+   * @deprecated HTTP support is being deprecated; Use the gRPC builder method (buildDaprClientGrpc) instead. This will be removed in a future release.
    */
   private DaprClient buildDaprClientHttp() {
     return new DaprClientHttp(this.daprHttpBuilder.build(), this.objectSerializer, this.stateSerializer);
```
mukundansundar, line 182: Log a warning that http client is deprecated and will be removed in future release...

mukundansundar (changes requested):
Small changes. 
Can you also create an issue to track the http client deprecation...

This requires a docs change as well

mukundansundar:
Please also see the checkstyle errors

sdk/src/main/java/io/dapr/config/Properties.java
artursouza: Yes, we still need this, unless we provide a new API that is called "invokeHTTP" or something. Then this one can be deprecated.

artursouza (changes requested):
In addition, we need to log warning if the HTTP protocol is chosen via the property we provide in the SDK, this way the user will see the warning in their logs. Users will not see a warning in code due to these deprecation annotations because this code is not used directly by the user, we have a builder that handles which instance of the client to build.

sdk/src/main/java/io/dapr/config/Properties.java
johnewart: Sure, but (for my understanding) is it common for one to use gRPC to talk to Dapr but not want to use it for method invocation?

sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
johnewart: I had thought about that but we don't have a standard logging mechanism in the client (I looked and it appears we don't log anything except in the tests). I would hesitate to take additional dependencies unless we are going to add logging elsewhere in the client (which might not be a bad idea but we don't currently). 

Thoughts?

johnewart:
> In addition, we need to log warning if the HTTP protocol is chosen via the property we provide in the SDK, this way the user will see the warning in their logs. Users will not see a warning in code due to these deprecation annotations because this code is not used directly by the user, we have a builder that handles which instance of the client to build.

I agree but see above, we don't have a logging façade dependency / logging in place in the client; do we want to add a new dependency like slf4j for that?

johnewart:
Fixed checkstyle errors, removed one of the depreciation tags as requested and also added signoff -- amended and force pushed to keep things simpler.

johnewart:
Created proposal #794 to track this

sdk/src/main/java/io/dapr/config/Properties.java
artursouza: When the service is gRPC, yes. When the service is HTTP, we want clients to use HTTP invocation and we have received feedback that users like to have a convenient method for invoking HTTP services. So, we need to keep an HTTP specific method in the client.

sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
artursouza: That is a great point. We don't have any logging in the SDK on purpose right now to avoid bringing dependencies that might conflict with apps' dependencies. I see 2 options here:

1. Depend on SLF4J, so users can still pick their own logging solution as SLF4J is an abstraction. Of course, the version of SLF4J can conflict and users will need to pick one (probably the latest).
2. Use System.out.Println() - I am OK since most users run this on containers or local mode where logs go to STDOUT or STDERR.

artursouza:
I responded. I see 2 options:

SLF4J or System.out.Println() because this is just for giving warnings for deprecation and not for other use cases (yet).

sdk/src/main/java/io/dapr/client/DaprClientBuilder.java
johnewart: I'm fine with just printing to STDOUT for the moment -- while most of the universe uses SLF4J, and it's pretty stable, there are enough issues with dependencies in general that it would be great if we could avoid piling on any more.

johnewart: Added a `System.out.println` statement when constructing `DaprClientHttp` objects.

johnewart: Added a `System.out.println` statement when constructing a `DaprClientHttp` object

sdk/src/main/java/io/dapr/config/Properties.java
johnewart: Removed the deprecation notice from this portion of the code.

mukundansundar (approved):
lgtm

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -114,6 +117,7 @@ public class DaprClientHttp extends AbstractDaprClient {
     this.client = client;
     this.isObjectSerializerDefault = objectSerializer.getClass() == DefaultObjectSerializer.class;
     this.isStateSerializerDefault = stateSerializer.getClass() == DefaultObjectSerializer.class;
+    System.out.println("NOTE: DaprClientHttp is deprecated and will be removed in a future release.");
```
tanvigour, line 120: NIT, but it might be more noticeable if we change "NOTE" to "deprecation ALERT" or something to the effect that holds attention as something that shouldn't be ignored as a warning

tanvigour (approved):
LGTM otherwise

mukundansundar:
@artursouza Can this be merged?

artursouza:
Closing this in favor of https://github.com/dapr/java-sdk/pull/824

Issue #792, opened 2022-09-28
Building the SDK will use cached protobuf definitions (if present) even in a clean checkout / build


johnewart:
## Expected Behavior

`git clone git@github.com:dapr/java-sdk.git && cd java-sdk && mvn package` should work out of the box, but it doesn't if you 
have previously build the Java SDK with a different set of protobuf definitions. They instead get cached in `$HOME/.m2/repository/.cache/...` and the failure is non-obvious until you hunt around for docs on Maven's `wget` plugin. 

## Actual Behavior

Build fails (iff the new code depends on protobuf definitions different than those you have in your cache).

## Steps to Reproduce the Problem

Have a set of protobuf files cached from a previous Java SDK build that differ from the ones required, then try building a clean clone -- it will fail. 

## Release Note

RELEASE NOTE:

**FIX** Skip caching protobuf definition files to make `mvn clean` behave more as one would expect.

PR #793, opened 2022-09-28
Update docs regarding changing the protobuf url and skip the cache

johnewart:
Signed-off-by: John Ewart <john@johnewart.net>

# Description

Changed the docs (they had incorrect property names regarding protobuf files) and added `skipCache=true` to the `wget` tasks because the files are small and it causes unexpected behavior when changing the protobuf URL (in that it keeps using a cached version even after using `mvn clean` and `git clean` which makes life a bit confusing until you realize they are being cached in `$HOME/.m2/repository/.cache`) so that it will always re-download protobuf definitions. 

Also moved the base URL to point to master (@artursouza requested it but it seems like a good idea to keep the SDK up-to-date with whatever the latest protobuf definitions are). 

Note that this will not re-download the protobuf files on every build, only when they are not already present (i.e if you do a `mvn clean` or `git clean` to remove `target` directories and other auto-generated bits.)

## Issue reference

This fixes issue #792

## Checklist

* [x] Code compiles correctly
* [NA] Created/updated tests


pom.xml
```diff
@@ -16,7 +16,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <grpc.version>1.42.1</grpc.version>
     <protobuf.version>3.17.3</protobuf.version>
-    <dapr.proto.baseurl>https://raw.githubusercontent.com/dapr/dapr/683a436ff1367e4ff2d27da2d79069c04ec2c46d/dapr/proto</dapr.proto.baseurl>
+    <dapr.proto.baseurl>https://raw.githubusercontent.com/dapr/dapr/master/dapr/proto</dapr.proto.baseurl>
```
mukundansundar, line 19: should we always only point to master? In releases this is always set to a particular commit hash for reproducibility of build.... can we continue that for normal master builds as well ?

artursouza: Same here. We don't point to master because we want the build to be predictable.

artursouza (changes requested):
We don't point to master on purpose. See comment.

pom.xml
johnewart: Updated back to ref `68a43...` in the pom for now; regarding reproducibility, should we consider doing something similar to the k8s Java client (and others) and simply add the generated source files to git? It would avoid downloading temporary protobuf files and also make it clear which source files are being used / generated for a build if there is a problem.

sdk-autogen/pom.xml
```diff
@@ -63,43 +63,40 @@
         <executions>
           <execution>
             <id>getCommonProto</id>
-            <!-- the wget goal actually binds itself to this phase by default -->
```
mukundansundar, line 66: why are the comments removed ?

johnewart: Because, at least according to the docs page where I found the cache directory / config settings for the wget plugin, this is an inaccurate statement. See http://maven-download-plugin.github.io/maven-download-plugin/docsite/1.6.0/wget-mojo.html. The docs say "Binds by default to the lifecycle phase: process-resources."

artursouza (approved):

PR #814, opened 2022-11-25
Fix App-IDs across examples and tests

shubham1172:
Signed-off-by: Shubham Sharma <shubhash@microsoft.com>

# Description

Fix breaking changes due to https://github.com/dapr/dapr/pull/5556

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/DaprRun.java
+++ b/sdk-tests/src/test/java/io/dapr/it/DaprRun.java
@@ -58,7 +58,8 @@ public class DaprRun implements Stoppable {
                   DaprApiProtocol protocol,
                   DaprApiProtocol appProtocol) {
     // The app name needs to be deterministic since we depend on it to kill previous runs.
-    this.appName = serviceClass == null ? testName : String.format("%s-%s", testName, serviceClass.getSimpleName());
+    this.appName = 
+        serviceClass == null ? testName : String.format("%s-%s", testName, serviceClass.getSimpleName()).toLowerCase();
     this.startCommand =
         new Command(successMessage, buildDaprCommand(this.appName, serviceClass, ports, protocol, appProtocol));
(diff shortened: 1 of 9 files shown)


sdk-tests/src/test/java/io/dapr/it/DaprRun.java
```diff
@@ -58,7 +58,7 @@ private DaprRun(String testName,
                   DaprApiProtocol protocol,
                   DaprApiProtocol appProtocol) {
     // The app name needs to be deterministic since we depend on it to kill previous runs.
-    this.appName = serviceClass == null ? testName : String.format("%s_%s", testName, serviceClass.getSimpleName());
+    this.appName = serviceClass == null ? testName : String.format("%s-%s", testName, serviceClass.getSimpleName());
```
pravinpushkar, line 61: @shubham1172 Is this change related to the change done in runtime ? That change only talk about `.` or empty string if I remember correctly.

pravinpushkar: Ya, I think it does prohibit "_" also.

pravinpushkar (dismissed):

sdk-tests/src/test/java/io/dapr/it/DaprRun.java
shubham1172: It does talk about `.` but it does prohibit `_` and upper case letters too.

mukundansundar (approved):
LGTM.
Irrespective of https://github.com/dapr/dapr/pull/5556, I think this change is correct. 
cc @artursouza thoughts?

pravinpushkar (approved):

shubham1172:
@mukundansundar can we merge this?

Issue #783, opened 2022-09-19
Unable to get the http response details by Dapr HttpBinding if the state code is not 200


ParrySMS:
## Expected Behavior

I have service A, and service A will call service B by the DaprClient. All thing works when service B returns the response with state code 200.

```java
    @Autowired
    private DaprClient client;
   
    @Override
    public Mono<BBBResponse> getUserData(String realms, String accessToken) {
        var metadata = setUpHeaderParams(accessToken);
        var queryParams = setUpQueryParams("m_getUserData");
        var path = toQueryParamsPath(String.format(AUTHENTICATE_PATH, realms), queryParams);
        metadata.put("path", path);
        return client.invokeBinding(properties.getHttpBindingName(), DaprBindingOperation.POST.getValue(), null, metadata, BBBResponse.class);
    }
```

If the state code is not 200, I want to get the original response details. The service B will return a response with a 401 state code like this one:

```json
{
    "code": "BS_0002"
    "message": "The access token is expired."
}
```

## Actual Behavior

I will get an error like this one if I get the 401 from Service B.

"error when invoke output binding aaa-service-http-binding: received status code 401."

There should be a way to catch this HTTP binding exception or somehow get the original response. I need to get the original response to do something.

## Steps to Reproduce the Problem

Use a DaprClient to call other APIs by function `invokeBinding()`.

Get a response but the state code is not 200.

No way to get the orginal response but a dapr error like "error when invoke output binding XXXX : received status code YYY."

## Release Note

RELEASE NOTE:

ParrySMS:
The root cause here but I am not sure how to fix it.

You can check this function `doInvokeApi` in the `DaprHttp.java`.

```
  /**
   * Invokes an API that returns a text payload.
   *
   * @param method        HTTP method.
   * @param pathSegments  Array of path segments (/a/b/c -> ["a", "b", "c"]).
   * @param urlParameters Parameters in the URL
   * @param content       payload to be posted.
   * @param headers       HTTP headers.
   * @param context       OpenTelemetry's Context.
   * @return CompletableFuture for Response.
   */
  private CompletableFuture<Response> doInvokeApi(String method,
                               String[] pathSegments,
                               Map<String, List<String>> urlParameters,
                               byte[] content, Map<String, String> headers,
                               Context context)
```

In the line 320, it enqueue a ResponseFutureCallback. It converts the okhttp3 response into the response object expected internally by the SDK.

[...]

PR #842, opened 2023-03-29
Custom non-2xx response handling - DaprErrorResponseParser - Issue 783

maciek047:
# Description

This PR aims to address the problems described in issue #783. It introduces the DaprErrorResponseParser interface, which allows for setting a custom parser for the DaprHttpClient. This enables the deserialization of responses with structures different from DaprError. The changes are backward compatible, as the current logic is wrapped in the DefaultDaprErrorResponseParser.

## Issue reference
https://github.com/dapr/java-sdk/issues/783

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
@@ -68,7 +68,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.FluxSink;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.MonoSink;
-import reactor.util.context.Context;
+import reactor.util.context.ContextView;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -188,7 +188,7 @@ public class DaprClientGrpc extends AbstractDaprClient {
         envelopeBuilder.putAllMetadata(metadata);
       }
 
-      return Mono.subscriberContext().flatMap(
+      return Mono.deferContextual(
           context ->
               this.<Empty>createMono(
                   it -> intercept(context, asyncStub).publishEvent(envelopeBuilder.build(), it)
@@ -261,7 +261,7 @@ public class DaprClientGrpc extends AbstractDaprClient {
       for (BulkPublishEntry<T> entry: request.getEntries()) {
         entryMap.put(entry.getEntryId(), entry);
       }
-      return Mono.subscriberContext().flatMap(
+      return Mono.deferContextual(
           context ->
               this.<DaprProtos.BulkPublishResponse>createMono(
                   it -> intercept(context, asyncStub).bulkPublishEventAlpha1(envelopeBuilder.build(), it)
@@ -297,15 +297,19 @@ public class DaprClientGrpc extends AbstractDaprClient {
       Object body = invokeMethodRequest.getBody();
       HttpExtension httpExtension = invokeMethodRequest.getHttpExtension();
       DaprProtos.InvokeServiceRequest envelope = buildInvokeServiceRequest(
-              httpExtension,
-              appId,
-              method,
-              body);
+          httpExtension,
+          appId,
+          method,
+          body);
+      // Regarding missing metadata in method invocation for gRPC:
+      // gRPC to gRPC does not handle metadata in Dapr runtime proto.
+      // gRPC to HTTP does not map correctly in Dapr runtime as per https://github.com/dapr/dapr/issues/2342
 
-      return Mono.subscriberContext().flatMap(
+      return Mono.deferContextual(
               context -> this.<CommonProtos.InvokeResponse>createMonoWithErrorHandling(
                       it -> intercept(context, asyncStub).invokeService(envelope, it),
-                      errorResponseParser)
+                      errorResponseParser
+              )
       ).flatMap(
               it -> {
                 try {
@@ -388,7 +392,7 @@ public class DaprClientGrpc extends AbstractDaprClient {
       }
       DaprProtos.InvokeBindingRequest envelope = builder.build();
 
-      return Mono.subscriberContext().flatMap(
+      return Mono.deferContextual(
               context -> this.<DaprProtos.InvokeBindingResponse>createMono(
                   it -> intercept(context, asyncStub).invokeBinding(envelope, it)
               )
@@ -435,7 +439,7 @@ public class DaprClientGrpc extends AbstractDaprClient {
 
       DaprProtos.GetStateRequest envelope = builder.build();
 
-      return Mono.subscriberContext().flatMap(
+      return Mono.deferContextual(
           context ->
               this.<DaprProtos.GetStateResponse>createMono(
                   it -> intercept(context, asyncStub).getState(envelope, it)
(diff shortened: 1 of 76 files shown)


sdk/src/main/java/io/dapr/client/DefaultDaprErrorResponseParser.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.charset.StandardCharsets;
+
+import static io.dapr.client.ObjectSerializer.OBJECT_MAPPER;
+
+public class DefaultDaprErrorResponseParser implements DaprErrorResponseParser {
+
+    DaprError error;
+
+    @Override
+    public DaprException parse(@NotNull okhttp3.Response response) throws IOException {
```
artursouza, line 17: Two points here. First, can we make this interface not depend on Okhttp3? This way we can migrate to any http client in the near future.

Second point is the the http implementation will be deprecated and removed soon and gRPC will be the only implementation. Users of the SDK should not care about the protocol, except for service invocation.

So, I see two ways to proceed with this PR: make this interface protocol agnostic, making it work on with gRPC and http; Or create a new method for HTTP service invocation only and make this be used only in HTTP service invocation.

Happy to get other idea on this that would address my first two concerns.

artursouza (changes requested):

sdk/src/main/java/io/dapr/client/DefaultDaprErrorResponseParser.java
maciek047: Thanks for your input, I've added changes to address the concerns raised, the interface works now both with HTTP and gRPC and is passed via the `DaprClientBuilder`.  Let me know if this is a good direction.

sdk/src/main/java/io/dapr/client/DaprClientGrpc.java
```diff
@@ -316,6 +324,45 @@ public <T> Mono<T> invokeMethod(InvokeMethodRequest invokeMethodRequest, TypeRef
...
+          sink.success(value);
+        }
+
+        @Override
+        public void onError(Throwable t) {
+          if (t instanceof StatusRuntimeException) {
+            StatusRuntimeException statusException = (StatusRuntimeException) t;
+            int statusCode = statusException.getStatus().getCode().value();
+            byte[] errorDetails = statusException.getStatus().getDescription() != null
+                    ? statusException.getStatus().getDescription().getBytes(StandardCharsets.UTF_8)
```
artursouza, line 343: Get encoding from our properties. See the Properties.java file.

sdk/src/main/java/io/dapr/client/DaprErrorResponseParser.java
```diff
@@ -0,0 +1,8 @@
+package io.dapr.client;
+
+import io.dapr.exceptions.DaprException;
+import java.io.IOException;
+
+public interface DaprErrorResponseParser {
+    DaprException parse(int statusCode, byte[] response) throws IOException;
```
artursouza, line 7: What is statusCode? These are different for gRPC or HTTP. Maybe add a new param for the protocol, so the implementation can handle both. Another option would be to have one method for gRPC and one for HTTP.

sdk/src/main/java/io/dapr/client/DefaultDaprErrorResponseParser.java
```diff
@@ -0,0 +1,35 @@
...
+
+import static io.dapr.client.ObjectSerializer.OBJECT_MAPPER;
+
+public class DefaultDaprErrorResponseParser implements DaprErrorResponseParser {
+
+    DaprError error;
+
+    @Override
+    public DaprException parse(int statusCode, byte[] response) {
+        String errorMessage = (response == null || new String(response).isEmpty()) ? "HTTP status code: " + statusCode : new String(response, StandardCharsets.UTF_8);
```
artursouza, line 16: Please, reference the encoding from the Properties.java file.

sdk/src/main/java/io/dapr/client/DefaultDaprGrpcErrorResponseParser.java
```diff
@@ -0,0 +1,16 @@
...
+import io.dapr.exceptions.DaprException;
+import io.grpc.Status;
+
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.ExecutionException;
+
+public class DefaultDaprGrpcErrorResponseParser implements DaprErrorResponseParser {
+    @Override
+    public DaprException parse(int statusCode, byte[] errorDetails) {
+        String errorMessage = new String(errorDetails, StandardCharsets.UTF_8);
```
artursouza, line 13: same.

sdk/src/main/java/io/dapr/client/DefaultDaprErrorResponseParser.java
```diff
@@ -0,0 +1,35 @@
...
+        String errorCode = "UNKNOWN";
+        DaprException unknownException = new DaprException(errorCode, errorMessage);
+
+        if ((response != null) && (response.length != 0)) {
+            try {
+                error = OBJECT_MAPPER.readValue(response, DaprError.class);
+            } catch (IOException e) {
+                return new DaprException("UNKNOWN", new String(response, StandardCharsets.UTF_8));
```
artursouza, line 24: same.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -139,7 +144,7 @@ public void invokeGetMethod() throws IOException {
     mockInterceptor.addRule()
       .get("http://127.0.0.1:3500/v1.0/get")
       .respond(serializer.serialize(EXPECTED_RESULT));
-    DaprHttp daprHttp = new DaprHttp(Properties.SIDECAR_IP.get(), 3500, okHttpClient);
+    DaprHttp daprHttp = new DaprHttp(Properties.SIDECAR_IP.get(), 3500, okHttpClient, DEFAULT_PARSER);
```
artursouza, line 147: Use an overloaded version of the constructor instead. So, there will not be changes needed everywhere.

artursouza (changes requested):
I like where this is going. Thanks a lot.

sdk/src/main/java/io/dapr/client/DaprErrorResponseParser.java
maciek047: My intention with the DaprErrorResponseParser interface was to provide separate implementations for handling different protocols (as demonstrated by DefaultDaprHttpErrorResponseParser and DefaultDaprGrpcErrorResponseParser). I believe this approach maintains a clean separation of concerns between gRPC and HTTP error handling, which is why I opted not to include a protocol parameter in the interface. Ofc if you think there is a strong reason for having a single implementation handling both protocols, then I can change it

ParrySMS:
@maciek047  Hi, is there any updated?

ParrySMS:
Hi @maciek047 , Could you please help to add more unit test to fix the coverage issue? It will not take a long time with GPT.

cicoyle:
gentle ping @maciek047 - mind updating this PR?

artursouza:
Issue #783 is now resolved by https://github.com/dapr/java-sdk/pull/1009 - it is a different approach but makes the raw error payload available.

Issue #839, opened 2023-03-13
Workflow implementation in SDK


RyanLettieri:
## Describe the proposal
Add Java SDK support for dapr workflow. This will include both the authoring and management functionalities.

Related dapr/dapr proposal: https://github.com/dapr/dapr/issues/4576

artursouza:
I think the management APIs are P1 but authoring is a P2. Maybe we should have separate issues, one for each?

RyanLettieri:
@dapr/maintainers-java-sdk can we move this from the v1.10 milestone to the v1.12 milestone?

mukundansundar:
> @dapr/maintainers-java-sdk can we move this from the v1.10 milestone to the v1.12 milestone?

This milestone is Java sdk milestone. Which is 1.10. Current version is 1.8. 1.9 is current milestone and 1.10 is the next one.

bderusha:
My team is interested in contributing some of this implementation. Is there anyone else currently working on this that we should coordinate with?

bderusha:
@dapr/maintainers-java-sdk Would love to get early feedback on this WIP PR #857 to validate the overall approach

PR #857, opened 2023-05-11
Initial Workflow SDK

bderusha:
# Description

This PR covers just enough of the authoring and management APIs to allow for anyone to be able to add the rest of them without requiring much/any coordination around structure.

Design decisions of note:
- `sdk-workflows` is implemented as a separate optional package (like actors)
- Runtime registration follows conventions/patterns used by sdk-actors with `WorkflowRuntime.getInstance().registerWorkflow(...)`

> **Warning**
> **Bug Alert:** 
> Ran into a `panic: runtime error: invalid memory address or nil pointer dereference` error while using `durabletask-java@v1.1.0` with dapr runtime `1.10.7`.  It seems to be related to some mismatch between the client and the `durabletask-go@v0.1.3` that comes with dapr 1.10.  When run against dapr runtime `1.11.0-rc.1` everything works nicely.

## Issue reference

#839 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


nyemade-uversky:
Putting this on our radars  @cgillum , @DeepanshuA , @RyanLettieri

bderusha:
@cgillum , @DeepanshuA , @RyanLettieri, @nyemade-uversky This PR is almost done in terms of foundations.  Would love a quick review before we add a bunch more methods if you have time this week.

examples/src/main/java/io/dapr/examples/workflows/DemoWorkflow.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Implementation of the DemoWorkflow for the server side.
+ */
+public class DemoWorkflow extends Workflow {
+
+  @Override
+  public void run(WorkflowContext ctx) {
+    System.out.println("Hi, my name is " + ctx.getName());
+    System.out.println("Waiting for event: 'myEvent'...");
```
cgillum, line 30: The problem with `println` statements in workflow code is that they will be repeated each time the workflow replays. In this case, everything before the external event will be written twice. The poor-man's way of avoiding this is to wrap each `println` in a `if (!ctx.getIsReplaying()) { ... }` block. We may want to think of how/whether we can bake this into the workflow context, and probably integrate it with the common Java logging framework.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,86 @@
...
+      );
+    } catch (NoSuchMethodException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Start the Workflow runtime.
+   */
+  public void start() {
```
cgillum, line 72: We should document whether or not this call is blocking.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,86 @@
...
+    this.worker.start();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void close() {
+    if (this.worker != null) {
+      this.worker.close();
```
cgillum, line 83: Should we set `this.worker` to `null` after closing it, to help make this `close()` method more idempotent?

examples/src/main/java/io/dapr/examples/workflows/DemoWorkflowService.java
```diff
@@ -0,0 +1,40 @@
...
+   * The main method of this app.
+   *
+   * @param args The port the app will listen on.
+   * @throws Exception An Exception.
+   */
+  public static void main(String[] args) throws Exception {
+    // Register the Workflow with the runtime.
+    WorkflowRuntime.getInstance().registerWorkflow(DemoWorkflow.class);
+    WorkflowRuntime.getInstance().start();
+    System.out.println("Start workflow runtime");
```
cgillum, line 38: Since `start()` is non-blocking, won't the program exit immediately after we start the runtime?

sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java
```diff
@@ -0,0 +1,121 @@
...
+    return channel;
+  }
+
+  /**
+   * Schedules a new workflow using DurableTask client.
+   *
+   * @param workflowName name of workflow to start.
+   * @return String for new Workflow instance id.
+   */
+  public String scheduleNewWorkflow(String workflowName) {
```
cgillum, line 95: Consider adding an overload that allows the user to specify an instance ID. This is common, especially for workflows that correspond to external entities (document workflows, etc.) and helps users who need their workflows to be singletons.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowContext.java
```diff
@@ -0,0 +1,33 @@
...
+public interface WorkflowContext {
+
+  String getName();
+
+  String getInstanceId();
+
+  void complete(Object o);
+
+  Task waitForExternalEvent(String eventName, Duration timeout);
+}
```
cgillum, line 33: I assume that over the course of this PR we'll add the other features, like calling activities, timers, and child-workflows?

sdk-workflows/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-testing</artifactId>
+      <version>${grpc.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.microsoft</groupId>
+      <artifactId>durabletask-client</artifactId>
+      <version>1.1.0</version>
+    </dependency>
```
cgillum, line 74: I'm not very familiar with pom.xml, but it seems there's a lot of duplication between this pom.xml and `sdk-workflows/pom.xml` in the parent directory. Is this intentional?

sdk-workflows/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+
+  <parent>
+    <groupId>io.dapr</groupId>
+    <artifactId>dapr-sdk-parent</artifactId>
+    <version>1.9.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>dapr-sdk-workflows</artifactId>
+  <packaging>jar</packaging>
+  <version>1.9.0-SNAPSHOT</version>
```
cgillum, line 15: For other projects, we've been using a different version number for the workflow SDK to help indicate that it's not yet production quality. Should we do something similar for `dapr-sdk-workflows`? For example, like having the version be `0.1.0-SNAPSHOT`?

cgillum (commented):
This is a great start! Some comments below.

examples/src/main/java/io/dapr/examples/workflows/DemoWorkflowService.java
bderusha: That hasn't been my experience running this example code. The new thread created by `start()` continues to run until it receives an interrupt.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowContext.java
bderusha: It would be my preference to merge this PR at its current level of functionality to allow for mgmt and authoring APIs to be contributed in parallel. But the minimum feature set required to merge this PR is ultimately something that requires agreement from the maintainers, so feedback certainly welcome.

If we were to expand upon something, I would say the mgmt APIs would be more valuable.  These initial authoring APIs are only implemented for ease of testing.

sdk-workflows/pom.xml
bderusha: I believe it is, but as it is a direct port of the file in `sdk-actors` there could be unnecessary duplication in both. Would love to know generally if there are anti-patterns or dated practices in `sdk-actors` that we should be wary of proliferating.

sdk-workflows/pom.xml
bderusha: Great question for maintainers more familiar with this codebase.  I am in favor of the approach. It is more defensible to have incremental merges that way as well.

bderusha:
@cgillum Thanks for the feedback! I pushed up a new round of changes to address them

examples/src/main/java/io/dapr/examples/workflows/DemoWorkflowClient.java
```diff
@@ -0,0 +1,56 @@
...
+      String workflowName = DemoWorkflow.class.getCanonicalName();
+      String instanceId = client.scheduleNewWorkflow(workflowName);
+      System.out.printf("Started new workflow instance with random ID: %s%n", instanceId);
+
+      System.out.println("Sleep and allow this workflow instance to timeout...");
+      TimeUnit.SECONDS.sleep(10);
+
+      System.out.println("*****");
+      String instanceToTerminateId = "terminateMe";
+      client.scheduleNewWorkflow(workflowName, null, instanceToTerminateId);
```
cgillum, line 44: On line 35 we get the workflow name from the workflow class canonical name. It might be useful and slightly more type-safe to have an overload of `scheduleNewWorkflow` that takes a class as a parameter (and internally uses that to get the name) as an alternative to a string for the name.

cgillum (commented):
A few additional comments/questions.

sdk-workflows/pom.xml
bderusha: Each package needs to declare its own dependencies, but there were deps in actors that were not needed by workflows. Those have been cleaned up.

examples/src/main/java/io/dapr/examples/workflows/DemoWorkflowClient.java
bderusha: Great call. Given that the class is the only way to register a workflow right now, I opted to only offer class-based parameters.   It only makes sense to overload with strings if/when users can register workflows with strings.

sdk-workflows/src/test/java/io/dapr/workflows/runtime/DaprWorkflowContextImplTest.java
```diff
@@ -0,0 +1,135 @@
...
+  public void waitForExternalEventTest() {
+    String expectedEvent = "TestEvent";
+    Duration expectedDuration = Duration.ofSeconds(1);
+
+    context.waitForExternalEvent(expectedEvent, expectedDuration);
+    verify(mockInnerContext, times(1)).waitForExternalEvent(expectedEvent, expectedDuration);
+  }
+
+  @Test
+  public void waitForExternalEventTimeoutTest() {
```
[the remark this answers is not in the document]
bderusha: They are. This was cruft they should have been removed.

sdk-workflows/src/test/java/io/dapr/workflows/runtime/DaprWorkflowContextImplTest.java
```diff
@@ -0,0 +1,135 @@
...
+    PrintStream originalOut = System.out;
+    System.setOut(mockOut);
+    when(mockInnerContext.getIsReplaying()).thenReturn(false);
+    DaprWorkflowContextImpl testContext = new DaprWorkflowContextImpl(mockInnerContext);
+
+    String expectedArg = "test print";
+    testContext.getOut().println(expectedArg);
+
+    verify(mockOut, times(1)).println(expectedArg);
+    System.setOut(originalOut);
```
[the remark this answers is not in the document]
bderusha: Yes and obviated by the `getLogger` implementation. Tests have been removed

sdk-workflows/src/main/java/io/dapr/workflows/runtime/DaprWorkflowContextImpl.java
```diff
@@ -0,0 +1,81 @@
...
+   */
+  public void complete(Object output) {
+    this.innerContext.complete(output);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public Task<Void> waitForExternalEvent(String eventName, Duration timeout) {
+    return innerContext.waitForExternalEvent(eventName, timeout);
```
cgillum, line 79: nit
```suggestion
    return this.innerContext.waitForExternalEvent(eventName, timeout);
```

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param <T> any Workflow type
+   * @param clazz the class being registered
+   */
+  public <T extends Workflow> void registerWorkflow(Class<T> clazz) {
+    try {
+      this.builder = this.builder.addOrchestration(
+          new OrchestratorWrapper<>(clazz)
+      );
+    } catch (NoSuchMethodException e) {
```
cgillum, line 70: When might `NoSuchMethodException` get thrown?

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,103 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  /**
+   * Start the Workflow runtime processing items on a non-blocking
+   * background thread indefinitely or until that thread is interrupted.
+   */
+  public void start() {
+    this.worker = this.builder.build();
```
cgillum, line 80: Should we check to see whether an existing worker has already started before initializing a new worker? Perhaps a simple null-check and a `throw` would be enough?

cgillum (commented):
Just a few small things/questions.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
bderusha: It's thrown by the `OrchestratorWrapper` when we call `clazz.getDeclaredConstructor();`

I have moved this `try/catch` block closer to where it is thrown.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/OrchestratorWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public TaskOrchestration create() {
+    return ctx -> {
+      try {
+        T workflow = this.workflowConstructor.newInstance();
+        workflow.run(new DaprWorkflowContextImpl(ctx, null));
```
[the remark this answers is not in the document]
bderusha: It's populated in the constructor if it's null.  I have updated the code to overload the constructor to make this default behavior less confusing.

bderusha:
Thanks @cgillum Pushed up a quick set of changes based on your feedback.

Is there anyone else we need to loop into this PR for it to land? It feels like we're getting close here

cgillum:
@bderusha this PR looks good to me. However, I don't have permissions to merge it as I'm not a maintainer for the Dapr Java SDK (my signoffs don't count either). We'll need @dapr/approvers-java-sdk and @dapr/maintainers-java-sdk to take a look.

FYI @nyemade-uversky in case we need to ping relevant folks directly.

examples/pom.xml
```diff
@@ -118,6 +118,16 @@
       <artifactId>dapr-sdk-actors</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.dapr</groupId>
+      <artifactId>dapr-sdk-workflows</artifactId>
+      <version>0.1.0-alpha-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>com.microsoft</groupId>
```
johnewart, line 127: Won't this already be a transitive dependency from the workflow SDK?

pom.xml
```diff
@@ -7,7 +7,27 @@
   <groupId>io.dapr</groupId>
   <artifactId>dapr-sdk-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.9.0-SNAPSHOT</version>
+    <dependencies>
```
johnewart, line 10: Do these need to be in the parent POM? They're only needed for the workflow project, right?

johnewart (commented):
Looks like a good start! I just had a few comments about dependencies that were added to POMs (looks like perhaps dependencies got added in places they don't need to be).

bderusha:
> Looks like a good start! I just had a few comments about dependencies that were added to POMs (looks like perhaps dependencies got added in places they don't need to be).

Thanks @johnewart. Good catch on those dependencies. Should be all set now.

nyemade-uversky:
@dapr/maintainers-java-sdk would you be able to take a look at this PR?

olitomlinson:
@cgillum I don't know if this is the time or the place, but can I ask that code examples are included that demonstrate how to unit test a Workflow from an end users perspective i.e how to mock Activity calls, timers, continueAsNew etc etc

sdk-workflows/src/main/java/io/dapr/workflows/runtime/DaprWorkflowContextImpl.java
```diff
@@ -0,0 +1,95 @@
...
+   */
+  public DaprWorkflowContextImpl(TaskOrchestrationContext context, Logger logger) throws IllegalArgumentException {
+    if (context == null) {
+      throw new IllegalArgumentException("Context cannot be null");
+    } else {
+      this.innerContext = context;
+    }
+
+    if (logger == null) {
+      throw new IllegalArgumentException("Logger cannot be null");
```
DeepanshuA, line 52: Should we just create a logger, if not provided - instead of IllegalArgument?

bderusha: That is the idea behind the [`public DaprWorkflowContextImpl(TaskOrchestrationContext context)`](https://github.com/dapr/java-sdk/pull/857/files/9b2881b830eb6d87ca9de0912346f4b31c817576#diff-d11922cd65de99fb47af4c04136f5cdd1a470855f363d5ac79afd1ce8e5a3404R33-R35) constructor.  I feel that, in that context, if you are calling an overloaded constructor without the required arguments an error is appropriate.

I could be convinced otherwise, but that was the intent.

examples/src/main/java/io/dapr/examples/workflows/DemoWorkflowClient.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * The main method.
+   * @param args Input arguments (unused).
+   * @throws InterruptedException If program has been interrupted.
+   */
+  public static void main(String[] args) throws InterruptedException {
+    DaprWorkflowClient client = new DaprWorkflowClient();
+
+    try (client) {
+      System.out.println("*****");
```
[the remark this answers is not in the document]
bderusha: The other examples tend to use `s.o.p` to keep things simple given that the examples are meant to be run from the command line. I think we should keep to that convention here, but let me know if you think that something unique to this example warrants a logger instead.

https://github.com/dapr/java-sdk/blob/master/examples/src/main/java/io/dapr/examples/actors/DemoActorClient.java#L64
https://github.com/dapr/java-sdk/blob/master/examples/src/main/java/io/dapr/examples/configuration/grpc/ConfigurationClient.java#L61
https://github.com/dapr/java-sdk/blob/master/examples/src/main/java/io/dapr/examples/exception/Client.java#L40

DeepanshuA:
> @dapr/approvers-java-sdk @DeepanshuA Can someone help kick off the build checks for this?

@artursouza @mukundansundar Could you please help trigger build?

nyemade-uversky:
@artursouza, @mukundansundar and @bderusha would it be possible to get this PR merged this week? Looks like we're waiting on the build to pass, is that the only ting blocking this pr rn?

bderusha:
> @artursouza, @mukundansundar and @bderusha would it be possible to get this PR merged this week? Looks like we're waiting on the build to pass, is that the only ting blocking this pr rn?

@nyemade-uversky There is a single failing test for `Build jdk:11 sb:2.5.7` around pub/sub. I have some time carved out to investigate tomorrow, but if anyone has time or insights to share I would welcome them.

bderusha:
Thank you for rerunning the tests @mukundansundar Looks like that failure was intermittent. All tests are passing now. 

FYI @nyemade-uversky

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,103 @@
...
+  private DurableTaskGrpcWorkerBuilder builder;
+  private DurableTaskGrpcWorker worker;
+
+  private WorkflowRuntime() throws IllegalStateException {
+
+    if (instance != null) {
+      throw new IllegalStateException("WorkflowRuntime should only be constructed once");
+    }
+
+    int port = Properties.GRPC_PORT.get();
```
artursouza, line 35: Can't this use the same client as the main SDK since it invokes the same gRPC stub?

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,103 @@
...
+    return instance;
+  }
+
+  /**
+   * Registers a Workflow object.
+   *
+   * @param <T> any Workflow type
+   * @param clazz the class being registered
+   */
+  public <T extends Workflow> void registerWorkflow(Class<T> clazz) {
```
artursouza, line 65: Mono

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,103 @@
...
+      this.worker = this.builder.build();
+      this.worker.start();
+    }
+  }
+
+  /**
+   * Start the Workflow runtime processing items on the current thread
+   * and block indefinitely or until that thread is interrupted.
+   */
+  public void startAndBlock() {
```
artursouza, line 86: Mono. Also, it will remove the need of both methods since user can decide to block or not via the Mono object.

artursouza (changes requested):
This is great. I have provided some feedback.

Overall, I think it should be built using more of the main SDK connectivity - otherwise it will require duplication of code to handle connection to sidecar.

Last but not least, please, consider making this a separate project under dapr-sandbox org - instructions at https://github.com/dapr-sandbox/

I do think I should continue helping in the review to keep consistent user experience for the Java SDKs since not all decisions are documented.

bderusha:
> This is great. I have provided some feedback.
> 
> Overall, I think it should be built using more of the main SDK connectivity - otherwise it will require duplication of code to handle connection to sidecar.
> 
> Last but not least, please, consider making this a separate project under dapr-sandbox org - instructions at https://github.com/dapr-sandbox/
> 
> I do think I should continue helping in the review to keep consistent user experience for the Java SDKs since not all decisions are documented.

@artursouza Thank you for the feedback. The code comments all make sense. I am curious to learn more about `dapr-sandbox` 
 though. Can you help me understand more about why this feature is a good fit for a sandbox project?

The context I am thinking about here is that other SDKs have early workflow support in their main repos ([python-sdk](https://github.com/dapr/python-sdk/tree/master/ext/dapr-ext-workflow), [dotnet-sdk](https://github.com/dapr/dotnet-sdk/tree/master/src/Dapr.Workflow)) and of course the workflow engine itself is in the mainline [dapr/dapr](https://github.com/dapr/dapr/tree/master/pkg/runtime/wfengine).

At first blush it feels like this feature is no different from other [alpha APIs](https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java#L251) introduced to the dapr runtime and then supported by the SDKs. But I am admittedly just learning about the `dapr-sandbox` so you may have additional context that draws a clearer line to why the sandbox is a good home. Regardless of where this code lands, I am interested in learning more so that we can streamline future contributions :) Thanks!

FYI I will be on vacation most of July and slow to respond, so I am tagging others who might be interested in this while I'm out.
@nyemade-uversky @cgillum @Narmathabala

artursouza:
> > This is great. I have provided some feedback.
> > Overall, I think it should be built using more of the main SDK connectivity - otherwise it will require duplication of code to handle connection to sidecar.
> > Last but not least, please, consider making this a separate project under dapr-sandbox org - instructions at https://github.com/dapr-sandbox/
> > I do think I should continue helping in the review to keep consistent user experience for the Java SDKs since not all decisions are documented.
> 
> @artursouza Thank you for the feedback. The code comments all make sense. I am curious to learn more about `dapr-sandbox` though. Can you help me understand more about why this feature is a good fit for a sandbox project?
> 
> The context I am thinking about here is that other SDKs have early workflow support in their main repos ([python-sdk](https://github.com/dapr/python-sdk/tree/master/ext/dapr-ext-workflow), [dotnet-sdk](https://github.com/dapr/dotnet-sdk/tree/master/src/Dapr.Workflow)) and of course the workflow engine itself is in the mainline [dapr/dapr](https://github.com/dapr/dapr/tree/master/pkg/runtime/wfengine).
> 
> At first blush it feels like this feature is no different from other [alpha APIs](https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java#L251) introduced to the dapr runtime and then supported by the SDKs. But I am admittedly just learning about the `dapr-sandbox` so you may have additional context that draws a clearer line to why the sandbox is a good home. Regardless of where this code lands, I am interested in learning more so that we can streamline future contributions :) Thanks!
> 
> FYI I will be on vacation most of July and slow to respond, so I am tagging others who might be interested in this while I'm out. @nyemade-uversky @cgillum @NarmathaBala

Making this a dapr-sandbox repo would need to be consistent with other SDKs, so all of them would move the WF part to it. In this case, there is a different since it is for composing workflow and not simply an Alpha API. The wf invocation is a simple Alpha API and it should still be here.

The main driving factor for this is that this is at a different maturity level and a new artifact. The fact that it is versioned differently from the other artifacts of the SDK says it all. That is why I am leaning torwards this being a separate SDK. For this to be merged in this repo, we would also need to update all the release scripts to handle specific versioning for this artifact. See https://github.com/dapr/java-sdk/wiki/Release-Process

NarmathaBala:
> > > This is great. I have provided some feedback.
> > > Overall, I think it should be built using more of the main SDK connectivity - otherwise it will require duplication of code to handle connection to sidecar.
> > > Last but not least, please, consider making this a separate project under dapr-sandbox org - instructions at https://github.com/dapr-sandbox/
> > > I do think I should continue helping in the review to keep consistent user experience for the Java SDKs since not all decisions are documented.
> > 
> > 
> > @artursouza Thank you for the feedback. The code comments all make sense. I am curious to learn more about `dapr-sandbox` though. Can you help me understand more about why this feature is a good fit for a sandbox project?
> > The context I am thinking about here is that other SDKs have early workflow support in their main repos ([python-sdk](https://github.com/dapr/python-sdk/tree/master/ext/dapr-ext-workflow), [dotnet-sdk](https://github.com/dapr/dotnet-sdk/tree/master/src/Dapr.Workflow)) and of course the workflow engine itself is in the mainline [dapr/dapr](https://github.com/dapr/dapr/tree/master/pkg/runtime/wfengine).
> > At first blush it feels like this feature is no different from other [alpha APIs](https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java#L251) introduced to the dapr runtime and then supported by the SDKs. But I am admittedly just learning about the `dapr-sandbox` so you may have additional context that draws a clearer line to why the sandbox is a good home. Regardless of where this code lands, I am interested in learning more so that we can streamline future contributions :) Thanks!
> > FYI I will be on vacation most of July and slow to respond, so I am tagging others who might be interested in this while I'm out. @nyemade-uversky @cgillum @NarmathaBala
> 
> Making this a dapr-sandbox repo would need to be consistent with other SDKs, so all of them would move the WF part to it. In this case, there is a different since it is for composing workflow and not simply an Alpha API. The wf invocation is a simple Alpha API and it should still be here.
> 
> The main driving factor for this is that this is at a different maturity level and a new artifact. The fact that it is versioned differently from the other artifacts of the SDK says it all. That is why I am leaning torwards this being a separate SDK. For this to be merged in this repo, we would also need to update all the release scripts to handle specific versioning for this artifact. See https://github.com/dapr/java-sdk/wiki/Release-Process

@artursouza - @nyemade-uversky shared with us re: letting this PR merge into the Dapr repo. We have Bill out of office for the month of July. Considering the feedback suggested will involved some deep changes, we agreed that it was best for Bill to pick this back up in August. I hope that works for the team.

brunoborges:
> /cc @brunoborges do you mind taking a look at this PR too? It is from Msft.

Hey Artur, sorry but I'm afraid I don't have the necessary knowledge of the Dapr SDK, nor the time, to provide meaningful feedback here.

craprivatebeta:
> > > > This is great. I have provided some feedback.
> > > > Overall, I think it should be built using more of the main SDK connectivity - otherwise it will require duplication of code to handle connection to sidecar.
> > > > Last but not least, please, consider making this a separate project under dapr-sandbox org - instructions at https://github.com/dapr-sandbox/
> > > > I do think I should continue helping in the review to keep consistent user experience for the Java SDKs since not all decisions are documented.
> > > 
> > > 
> > > @artursouza Thank you for the feedback. The code comments all make sense. I am curious to learn more about `dapr-sandbox` though. Can you help me understand more about why this feature is a good fit for a sandbox project?
> > > The context I am thinking about here is that other SDKs have early workflow support in their main repos ([python-sdk](https://github.com/dapr/python-sdk/tree/master/ext/dapr-ext-workflow), [dotnet-sdk](https://github.com/dapr/dotnet-sdk/tree/master/src/Dapr.Workflow)) and of course the workflow engine itself is in the mainline [dapr/dapr](https://github.com/dapr/dapr/tree/master/pkg/runtime/wfengine).
> > > At first blush it feels like this feature is no different from other [alpha APIs](https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/DaprClientGrpc.java#L251) introduced to the dapr runtime and then supported by the SDKs. But I am admittedly just learning about the `dapr-sandbox` so you may have additional context that draws a clearer line to why the sandbox is a good home. Regardless of where this code lands, I am interested in learning more so that we can streamline future contributions :) Thanks!
> > > FYI I will be on vacation most of July and slow to respond, so I am tagging others who might be interested in this while I'm out. @nyemade-uversky @cgillum @NarmathaBala
> > 
> > 
> > Making this a dapr-sandbox repo would need to be consistent with other SDKs, so all of them would move the WF part to it. In this case, there is a different since it is for composing workflow and not simply an Alpha API. The wf invocation is a simple Alpha API and it should still be here.
> > The main driving factor for this is that this is at a different maturity level and a new artifact. The fact that it is versioned differently from the other artifacts of the SDK says it all. That is why I am leaning torwards this being a separate SDK. For this to be merged in this repo, we would also need to update all the release scripts to handle specific versioning for this artifact. See https://github.com/dapr/java-sdk/wiki/Release-Process
> 
> @artursouza - @nyemade-uversky shared with us re: letting this PR merge into the Dapr repo. We have Bill out of office for the month of July. Considering the feedback suggested will involved some deep changes, we (Bill and I) agreed that it was best for Bill to pick this back up in August. I hope that works for the team.

That works for me. Yes, the PR can be added here.

mukundansundar:
@NarmathaBala @bderusha 
Do we have an ETA for updates in this PR? Thanks.

sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java
```diff
@@ -0,0 +1,148 @@
...
+        .build();
+  }
+
+  /**
+   * Static method to create the GRPC Channel for the DurableTaskClient.
+   *
+   * @return a Managed GRPC channel.
+   * @throws IllegalStateException if the GRPC port is invalid.
+   */
+  private static ManagedChannel createGrpcChannel() throws IllegalStateException {
```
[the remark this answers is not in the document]
bderusha: @artursouza I moved this logic into the shared `NetworkUtils` class

bderusha:
@mukundansundar @artursouza I believe I have addressed all outstanding comments. Please take a look when you get a chance.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/OrchestratorWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+class OrchestratorWrapper<T extends Workflow> implements TaskOrchestrationFactory {
+  private final Constructor<T> workflowConstructor;
+  private final String name;
+
+  public OrchestratorWrapper(Class<T> clazz) {
+    this.name = clazz.getCanonicalName();
+    try {
+      this.workflowConstructor = clazz.getDeclaredConstructor();
+    } catch (NoSuchMethodException e) {
+      throw new RuntimeException(e);
```
skyao, line 34: Suggest to add a error message with more concrete information when we wrap a catch exception to RuntimeException to improve the experience.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/OrchestratorWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public TaskOrchestration create() {
+    return ctx -> {
+      try {
+        T workflow = this.workflowConstructor.newInstance();
+        workflow.runAsync(new DaprWorkflowContextImpl(ctx)).block();
+      } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+        throw new RuntimeException(e);
```
skyao, line 50: Suggest to add a error message with more concrete information when we wrap a catch exception to RuntimeException to improve the experience.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/OrchestratorWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+    return name;
+  }
+
+  @Override
+  public TaskOrchestration create() {
+    return ctx -> {
+      try {
+        T workflow = this.workflowConstructor.newInstance();
+        workflow.runAsync(new DaprWorkflowContextImpl(ctx)).block();
+      } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
```
skyao, line 49: I found that this try/catch is for the first line :

```java
T workflow = this.workflowConstructor.newInstance();
```

Suggest to move the second line out of this try / catch.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/Workflow.java
```diff
@@ -0,0 +1,46 @@
...
+
+package io.dapr.workflows.runtime;
+
+import com.google.protobuf.Empty;
+import reactor.core.publisher.Mono;
+
+/**
+ * Common interface for workflow implementations.
+ */
+public abstract class Workflow {
```
skyao, line 22: This Workflow class and related WorkflowContext/DaprWorkflowContextImpl should move the its parent package "io.dapr.workflows". Because they are basic definition for workflow, not for workflow runtime.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -0,0 +1,91 @@
...
+import io.dapr.utils.NetworkUtils;
+import reactor.core.publisher.Mono;
+
+/**
+ * Contains methods to register workflows and activities.
+ */
+public class WorkflowRuntime implements AutoCloseable {
+
+  private static volatile WorkflowRuntime instance;
+  private DurableTaskGrpcWorkerBuilder builder;
```
skyao, line 28: The code for this class is a bit confusing, because there are two responsibilities in this class: 

1. how to build up WorkflowRuntime: and WorkflowRuntime should only have one instance
2. workflowRuntime implementation

I suggest to move all the code about how to build up a WorkflowRuntime instance to a standalone class named WorkflowRuntimeBuilder, so that we can get simple and clean code for WorkflowRuntime.

WorkflowRuntimeBuilder class should contain:

- private DurableTaskGrpcWorkerBuilder builder;
- public <T extends Workflow> void registerWorkflow(Class<T> clazz) {}
- a new Build() method to do something like this.worker = this.builder.build();
- the logic that we should only have one instance of WorkflowRuntime

skyao (changes requested):
it is a gread PR! 

I put some very detail comments about the code implementation.

Let's make it perfect.

sdk-workflows/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+
+  <parent>
+    <groupId>io.dapr</groupId>
+    <artifactId>dapr-sdk-parent</artifactId>
+    <version>1.10.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>dapr-sdk-workflows</artifactId>
+  <packaging>jar</packaging>
+  <version>0.1.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
bderusha: It originally did match the parent and was changed based on feedback, with the plan being to wait until it is stable before aligning versions with the parent version.

sdk-workflows/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.5.2</version>
```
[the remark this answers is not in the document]
bderusha: I was able to use the root dependency. No need to declare a version in this pom.xml

artursouza (changes requested):

sdk-workflows/src/main/java/io/dapr/workflows/runtime/OrchestratorWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  @Override
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public TaskOrchestration create() {
```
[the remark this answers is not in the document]
skyao: > DurableTask does not expose any async methods in the client or runtime. 

This is so bad! So since DurableTask is blocking, then we use reactor in java sdk won't bring any value but make it complex. 

It seems that we need to give up Mono and use normal blocking call.

sdk-workflows/src/main/java/io/dapr/workflows/DaprWorkflowContextImpl.java
```diff
@@ -0,0 +1,101 @@
...
+  public Mono<String> getInstanceId() {
+    return Mono.create(it -> it.success(this.innerContext.getInstanceId()));
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public Mono<Void> complete(Object output) {
+    return Mono.<Empty>create(it -> {
+      this.innerContext.complete(output);
```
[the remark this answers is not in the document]
bderusha: I checked it out and it does not. Removed `Mono`

sdk-workflows/src/main/java/io/dapr/workflows/DaprWorkflowContextImpl.java
```diff
@@ -0,0 +1,101 @@
...
+      it.success();
+    }).then();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public Mono<Void> waitForExternalEvent(String eventName, Duration timeout) {
+    return Mono.<Empty>create(it -> {
+      this.innerContext.waitForExternalEvent(eventName, timeout).await();
```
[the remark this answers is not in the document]
bderusha: I looked into this and I don't believe so, unfortunately. If the Task returned by the method gave us access to the underlying `CompleteableFuture` we could use `Mono.fromFuture()`, but it does not.

sdk-workflows/src/main/java/io/dapr/workflows/Workflow.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Executes the workflow logic.
+   *
+   * @param ctx provides access to methods for scheduling durable tasks and getting information about the current
+   *            workflow instance.
+   * @return A Mono Plan of type Void.
+   */
+  public Mono<Void> runAsync(WorkflowContext ctx) {
+    return Mono.<Empty>create(it -> {
+      this.create().run(ctx);
```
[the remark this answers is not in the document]
bderusha: There is not. Removed the `Mono` wrapper

sdk-workflows/src/main/java/io/dapr/workflows/WorkflowContext.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Waits for an event to be raised with name and returns the event data.
+   *
+   * @param eventName The name of the event to wait for. Event names are case-insensitive.
+   *                  External event names can be reused any number of times; they are not
+   *                  required to be unique.
+   * @param timeout   The amount of time to wait before cancelling the external event task.
+   * @return A Mono Plan of type Void.
+   */
+  Mono<Void> waitForExternalEvent(String eventName, Duration timeout);
```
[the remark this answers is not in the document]
bderusha: Having delved into the DurableTask source, I think it is best to leave the API for this "as is". This timeout value is used to create a timer in the task engine to manage when it needs to wake up and throw a CancellationException if the event has not been received.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/OrchestratorWrapper.java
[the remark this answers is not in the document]
artursouza: I agree. We should keep it blocking until the DurableTask SDK addresses this: https://github.com/microsoft/durabletask-java/issues/160

artursouza (approved):

Issue #779, opened 2022-09-14
Content-Type mismatch in publishEvent call


mukundansundar:
## Expected Behavior

Content-Type mismatch in publishEvent call.

When a string event is published, using the 
```java
client.publishEvent("pubsubName", "topicName", "This is message #0")
```
What is seen in the broker for publish using gRPC protocol is: 
```
{"data":"This is message #0","datacontenttype":"application/json","expiration":"2022-09-14T07:34:04Z","id":"87bc9a7b-ab08-4e7a-8583-3f1e0433ab67","pubsubname":"kafka-pubsub","source":"test-batch-pub","specversion":"1.0","topic":"testingtopic","traceid":"00-6b8bfc12d6d2bb6c7e2f80503628f2b8-8a495373d49513aa-01","traceparent":"00-6b8bfc12d6d2bb6c7e2f80503628f2b8-8a495373d49513aa-01","tracestate":"","type":"com.dapr.event.sent"}
```
Here because the data is not given in the `client.publishEvent` call, it is being inferred by the default serializer as `application/json` and that is the `datacontenttype` field that is set in the cloudEvent in broker. 

On subscription, the subscriber correctly reads the value and users see something like 
```
Subscriber got: This is message #0
```

But when we publish the same data `This is message #0` with the publishEvent call with the content type correctly set
```java
 PublishEventRequest req = new PublishEventRequest("pubsubName", "topicName", "This is message #0");
req.setContentType("text/plain");
client.publishEvent(req)
```

The data seen in the broker is such: 
```
{"data":"\"This is message #0\"","datacontenttype":"text/plain","expiration":"2022-09-14T12:32:37Z","id":"e996716d-1d11-46d7-9112-6866cbb633ce","pubsubname":"kafka-pubsub","source":"test-batch-pub","specversion":"1.0","topic":"simpletest","traceid":"00-a58c282daed4035a660abbf14208cac2-283bb02770d94908-01","traceparent":"00-a58c282daed4035a660abbf14208cac2-283bb02770d94908-01","tracestate":"","type":"com.dapr.event.sent"}
```
and the subscriber actually gets `Subscriber got: "This is message #0"` 

With the right content type given, the serialization is escaping the quotes. But when no content type is given, a wrong content type is assumed. 

## Actual Behavior

Proper match should be there between datacontenttype field and the actually data field when saved as cloudEvent. 

## Steps to Reproduce the Problem

[...]

PR #861, opened 2023-05-20
Fix Content-Type mismatch in publishEvent call

MregXN:
# Description

Please reference the issue #779

If contentType is set as ""text/plain" serialization will be skipped.


sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -165,13 +165,23 @@ public Mono<Void> publishEvent(PublishEventRequest request) {
...
       // It allows CloudEvents to be handled differently, for example.
       String contentType = request.getContentType();
       if (contentType == null || contentType.isEmpty()) {
         contentType = objectSerializer.getContentType();
       }
+
+      byte[] serializedEvent;
+      if (contentType.equals("text/plain")) 
+      {
+        serializedEvent = ((String)data).getBytes();
```
artursouza, line 178: This logic should be in the serializer.

artursouza (changes requested):

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
MregXN: have moved it in latest commit

MregXN:
Hi @artursouza , if I modify the serialization method of the serializer (to exclude quotation marks when serializing a string), I will also need to modify the deserialization method and its corresponding tests, otherwise I may fail to pass most of the tests.

artursouza:
I will need to think this through. It is not a simple change without breaking existing apps.

MregXN:
After I debugged step by step, it is found that this bug is not directly related to the java-sdk serializer. The reason can be divided into the following two parts:

1. When publishing data, in addition to using the serializer, the data will be encapsulated into a cloudenvet envelop in the dapr runtime, and the data will be further encapsulated according to the dataContentType set by the user:
```go
// NewCloudEventsEnvelope returns a map representation of a cloudevents JSON.
func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string, pubsubName string,
	dataContentType string, data []byte, traceParent string, traceState string,
) map[string]interface{} {
	....

	var ceData interface{}
	ceDataField := DataField
	var err error
	if contribContenttype.IsJSONContentType(dataContentType) {
		err = unmarshalPrecise(data, &ceData)
	} else if contribContenttype.IsBinaryContentType(dataContentType) || contribContenttype.IsCloudEventProtobuf(dataContentType, data) {
		ceData = base64.StdEncoding.EncodeToString(data)
		ceDataField = DataBase64Field
	} else {
		ceData = string(data)
	}

	if err != nil {
		ceData = string(data)
	}

	....
}
```

And there are similar operations when using gRPC protocal. Can find more details in `dapr/pkg/runtime/runtime.go publishMessageGRPC()`

So the string `"This is message #0"`  will store differently in this envelop if user set the dataContentType as `text/plain`  instead of `application/json` .


2.  But when subscribe there is no  further process of the data according to the dataContentType. 

As http subscriber controller for example, it uses the annotation @RequsetBody to parse the data to CloudEvent in default way. But there is no further process of the data according to the dataContentType
```java
...
    @Topic(...)
    @PostMapping(...)
    public Mono<ResponseEntity> getCheckout(@RequestBody(required = false) CloudEvent<Order> cloudEvent) {
        return Mono.fromSupplier(() -> {
            try {
                logger.info("Subscriber received: " + cloudEvent.getData().getOrderId());
                return ResponseEntity.ok("SUCCESS");
            } catch (Exception e) {
            }
        });
    }
...
```

gPRC subscriber is the same. User always use the data directly but not porcess it according dataContentType.

So we need to **add more data processing instead of modifying the java-sdk serizer when implementing subscriber**.
Or directly **add new methods to CloudEvent (for http subscriber) and TopicEventRequest (for gRPC subscriber)**  to parse data according to dataContentType.
@artursouza @artursouza

MregXN:
> I will need to think this through. It is not a simple change without breaking existing apps.

After I debugged step by step, it is found that this bug is not directly related to the java-sdk serializer. The reason can be divided into the following two parts:

1. When publishing data, in addition to using the serializer, the data will be encapsulated into a cloudenent envelop **in dapr runtime**, and the data will be further encapsulated according to the dataContentType set by the user:
```go
// NewCloudEventsEnvelope returns a map representation of a cloudevents JSON.
func NewCloudEventsEnvelope(id, source, eventType, subject string, topic string, pubsubName string,
	dataContentType string, data []byte, traceParent string, traceState string,
) map[string]interface{} {
	....

	var ceData interface{}
	ceDataField := DataField
	var err error
	if contribContenttype.IsJSONContentType(dataContentType) {
		err = unmarshalPrecise(data, &ceData)
	} else if contribContenttype.IsBinaryContentType(dataContentType) || contribContenttype.IsCloudEventProtobuf(dataContentType, data) {
		ceData = base64.StdEncoding.EncodeToString(data)
		ceDataField = DataBase64Field
	} else {
		ceData = string(data)
	}

	if err != nil {
		ceData = string(data)
	}

	....
}
```

And there are similar operations when using gRPC protocal. Can find more details in `dapr/pkg/runtime/runtime.go publishMessageGRPC()`

So the string `"This is message #0"`  will store differently in this envelop if user set the dataContentType as `text/plain`  instead of `application/json` .


2.  But when subscribe there is no  further process of the data according to the dataContentType. 

As http subscriber controller for example, it uses the annotation @RequsetBody to parse the data to CloudEvent in default way. But there is no further process of the data according to the dataContentType
```java
...
    @Topic(...)
    @PostMapping(...)
    public Mono<ResponseEntity> getCheckout(@RequestBody(required = false) CloudEvent<Order> cloudEvent) {
        return Mono.fromSupplier(() -> {
            try {
                logger.info("Subscriber received: " + cloudEvent.getData().getOrderId());
                return ResponseEntity.ok("SUCCESS");
            } catch (Exception e) {
            }
        });
    }
...
```

gPRC subscriber is the same. User always use the data directly but not porcess it according dataContentType.

So we need to **add more data processing instead of modifying the java-sdk serizer when implementing subscriber**.
Or directly **add new methods to CloudEvent (for http subscriber) and TopicEventRequest (for gRPC subscriber)**  to parse data according to dataContentType.

artursouza:
Those are great insights. I think there is a way for CloudEvent to declare a custom serializer for Jackson. I think it is acceptable to have an indirect dependency on Jackson as long as people can still deserialize it without it if they decide to.

cicoyle:
gentle ping @MregXN, any updates?

salaboy:
I am closing this PR due inactivity @cicoyle

PR #866, opened 2023-05-24
Add additional CloudEvent fields (pubsubname, topic, time, etc)

siebenluke:
# Description

    Added additional CloudEvent fields (pubsubname, topic, time, traceid, traceparent, & tracestate)
    
    Added the com.fasterxml.jackson:jackson-datatype-jsr310 dependency to handle serdes of OffsetDateTime for the CloudEvent time field via ObjectMapper settings .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false) & .findAndRegisterModules()
    
    Updated com.fasterxml.jackson dependencies to the latest 2.15.1
    
    Added OffsetDateTime as timeValue to test the DefaultObjectSerializer
    
    Added more tests for new & old CloudEvent fields in CloudEventTest & DefaultObjectSerializerTest

    Signed-off-by: Luke Sieben <siebenluke@gmail.com>

## Issue reference

Add additional CloudEvent fields (pubsubname, topic, time, traceid, traceparent, & tracestate) per discussion in Discord chat

## Checklist

* [✓] Code compiles correctly
* [✓] Created/updated tests
* [N/A] Extended the documentation


sdk/src/main/java/io/dapr/client/domain/CloudEvent.java
```diff
@@ -76,6 +79,37 @@
   @JsonProperty("data_base64")
   private byte[] binaryData;
 
+  /**
+   * The pubsub component this CloudEvent came from.
+   */
+  private String pubsubname;
```
artursouza, line 85: camelCase on properties, please.

sdk/src/main/java/io/dapr/client/domain/CloudEvent.java
```diff
@@ -127,7 +161,89 @@ public CloudEvent(
...
+   * @param datacontenttype Type of the payload.
+   * @param data            Payload.
+   * @param pubsubname      Pubsub component name.
+   * @param topic           Topic name.
+   * @param time            Time.
+   * @param traceid         Trace id.
+   * @param traceparent     Trace parent.
+   * @param tracestate      Trace state.
+   */
+  public CloudEvent(
```
artursouza, line 180: This is a breaking change. Please, remove this constructor and leave the default. A constructor this long is not that readable either.

sdk/src/main/java/io/dapr/client/domain/CloudEvent.java
```diff
@@ -255,6 +371,104 @@ public void setBinaryData(byte[] binaryData) {
...
+  public String getTraceid() {
+    return traceid;
+  }
+
+  /**
+   * Sets the trace id which is the legacy name for trace parent.
+   * @param traceid the trace id.
+   */
+  @Deprecated
+  public void setTraceid(String traceid) {
```
artursouza, line 436: camelCase in every setter and getter too: setTraceId and getTraceId, for example.

artursouza (changes requested):
Please, sign-off to your commits so DCO can pass.

sdk/src/main/java/io/dapr/client/domain/CloudEvent.java
siebenluke: Sure, I was just trying to follow the above that didn't have camelCase:
private String specversion;
private String datacontenttype;

sdk/src/main/java/io/dapr/client/domain/CloudEvent.java
siebenluke: Sure, I will remove the 2 new constructors I added. The old default ones still exist so that shouldn't have broken anything.

siebenluke:
I will also look into how I can add a DCO sign-off to my commit(s).

sdk/src/main/java/io/dapr/client/domain/CloudEvent.java
artursouza: OK. We should fix those too.

siebenluke: I've made them camelCase and included @JsonProperty("lowercase") for each of them as the Dapr CloudEvent uses lowercasenames and JSON/OBJECT_MAPPER are case-sensitive.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
```diff
@@ -35,7 +36,8 @@ public class ObjectSerializer {
    */
   protected static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
       .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-      .setSerializationInclusion(JsonInclude.Include.NON_NULL);
+      .configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)
```
artursouza, line 39: This is a breaking change. Is this needed? I see that we already have a separate ObjectMapper for CloudEvent.

artursouza (changes requested):

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
siebenluke: I see, I just wanted to keep the OBJECT_MAPPERs in sync.

 For the CloudEvent OBJECT_MAPPER we don't need:
".configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)"
because we are just reading data and not writing it.
On the other hand the DefaultObjectSerializer is used to read/write data and if we want the output time format to match the input time format we need that setting.
ex.
input --> "timeValue":"-999999999-01-01T00:00:00+18:00"
when serialized becomes:
output --> "timeValue":-31557014135661600.000000000"
when we don't have that option configured.

In order to support OffsetDateTime for the time fields we do need:
".findAndRegisterModules()"
in CloudEvent and ObjectSerializer, otherwise we'll get an error like the following:
"com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type java.time.OffsetDateTime not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling".

In case we didn't want to modify the ObjectMapper at all I tried to use a field level annotation but couldn't get it to work.

ex.
"
// didn't deserialize with/without Z, and/or timezone among other setting combinations
@JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ssZ", timezone = "UTC")
private OffsetDateTime time;
"
or
"
// error said that this deserializer is not a constant
@JsonDeserialize(using = InstantDeserializer.OFFSET_DATE_TIME)
private OffsetDateTime time;
"

So to avoid modifying the ObjectMappers we would need to create our own CustomDeserializer ( see: https://stackoverflow.com/questions/46263773/jackson-parse-custom-offset-date-time ) and duplicate the work that the "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" JavaTimeModule gave us for free.

siebenluke: @artursouza I've removed the com.fasterxml.jackson.datatype:jackson-datatype-jsr310 dependency in favor of custom field level serdes for time.

artursouza (dismissed):

cicoyle:
@siebenluke - mind resolving the conflicts? Then I think we can merge this

siebenluke:
Hi @cicoyle / @artursouza, I have resolved the conflicts, so if someone (Artur) could take another look that would be great. Thanks.

artursouza (dismissed):

siebenluke:
Hi @cicoyle / @artursouza, I believe I've added all the necessary CloudEvent test cases to appease Codecov. Let me know if there are any other additional changes I need to make before this can be merged.

artursouza (approved):

artursouza (approved):

PR #889, opened 2023-07-13
Implement retry and timeout policy for gRPC client.

artursouza:
# Description

This is intended as a reference implementation for the proposal to support resiliency in SDKs.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/6609

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


sdk-tests/components/mongo-statestore.yml
```diff
@@ -11,4 +11,7 @@ spec:
     - name: databaseName
       value: local
     - name: collectionName
-      value: testCollection
\ No newline at end of file
+      value: testCollection
+scopes:
+  - grpcstateclientit
```
artursouza, line 16: This is to make it easier to run individual ITs without requiring MongoDB

sdk-tests/src/test/java/io/dapr/it/actors/ActorSdkResiliencytIT.java
```diff
@@ -0,0 +1,141 @@
...
+
+  @AfterClass
+  public static void tearDown() throws Exception {
+    if (toxiProxyRun != null) {
+      toxiProxyRun.stop();
+    }
+  }
+
+  @Test
+  @Ignore("Flaky when running on GitHub actions")
```
artursouza, line 109: I spent significant amount of time to remove flakiness. I did fix an issue in AbstractActor because of this test and removed flakiness running locally - not sure why it still flaky in GH Actions.

sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
```diff
@@ -31,6 +34,7 @@
 import reactor.core.publisher.MonoSink;
 import reactor.util.context.ContextView;
 
+import java.sql.Time;
```
mukundansundar, line 37: why is this needed? I am not seeing a usage in this file.

sdk/src/main/java/io/dapr/utils/Retry.java
```diff
@@ -1,49 +0,0 @@
...
-package io.dapr.utils;
-
-class Retry {
-
-  private static final long RETRY_WAIT_MILLISECONDS = 1000;
-
-  private Retry() {
-  }
-
-  static void callWithRetry(Runnable function, long retryTimeoutMilliseconds) throws InterruptedException {
```
mukundansundar, line 23: this was used in tests and not being used anymore?

mukundansundar:
LGTM just few questions.

sdk-actors/src/main/java/io/dapr/actors/client/DaprGrpcClient.java
artursouza: Left over. Removing it.

sdk/src/main/java/io/dapr/utils/Retry.java
artursouza: It was only used in the NetworkUtils. So, I moved it there. I want to avoid ambiguity into people thinking this is the RetryPolicy.

mukundansundar: isn't this a breaking change? 
Though I believe that the users have to use the same package name, to access the method.

mukundansundar (approved):
lgtm

sdk/src/main/java/io/dapr/utils/Retry.java
artursouza: Correct. Package here is used for internal scope.

Issue #897, opened 2023-08-30
[Proposal] Migrate `waitForSidecar` to `v1.0/healthz/outbound` to provide parity with dotnet SDK


olitomlinson:
## Describe the proposal

- In `WaitForSidecarAsync` in the dotnet SDK, it waits for a specific health check against `v1.0/healthz/outbound` https://github.com/dapr/dotnet-sdk/blob/17f849b17505b9a61be1e7bd3e69586718b9fdd3/src/Dapr.Client/DaprClientGrpc.cs#L1758-L1785

- In `waitForSidecar` in the Java SDK, it just waits for the dapr http port to become available https://github.com/dapr/java-sdk/blob/cf8040dd6d2e77b7e81d646e2496b871ade92fe4/sdk/src/main/java/io/dapr/client/DaprClientHttp.java#L138-L146

In my opinion, the dotnet SDK provides a more explicit implementation that is better isolated from accidental change. 

For example, if the ordering of a dapr runtime init ever changed (accidentally or deliberately), and the dapr http port was open **before** all components were initialised, then this would cause a break in behaviour.

Therefore, for this reason, I propose that the Java SDK (and other SDKs) should follow the same behaviour as the dotnet SDK and wait for a successful response from `v1.0/healthz/outbound` rather than just waiting for the dapr http port to be open.

---

This is my understanding of the relationship between Dapr and the App during init

![image](https://github.com/dapr/java-sdk/assets/4224880/9ba5af7b-b085-4ab7-86b9-6ef73a472ec4)

cicoyle:
[once the docs PR I have open gets merged](https://github.com/dapr/java-sdk/pull/981), then this issue can be closed.

Issue #898, opened 2023-09-04
Recommendation for planned deprecation of `DaprClientHttp`


olitomlinson:
Given a DaprClient created like this...

```java
 try (DaprClient client = (new DaprClientBuilder()).build()) {
      ...
 }
```

...will result in a `DaprClientGrpc` by default.

When using `waitForSidecar` method, this exhibits the same behaviour as calling `v1.0/healthz` --  as this endpoint won't return success until both the components are initialised and the app channel is established.

---

We discovered this because in our Java app we are trying to read secrets from the secret component as part of the app initialisation, therefore before the app channel becomes available.

**The impact was the app dead locked with the sidecar.** Eventually `waitForSidecar` timeout elapsed, and the App crashed as it wasn't able to read the secrets that were depended later on.

---

I did some digging in the Java SDK code, and found that I could force the `DaprClientBuilder` to return a `DaprClientHttp` by doing the following. 

```java
System.getProperties().setProperty(Properties.API_PROTOCOL.getName(), DaprApiProtocol.HTTP.name());
 try (DaprClient client = (new DaprClientBuilder()).build()) {
      ...
 }
```

In this case,  when using `waitForSidecar` method, this exhibits the same behaviour as calling `v1.0/healthz/outbound` -- this endpoint returns successfully when the components are established, **but the app channel is not yet established.** which is exactly what we needed.

The impact was that we could successfully read secrets during the init phase of our Java App, without causing a Deadlock.

Unfortunately, now this means our code depends on a deprecated capability (`DaprClientHttp`), which is planned to be removed in the next version of the Java SDK, 1.10

---

My ask is that the following changes are made **before** `DaprClientHttp` is removed from the SDK.
- Migrate `waitForSidecar` in `DaprClientGrpc` to use `v1.0/healthz/outbound`
  - This would then allow this issue to be closed without fix https://github.com/dapr/java-sdk/issues/897
- Introduce a new method called `healthCheck`,  which has the same behaviour as calling `v1.0/healthz`

This will allow a migration path for us, and also bring [parity with the dotnet SDK](https://github.com/dapr/dotnet-sdk/blob/17f849b17505b9a61be1e7bd3e69586718b9fdd3/src/Dapr.Client/DaprClientGrpc.cs#L1742-L1785)

artursouza:
I agree with this. Having parity with .Net seems the right thing hung here.

sicoyle:
[...]

PR #918, opened 2023-09-20
add health check to wait for sidecar and test it

cicoyle:
# Description

Adding a health check to the wait for sidecar logic such that there is better parity with the dotnet sdk. Updated the README. Updated the `pom.xml` to fix the err: `Caused by: java.lang.ClassNotFoundException: org.junit.jupiter.api.extension.ScriptEvaluationException`

## Issue reference

https://github.com/dapr/java-sdk/issues/897

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


salaboy:
This looks good to me, great work @cicoyle !

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -139,12 +139,22 @@ public class DaprClientHttp extends AbstractDaprClient {
...
-    return Mono.fromRunnable(() -> {
+    return Mono.defer(() -> {
       try {
+        String[] pathSegments = new String[] { DaprHttp.API_VERSION, "healthz", "outbound"};
+
+        Mono<DaprHttp.Response> responseMono = this.client.invokeApi(DaprHttp.HttpMethods.GET.name(), pathSegments,
+                null, "", null, null);
+        responseMono.block();
+
         NetworkUtils.waitForSocket(Properties.SIDECAR_IP.get(), Properties.HTTP_PORT.get(), timeoutInMilliseconds);
```
olitomlinson, line 150: `NetworkUtils.WaitForSocket(...`

I'm not a Java dev, but It's my understanding that this call should be removed? Otherwise the behaviour hasn't changed? Sorry if I've got this wrong!

olitomlinson: Also, this check for `healthz/outbound` should probably be put in a loop, and retried on non-2xx until `timeoutInMilliseconds` has elapsed?

olitomlinson: See the .NET SDK for inspiration :)

https://github.com/dapr/dotnet-sdk/blob/17f849b17505b9a61be1e7bd3e69586718b9fdd3/src/Dapr.Client/DaprClientGrpc.cs#L1758-L1785

README.md
```diff
@@ -163,13 +164,13 @@ This SDK provides a basic serialization for request/response objects but also fo
...
 
 
-### Debug Java application or Dapr's Java SDK
+### Debug a Java application or Dapr's Java SDK
 
-**In IntelliJ Community Edition, consider [debugging in IntelliJ](https://docs.dapr.io/developing-applications/ides/intellij/).**
+**In IntelliJ Community Edition, consider [debugging in IntelliJ](https://www.jetbrains.com/help/idea/debugging-your-first-java-application.html).**
 
-**In Visual Studio Code, consider [debugging in Visual Studio Code](https://docs.dapr.io/developing-applications/ides/vscode-debugging/).**
+**In Visual Studio Code, consider [debugging in Visual Studio Code](https://code.visualstudio.com/docs/editor/debugging).**
```
mukundansundar, line 171: Is there a reason to change this link?

README.md
```diff
@@ -163,13 +164,13 @@ This SDK provides a basic serialization for request/response objects but also fo
     ```
 
 
-### Debug Java application or Dapr's Java SDK
+### Debug a Java application or Dapr's Java SDK
 
-**In IntelliJ Community Edition, consider [debugging in IntelliJ](https://docs.dapr.io/developing-applications/ides/intellij/).**
+**In IntelliJ Community Edition, consider [debugging in IntelliJ](https://www.jetbrains.com/help/idea/debugging-your-first-java-application.html).**
```
mukundansundar, line 169: Is there a reason to change this link?

README.md
```diff
@@ -178,24 +179,24 @@ dapr run --app-id testapp --app-port 3000 --dapr-http-port 3500 --dapr-grpc-port
...
 DAPR_GRPC_PORT=5001
 ```
 
-Now you can go to your IDE (like Eclipse, for example) and debug your Java application, using port `3500` to call Dapr while also listening to port `3000` to expose Dapr's callback endpoint.
+Now you can go to your IDE and debug your Java application, using port `3500` to call Dapr while also listening to port `3000` to expose Dapr's callback endpoint.
 
 ### Exception handling
 
-Most exceptions thrown from the SDK are instances of `DaprException`. `DaprException` extends from `RuntimeException`, making it compatible with Project Reactor. See [example](./examples/src/main/java/io/dapr/examples/exception) for more details.
+Most exceptions thrown from the SDK are instances of `DaprException`. `DaprException` extended from `RuntimeException`, making it compatible with Project Reactor. See the [exception example](./examples/src/main/java/io/dapr/examples/exception) for more details.
```
mukundansundar, line 192: Extends seems the right word to use here as the exceptions still extend the class RuntimeException

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
mukundansundar: Agreed this is not needed.

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
```diff
@@ -139,12 +139,22 @@ public class DaprClientHttp extends AbstractDaprClient {
...
       } catch (InterruptedException e) {
         throw new RuntimeException(e);
+      } catch (DaprException e) {
+        throw new IllegalStateException(String.format("Components aren't initialized"));
```
mukundansundar, line 154: Can this still be RuntimeException(e) as above? Otherwise the exception is being swallowed here.

sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
```diff
@@ -100,17 +101,45 @@ public void setUp() {
...
+
     assertThrows(RuntimeException.class, () -> daprClientHttp.waitForSidecar(1).block());
   }
 
   @Test
   public void waitForSidecarTimeoutOK() throws Exception {
+    mockInterceptor.addRule()
+            .get()
+            .path("/v1.0/healthz/outbound")
+            .respond(200);
```
mukundansundar, line 142: Nitpick : I think this is 204 from dapr

mukundansundar (changes requested):

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
cicoyle: yup, I can change that

README.md
cicoyle: Oh, yeah, actually I see that now. I'll change that back

README.md
cicoyle: that original link is a 404: Not Found

README.md
cicoyle: that original link is a 404: Not Found

sdk/src/main/java/io/dapr/client/DaprClientHttp.java
cicoyle: I'll rm it and add the retry logic.

olitomlinson:
@cicoyle @mukundansundar 

If you haven't already, It's worth checking in with @artursouza as this class `DaprClientHttp` is planned to be deprecated in the next version of the SDK.

If that is still true, these changes also need porting over to `DaprClientGrpc` - as this issue suggests https://github.com/dapr/java-sdk/issues/898

mukundansundar:
> @cicoyle @mukundansundar
> 
> If you haven't already, It's worth checking in with @artursouza as this class `DaprClientHttp` is planned to be deprecated in the next version of the SDK.
> 
> If that is still true, these changes also need porting over to `DaprClientGrpc` - as this issue suggests #898

Good point @olitomlinson, also the one issue with the gRPC waitForSidecar function is that it only waits for the channel to be ready, not for Dapr outbound to be ready. Either as done in .NET SDK as [linked above](https://github.com/dapr/java-sdk/pull/918#discussion_r1334722326) or as mentioned in the Python SDK [issue](https://github.com/dapr/python-sdk/issues/611#issuecomment-1745607578), we either need gRPC client to also check the HTTP `healthz/outbound` endpoint or create a `IsOutboundReady` gRPC method in dapr Runtime.

olitomlinson:
> her need gRPC client to also check t



> > @cicoyle @mukundansundar
> > If you haven't already, It's worth checking in with @artursouza as this class `DaprClientHttp` is planned to be deprecated in the next version of the SDK.
> > If that is still true, these changes also need porting over to `DaprClientGrpc` - as this issue suggests #898
> 
> Good point @olitomlinson! One issue with the gRPC `waitForSidecar` function is that it only waits for the channel to be ready, not for Dapr outbound to be ready. Either as done in .NET SDK as [linked above](https://github.com/dapr/java-sdk/pull/918#discussion_r1334722326) or as mentioned in the Python SDK [issue](https://github.com/dapr/python-sdk/issues/611#issuecomment-1745607578), we either need gRPC client to also check the HTTP `healthz/outbound` endpoint or create a `IsOutboundReady` gRPC method in dapr Runtime.

Not that I have any say in the matter, but I think being consistent with the dotnet SDK and calling `healthz/outbound` is the pragmatic thing to do right now :)

README.md
rob-dotmatics: The original page appears to have moved to https://docs.dapr.io/developing-applications/local-development/ides/intellij/

It would be useful to include this link rather than the generic Jetbrains page, as there is specific information for setting up a run/debug configuration for DAPR.

README.md
rob-dotmatics: https://docs.dapr.io/developing-applications/local-development/ides/vscode/

cicoyle:
Unfortunately, this works locally on my Mac, but not in CI - trying to debug

artursouza:
I reproduced this error locally.

The error is because the test is not passing the interceptor to an OkHTTP client. You need this in GrpcChannelFacadeTest:

```
    okHttpClient = new OkHttpClient.Builder().addInterceptor(mockInterceptor).build();
```

And then, use the custom okHttpClient:

```
DaprHttp daprHttp = new DaprHttp(Properties.SIDECAR_IP.get(), 3500, okHttpClient);
```

It might be working locally because you might have a running sidecar and makes the test pass.

sdk/src/main/java/io/dapr/client/GrpcChannelFacade.java
```diff
@@ -43,6 +64,8 @@ class GrpcChannelFacade implements Closeable {
    */
   GrpcChannelFacade(ManagedChannel channel) {
     this.channel = channel;
+    OkHttpClient okHttpClient = new OkHttpClient.Builder().build();
+    this.daprHttp = new DaprHttp(Properties.SIDECAR_IP.get(), 3500, okHttpClient);
```
artursouza, line 68: This should not be forcing port 3500 since the sidecar can run on any port.

artursouza (approved):

Issue #931, opened 2023-09-27
Provide getGuid() method in java workflow SDK


skyao:
## Describe the proposal

As required by documentation "Workflow functions must call deterministic APIs"

https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-features-concepts/#workflow-functions-must-call-deterministic-apis

```cshap
// DON'T DO THIS!
DateTime currentTime = DateTime.UtcNow;
Guid newIdentifier = Guid.NewGuid();
string randomString = GetRandomString();
```

```cshap
DateTime currentTime = context.CurrentUtcDateTime;
Guid newIdentifier = context.NewGuid();
string randomString = await context.CallActivityAsync<string>("GetRandomString");
```

Now in java workflow SDK, we don't provide such getGuid() method (yeat context.getCurrentInstant() method is provided) .

We need to improve our java workflow sdk to provide this getGuid() method. 

And this getGuid() method is NOT provided by durable task java SDK, so we need to update durable task java SDK first.

This improvement should be included in next release.

skyao:
https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskOrchestrationContext.cs

```cshap
    /// <summary>
    /// Creates a new GUID that is safe for replay within an orchestration or operation.
    /// </summary>
    /// <remarks>
    /// The default implementation of this method creates a name-based UUID V5 using the algorithm from RFC 4122 §4.3.
    /// The name input used to generate this value is a combination of the orchestration instance ID, the current time,
    /// and an internally managed sequence number.
    /// </remarks>
    /// <returns>The new <see cref="Guid"/> value.</returns>
    public abstract Guid NewGuid();
```

Issue #944, opened 2023-10-31
Probably rewrite workflow example for better understanding


kaibocai:
## Describe the proposal
I am reading the workflow example and find it a little bit messy due to combining all APIs in a single workflow example. I was wondering should we separate the single huge workflow into several examples, each giving a more concrete example of a set of APIs. This may help user to better understand it. 

Also `continueAsNew` API has a break change with the release here https://github.com/microsoft/durabletask-java/releases/tag/1.2.0. It will not continue executing logic after it (an exception is thrown out in the durabletask SDK to force yield control back to the sidecar). So example here https://github.com/dapr/java-sdk/blob/0d0d59946f81a20cb17d8f7d3d794969878e9e45/examples/src/main/java/io/dapr/examples/workflows/DemoWorkflow.java#L109-L121 will never be triggered once upgrade durabletask-java to v1.2.0.

rabollin:
Related Issue: https://github.com/dapr/java-sdk/issues/911

cicoyle:
I think that would be a valuable change - breaking up and making each example more explicit and complete. Is this something you would be able to and want to work on?

kaibocai:
Yeah, definitely. I can work on improving the workflow example.

PR #947, opened 2023-11-03
Add determinstic UUID generation

kaibocai:
# Description

This PR add logic to generated a determinstic UUID to for workflow. 
```java
    UUID uuid = context.newUuid();
```

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/931

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk-workflows/pom.xml
+++ b/sdk-workflows/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>com.microsoft</groupId>
       <artifactId>durabletask-client</artifactId>
-      <version>1.1.1</version>
+      <version>1.5.0</version>
     </dependency>
     <!--
     manually declare durabletask-client's jackson dependencies
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <grpc.version>1.42.1</grpc.version>
+    <grpc.version>1.59.0</grpc.version>
     <protobuf.version>3.17.3</protobuf.version>
     <dapr.proto.baseurl>https://raw.githubusercontent.com/dapr/dapr/v1.12.0-rc.2/dapr/proto</dapr.proto.baseurl>
     <dapr.sdk-workflows.version>0.11.0-SNAPSHOT</dapr.sdk-workflows.version>
--- a/sdk-actors/pom.xml
+++ b/sdk-actors/pom.xml
@@ -18,7 +18,7 @@
 
   <properties>
     <maven.deploy.skip>false</maven.deploy.skip>
-    <grpc.version>1.42.1</grpc.version>
+    <grpc.version>1.59.0</grpc.version>
   </properties>
 
   <dependencies>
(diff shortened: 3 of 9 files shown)


artursouza (dismissed):

sdk-workflows/pom.xml
```diff
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>com.microsoft</groupId>
       <artifactId>durabletask-client</artifactId>
-      <version>1.1.1</version>
+      <version>1.5.0</version>
```
cicoyle, line 54: does upgrading to this version enable us to skip over the breaking change mentioned here https://github.com/dapr/java-sdk/issues/944?

kaibocai: The break change will be introduced with this upgrade. 
I would be able to help work on the workflow example. 

The reason we introduce this change to `continueAsNew` is that we hear many customers say it's confusing as they expect it will immediately restart the orchestration instead of continuing to run the code after it. Just like the `continue` in a for loop, no one expects code after `continue` to be run in current loop. 
Hopefully, this won't raise any concerns, but let me know if we want to approach this differently.

cicoyle: I can see why that would cause confusion... Will we need to update our workflow examples and code as a result of this change in behavior, as a result of the upgrade? If yes, I vote to pull the upgrade out of this PR to a new one with the relevant changes, if not, then it's probably fine as is.

kaibocai: The deterministic UUID feature is added to `durabletask-client` in this version v1.5.0, so it's needed for this PR. I can try to create another PR to update the workflow example, at least make it work with this breaking change. Or do you think including the example update in this PR is better? 

Besides the example update, I don't think there is any other code change required, since we are calling `continueAsNew ` API of the `durabletask-client` at the end. But please let me know if I miss something.

cicoyle: oh I see, I think the example can be updated in a follow up PR. @artursouza do you agree?

artursouza: Waiting on this to merge.

kaibocai: I'm a little bit confused about the CI error, I don't think this change is related to those errors. Also for some jdk version it does pass. Could they be intermittent errors? Is there something I missed? Thanks.

kaibocai:
Hi @cicoyle , would you mind helping me to rerun the CI, I added one more test to improve the coverage.
Also, is there a way to gain permission to run the CI by myself? I read the CONTRIBUTING.md but didn't find anything about it. Thanks.

cicoyle:
> Hi @cicoyle , would you mind helping me to rerun the CI, I added one more test to improve the coverage. Also, is there a way to gain permission to run the CI by myself? I read the CONTRIBUTING.md but didn't find anything about it. Thanks.

Hey @kaibocai, the Java-sdk maintainers are the only ones able to run CI and my understanding is that there is flakiness. However, I can ping Artur for re-running CI. I believe he won't merge this until there is a PR updating the example

sdk-workflows/pom.xml
artursouza: Merging this after tests pass.

kaibocai: I think tests passed.

PR #956, opened 2023-11-16
Add Saga pattern support in java-sdk

skyao:
# Description

This PR is to add saga pattern support in Dapr java sdk.

The implementation of saga pattern is based on java workflow sdk which has been released in lasted v1.10.0 version. 

At this time, This PR is a preview edition: 

- Saga pattern support is still a proposal and under review and discussion.
-  This PR is to help to understand how saga works, and we provide [a saga quickstart](https://github.com/dapr/quickstarts/pull/957) as demo
- This PR is not a final version and will keep updating during the proposal review and discussion
- Don't merge this PR before the proposal is accepted and the design is finalized.

To run the saga quickstart, please reference to:

https://github.com/dapr/quickstarts/pull/957

## Issue reference

https://github.com/dapr/java-sdk/issues/968

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk-workflows/src/main/java/io/dapr/workflows/Workflow.java
```diff
@@ -30,10 +34,49 @@ public Workflow(){
...
+    WorkflowStub stub = this.create();
+
+    if (!this.isSagaEnabled()) {
+      // saga disabled
+      stub.run(ctx);
+    } else {
+      // saga enabled
+      try {
+        stub.run(ctx);
+      } catch (OrchestratorBlockedException e) {
```
kaibocai, line 51: Besides `OrchestratorBlockedException`, there is another exception [`ContinueAsNewInterruption `](https://github.com/microsoft/durabletask-java/blob/main/client/src/main/java/com/microsoft/durabletask/interruption/ContinueAsNewInterruption.java) that used as an interruption when customer call `continueAsNew`, it interrupts the workflow and yields control back to the workflow Grpc server. Should we also re-throw it here? 

I think it will be a good test case to add for combining `continueAsNew` and Saga option.

sdk-workflows/src/main/java/io/dapr/workflows/DaprWorkflowContextImpl.java
```diff
@@ -51,6 +53,23 @@ public DaprWorkflowContextImpl(TaskOrchestrationContext context) throws IllegalA
    * @throws IllegalArgumentException if context or logger is null
    */
   public DaprWorkflowContextImpl(TaskOrchestrationContext context, Logger logger) throws IllegalArgumentException {
+    this(context, logger, null);
+  }
+
+  public DaprWorkflowContextImpl(TaskOrchestrationContext context, Saga saga) throws IllegalArgumentException {
```
artursouza, line 59: Can the Saga pattern be separate classes that takes in the workflow as params in the constructor? In this case, it would be DaprWorkflowSagaContextImpl.

cc @cgillum

sdk-workflows/src/main/java/io/dapr/workflows/Workflow.java
skyao: Thanks for your reminder, this exception must be handled the same way. I will update the proposal and code.

sdk-workflows/src/main/java/io/dapr/workflows/DaprWorkflowContextImpl.java
skyao: PR code updated. Add SagaContext and getSagaContext() in WorkflowContext.

artursouza (approved):

Issue #962, opened 2023-11-23
Dapr API Token not added for Durable Task Worker in Workflows SDK


salaboy:
## Expected Behavior

I expect both the Workflow APIs and the Durable Tasks workers to forward the Dapr API Token for each request. 

## Actual Behavior

Only the workflow APIs are setting the token using an interceptor but not the Durable Tasks workers. 
https://github.com/dapr/java-sdk/blob/e4cb03f9ec7fd652159b35cbe738e649a63dd153/sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java#L51

This should be also used here: 

https://github.com/dapr/java-sdk/blob/e4cb03f9ec7fd652159b35cbe738e649a63dd153/sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntimeBuilder.java#L25

## Steps to Reproduce the Problem

I couldn't find any tests to show that the interceptor is working for -> https://github.com/dapr/java-sdk/blob/e4cb03f9ec7fd652159b35cbe738e649a63dd153/sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java#L51 

But it was working when I tested it. 
Then the worker fails to connect to the sidecar

## Release Note

RELEASE NOTE:

PR #963, opened 2023-11-23
Adding dapr api token interceptor for durable task

salaboy:
# Description

This PR adds the interceptor to inject the Dapr API Token to the Durable Tasks Worker, which is needed to make sure that the token gets sent by the workflows SDK. 

## Issue reference

Please reference the issue this PR will close: #962

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java
```diff
@@ -262,7 +262,7 @@ public void close() throws InterruptedException {
     }
   }
 
-  private static ClientInterceptor WORKFLOW_INTERCEPTOR = new ClientInterceptor() {
+  public static ClientInterceptor WORKFLOW_INTERCEPTOR = new ClientInterceptor() {
```
artursouza, line 265: Is there a way to avoid making this public? This makes this part of the SDK user facing surface with this change.

artursouza (changes requested):

sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java
salaboy: Because this is an interceptor, I believe it is ok to make it public and part of the interface, as this functionality can apply to other application code where the Dapr Token might want to be added.. does that make sense @artursouza ?

artursouza: I propose we create a new class, called WorkflowGrpcInterceptor (or something similar) in a packaged called io.dapr.workflows.internal because that communicate that anything in that package is for internal use and is not part of our interface with the application.

So, we can use the same class in both places here but just will not be a shared instance.

salaboy: done

artursouza (approved):

Issue #971, opened 2023-12-13
[Proposal]Support JDK17 and springboot3 in Dapr Java SDK


skyao:
* Author(s): Sky Ao (@skyao)
* State: Ready for Implementation
* Updated: 2024-01-08

## Overview

This document is a design proposal to support JDK17 and springboot 3.0 in Dapr Java SDK.

This proposal will allow customers to use the Dapr Java SDK with JDK17 and springboot3.0 while trying to maintain compatibility with existing implementations: continue to support JDK8/JDK11, continue to support springboot2.x.

### Current Versions Overview

In this proposal, I will refer to the following four subprojects by "SDK Core" for convenience:

- sdk-autogen
- sdk
- sdk-actors
- sdk-workflows

The following picture shows the overview of current JDK versions and springboot versions in the Dapr Java SDK:

![jdk-springboot-versions-now](https://github.com/dapr/java-sdk/assets/1582369/88266f23-6c1f-4f42-ad8e-a338e1f194ec)

> In this picture, we've ignored third-party dependencies other than springboot, as well as the dependencies of the internal subprojects examples and sdk-tests on sdk-workflows/sdk-actors/sdk subprojects.

I will describe the details of this overview picture.

#### JDK versions in java-sdk now

Currently, the Dapr Java SDK is compiled with JDK11, but for Java8 compatibility, set both the compiler source and target to 8:  

```xml
<!-- pom.xml in root -->  
<properties>
    <maven.compiler.source>8</maven.compiler.source>
    <maven.compiler.target>8</maven.compiler.target>
</properties>
```

The examples and sdk-tests subprojects have no need for Java8 compatibility because they do not need to be published to customers, so their source and target are set to 11:

```xml
<!-- pom.xml in sdk-tests and examples -->   
<properties>
    <maven.compiler.source>11</maven.compiler.source>
    <maven.compiler.target>11</maven.compiler.target>
</properties>
```

Therefore, the current Java versions supported in Dapr Java SDK  are Java8 and Java11:

![java-version-in-sdk-core](https://github.com/dapr/java-sdk/assets/1582369/00c0690b-d122-4411-88b0-ee6bede04084)

> In this proposal, We focus only on the LTS(Long-Time-Support) versions of Java: Java8/Java11/Java17 and possibly Java21 in the future. 

#### Springboot versions in java-sdk now

Currently, Dapr's springboot support is integrated with springboot 2.x. The specific version is 2.7.8, which is set in the pom.xml:

```xml
[...]

PR #975, opened 2023-12-16
Add JDK17 and springboot 3.0 support 

skyao:
# Description

Upgrade to add springboot 3.0 support in dapr java sdk.

For the details, please reference to the proposal: https://github.com/dapr/java-sdk/issues/971 

## Issue reference

https://github.com/dapr/java-sdk/issues/971

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


.github/workflows/build.yml
```diff
@@ -21,22 +21,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11, 13, 15, 16 ]
```
mukundansundar, line 24: Do not remove 11 as yet. Have that also

mukundansundar (changes requested):

.github/workflows/build.yml
skyao: I reference to the previous design as:

```toml
      matrix:
        java: [ 11, 13, 15, 16 ]
        spring-boot-version: [ 2.7.8 ]
        experimental: [ false ]
        include:
          - java: 11
            spring-boot-version: 2.6.14
            experimental: false
          - java: 11
            spring-boot-version: 2.5.7
            experimental: false
          - java: 11
            spring-boot-version: 2.4.0
            experimental: false
          - java: 11
            spring-boot-version: 2.3.6.RELEASE
            experimental: false
```

Through java8 is supported, sdk-tests subproject used jdk11 and set compile source and target to jdk11, and test with springboot 2.x versions with JDK11 as list above.

skyao: If a jdk version below jdk17 appears here, it means that no jdk17 features can be used in the sdk-tests sub-project, or it will fail to compile.

This means that the sdk-tests project has to be split into two sub-projects, jdk17 and jdk11, and the various test cases have to be rewritten in both sub-projects. Otherwise it would not be covered by both jdk17 and jdk11.

skyao: That's why it costs me so much to keep the sdk-tests subproject from being split.

mukundansundar: @skyao For now we are not using any jdk17 features in the idk-tests sub-project right? 
And as long as we support jdk8 and jdk11, won't that continue?

sdk-tests/pom.xml
```diff
@@ -12,16 +12,16 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <maven.compiler.source>11</maven.compiler.source>
-    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
```
mukundansundar, line 16: why is the target changing here?

examples/pom.xml
```diff
@@ -18,12 +18,13 @@
   <properties>
     <protobuf.output.directory>${project.build.directory}/generated-sources</protobuf.output.directory>
     <protobuf.input.directory>${project.basedir}/proto</protobuf.input.directory>
-    <java.version>11</java.version>
+    <java.version>17</java.version>
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
```
mukundansundar, line 23: can the target still be 8/11? why does that need to change?

skyao: As we discussed, we will use jdk17 and springboot3 for examples and sdk-tests subprojects.

spotbugs-exclude.xml
```diff
@@ -3,4 +3,34 @@
...
+    <Match>
+        <Package name="~io\.dapr.*"/>
+        <Bug pattern="EI_EXPOSE_REP"/>
+    </Match>
+
+    <!--Temporarily ignoring checking after upgrade to new spotbugs version-->
+    <Match>
+        <Package name="~io\.dapr.*"/>
+        <Bug pattern="EI_EXPOSE_REP2"/>
+    </Match>
```
mukundansundar, line 35: is there a reason why we need this?

mukundansundar (dismissed):
lgtm. waiting for tests to pass

sdk-springboot30/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.dapr</groupId>
+    <artifactId>dapr-sdk-parent</artifactId>
+    <version>1.11.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>dapr-sdk-springboot30</artifactId>
```
artursouza, line 13: Just dapr-sdk-springboot3, please.

sdk-tests/src/test/java/io/dapr/it/tracing/http/OpenTelemetryInterceptor.java
```diff
@@ -16,15 +16,14 @@
...
-import javax.annotation.Nullable;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.springframework.web.servlet.HandlerInterceptor;
 import org.springframework.web.servlet.ModelAndView;
 
-import javax.servlet.DispatcherType;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
+import jakarta.servlet.DispatcherType;
```
artursouza, line 24: Why not javax? Is it required because of Java17?

artursouza (changes requested):

artursouza:
Validation of examples is failing.

spotbugs-exclude.xml
skyao: https://github.com/dapr/java-sdk/issues/971

See “Spotbugs bug pattern” section in Appendices  part.

sdk-tests/src/test/java/io/dapr/it/tracing/http/OpenTelemetryInterceptor.java
skyao: Yes, Java 17 removed `javax.*` and required to use `jakarta.*`

sdk-springboot30/pom.xml
skyao: OK. updated!

skyao:
I have confused by the behavior of NOP Logger and backlog. 

Before this PR, examples subproject in java-sdk repo depends on springboot2.x and introduced logback as its slf4j implementation:

```bash
[INFO] +- org.springframework.boot:spring-boot-starter-web:jar:2.7.8:compile
[INFO] |  +- org.springframework.boot:spring-boot-starter:jar:2.7.8:compile
[INFO] |  |  +- org.springframework.boot:spring-boot-starter-logging:jar:2.7.8:compile
[INFO] |  |  |  +- ch.qos.logback:logback-classic:jar:1.2.11:compile
[INFO] |  |  |  |  \- ch.qos.logback:logback-core:jar:1.2.11:compile
```

In this PR, I updated the version of springboot to springboot 3.0, also introduced logback as its slf4j implementation:

```bash
[INFO] +- org.springframework.boot:spring-boot-starter-web:jar:3.0.13:compile
[INFO] |  +- org.springframework.boot:spring-boot-starter:jar:3.0.13:compile
[INFO] |  |  +- org.springframework.boot:spring-boot-starter-logging:jar:3.0.13:compile
[INFO] |  |  |  +- ch.qos.logback:logback-classic:jar:1.4.11:compile
[INFO] |  |  |  |  \- ch.qos.logback:logback-core:jar:1.4.11:compile
```

But now the github action failed because of the conflicts between logback and NOP logger from slf4j-api: 

```bash
		== APP == 	... 8 more
		== APP == Caused by: java.lang.IllegalArgumentException: LoggerFactory is not a Logback LoggerContext but Logback is on the classpath. Either remove Logback or the competing implementation (class org.slf4j.helpers.NOPLoggerFactory loaded from jar:file:/home/runner/work/java-sdk/java-sdk/examples/target/dapr-java-sdk-examples-exec.jar!/BOOT-INF/lib/slf4j-api-1.7.25.jar!/). If you are using WebLogic you will need to add 'org.slf4j' to prefer-application-packages in WEB-INF/weblogic.xml: org.slf4j.helpers.NOPLoggerFactory
```

We used NOP logger in workflow, in DaprWorkflowContextImpl: 

```java
  public Logger getLogger() {
    if (this.innerContext.getIsReplaying()) {
      return NOPLogger.NOP_LOGGER;
    }
    return this.logger;
  }
```

The problem is that it works with jdk11 + springboot2.x (before this PR), but it doesn't work with jdk17 + springboot3.0. 

I tried to exclude logback from examples subproject to fix above conflicts, then all the examples will pass exception workflow example: withtout logback, only NOP logger it can't output the text for validation.

```bash
== APP == SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
== APP == SLF4J: Defaulting to no-operation (NOP) logger implementation
== APP == SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

I also checked other examples to see why they didn't fail without logback, there are 

- using system out
- or using java.util.logging.Logger

skyao:
Done!

```java
   <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>slf4j-api</artifactId>
        <version>2.0.9</version>
      </dependency>
```

After adding slf4j-api  as compile dependency and upgrade its version to 2.0.9, then logback and NOP logger can work together.

@artursouza @mukundansundar This PR can be merged now!

artursouza (approved):

PR #978, opened 2023-12-26
Improving and Adding Logs for Workflow

RyanLettieri:
# Description

Added in logging statements for dapr workflow and optional parameter for setting the log level.

*Note that the failures are due to every individual case of a large switch-case statement not being tested*

## Issue reference

Please reference the issue this PR will close: #946 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


mukundansundar:
@RyanLettieri Can you look into the coverage checks that are failing in this PR?

RyanLettieri:
> @RyanLettieri Can you look into the coverage checks that are failing in this PR?

The coverage checks are due to the switch case statement for the log levels. We log what the log level was set to in the switch case statements, I.e. "Severe", "Warning", etc. I didn't believe it made much sense to create test cases for each individual log level but if needed I certainly can.

sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntimeBuilder.java
```diff
@@ -70,5 +90,77 @@ public <T extends WorkflowActivity> void registerActivity(Class<T> clazz) {
...
     );
+    String logTime = getLogTime();
+    this.logger.log(Level.INFO, logTime + " Registered Activity: " +  clazz.getSimpleName());
+    this.activities.add(clazz.getSimpleName());
+  }
+
+  /**
+   * Sets the log level for the workflow runtime builder.
+   */
+  private void setLogLevel() {
```
artursouza, line 101: In Java, log level and date format is done from the application's perspective. See the example here: https://github.com/dapr/java-sdk/blob/master/examples/src/main/resources/logback.xml

If you need this from the test's perspective, you can modify this class to have a private constructor that will receive the custom logger from the test.

artursouza (changes requested):

artursouza (approved):

PR #998, opened 2024-02-01
Add the dapr runtime returned error details to the Java DaprException

cicoyle:
# Description

Add the error details from dapr runtime to the java dapr exception, such that users may parse for error details in their code accordingly.

Here is the output from running the PubSub Client code:
image

## Issue reference

https://github.com/dapr/docs/issues/3907#issuecomment-1908335376

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests -> added a test for 1 & more than 1 error details


msfussell:
@cicoyle - Are you doing this for v1.13?

examples/src/main/java/io/dapr/examples/exception/Client.java
```diff
@@ -35,11 +40,39 @@ public static void main(String[] args) throws Exception {
...
       try {
-        client.getState("Unknown state store", "myKey", String.class).block();
+        client.publishEvent("", "", "").block();
+        // client.publishEvent("fake", "fake", "someData").block();
+        // client.getState("Unknown state store", "myKey", String.class).block();
       } catch (DaprException exception) {
         System.out.println("Error code: " + exception.getErrorCode());
         System.out.println("Error message: " + exception.getMessage());
 
+        try {
```
artursouza, line 50: This is too complex for users. Can we have an opinionated structure that encapsulates this into the DaprException object?

artursouza (changes requested):

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -73,6 +74,11 @@ public class DaprHttp implements AutoCloseable {
   private static final Set<String> ALLOWED_CONTEXT_IN_HEADERS =
       Collections.unmodifiableSet(new HashSet<>(Arrays.asList("grpc-trace-bin", "traceparent", "tracestate")));
 
+  /**
+   * Object mapper to parse DaprError with or without details.
+   */
+  private static final ObjectMapper DAPR_ERROR_DETAILS_OBJECT_MAPPER = new ObjectMapper();
```
artursouza, line 80: No need for a custom ObjectMapper. The stock one is enough.

sdk/src/main/java/io/dapr/exceptions/DaprErrorDetails.java
```diff
@@ -0,0 +1,223 @@
...
+   * @return A Map representing the fields of the proto message.
+   */
+  private static Map<String, Object> messageToMap(Message message) {
+    Map<String, Object> result = new HashMap<>();
+    Field[] fields = message.getClass().getDeclaredFields();
+
+    result.put("@type", generateErrorTypeFqn(message.getClass()));
+
+    for (Field field : fields) {
+      if (field.isSynthetic() || Modifier.isStatic(field.getModifiers())) {
```
artursouza, line 163: Clarifying my change: This avoids the need to check for _ in the name.

sdk/src/main/java/io/dapr/exceptions/DaprErrorDetails.java
```diff
@@ -0,0 +1,223 @@
...
+        continue;
+      }
+
+      String normalizedFieldName = field.getName().replaceAll("_$", "");
```
artursouza, line 167: Since the code still contains _, we need to remove it. If the library decided to stop appending the _ to the attribute names, we would not need to change this code.

sdk/src/main/java/io/dapr/exceptions/DaprErrorDetails.java
```diff
@@ -0,0 +1,223 @@
...
+        result.put(normalizedFieldName, value);
+      } catch (IllegalAccessException e) {
+        // no-op, just ignore this attribute.
+      }
+    }
+
+    return Collections.unmodifiableMap(result);
+  }
+
+  private static <T extends com.google.protobuf.Message> String generateErrorTypeFqn(Class<T> clazz) {
```
artursouza, line 180: Clarifying my change: just a small refactoring to allow this logic to be reused.

sdk/src/main/java/io/dapr/exceptions/DaprErrorDetails.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.stream.Collectors;
+
+public class DaprErrorDetails {
+
+  static final DaprErrorDetails EMPTY_INSTANCE = new DaprErrorDetails((Status) null);
+
+  private static final Map<Class<? extends Message>, ErrorDetailType> SUPPORTED_ERROR_TYPES =
+      Collections.unmodifiableMap(new HashMap<>() {
+        {
+          put(com.google.rpc.ErrorInfo.class, ErrorDetailType.ERROR_INFO);
```
artursouza, line 38: Clarifying my change: avoids exposing 3rd-party library types to the app.

sdk/src/main/java/io/dapr/exceptions/DaprErrorDetails.java
```diff
@@ -0,0 +1,223 @@
...
+          put(com.google.rpc.PreconditionFailure.class, ErrorDetailType.PRECONDITION_FAILURE);
+          put(com.google.rpc.BadRequest.class, ErrorDetailType.BAD_REQUEST);
+          put(com.google.rpc.RequestInfo.class, ErrorDetailType.REQUEST_INFO);
+          put(com.google.rpc.ResourceInfo.class, ErrorDetailType.RESOURCE_INFO);
+          put(com.google.rpc.Help.class, ErrorDetailType.HELP);
+          put(com.google.rpc.LocalizedMessage.class, ErrorDetailType.LOCALIZED_MESSAGE);
+        }
+      });
+
+  private static final Map<String, Class<? extends Message>> ERROR_TYPES_FQN_REVERSE_LOOKUP =
```
artursouza, line 51: Clarifying my change: handy to map a string to a class when parsing JSON error response in HTTP API.

examples/src/main/java/io/dapr/examples/exception/Client.java
```diff
@@ -33,17 +40,17 @@ public class Client {
    */
   public static void main(String[] args) throws Exception {
     try (DaprClient client = new DaprClientBuilder().build()) {
-
       try {
-        client.getState("Unknown state store", "myKey", String.class).block();
+        client.publishEvent("unknown_pubsub", "mytopic", "mydata").block();
```
artursouza, line 44: Clarifying my change: keeping this example simple.

examples/src/main/java/io/dapr/examples/exception/Client.java
```diff
@@ -33,17 +40,17 @@ public class Client {
...
       } catch (DaprException exception) {
-        System.out.println("Error code: " + exception.getErrorCode());
-        System.out.println("Error message: " + exception.getMessage());
-
-        exception.printStackTrace();
+        System.out.println("Dapr exception's error code: " + exception.getErrorCode());
+        System.out.println("Dapr exception's message: " + exception.getMessage());
+        System.out.println("Dapr exception's reason: " + exception.getStatusDetails().get(
```
artursouza, line 48: Clarifying my change: this is the opinionated layer in the SDK and determines how users can retrieve error details. We don't enumerate all the attributes on purpose. We can decide to change this later but I think this is a good first design to give a semi-structured way to fetch error details.

examples/src/main/java/io/dapr/examples/exception/Client.java
artursouza: See the latest version with the "opinionated" layer I added.

cicoyle (commented):
Changes LGTM Thx! @artursouza

artursouza (approved):

PR #1006, opened 2024-02-07
Fix unit tests hanging

artursouza:
# Description

Fix unit tests hanging.
Other improvements to workflow and unit tests.

## Issue reference

Please reference the issue this PR will close: n/a

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
+++ b/sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
@@ -107,22 +107,17 @@ public class DaprClientHttpTest {
 
   @Test
   public void waitForSidecarTimeOutHealthCheck() throws Exception {
-    VirtualTimeScheduler virtualTimeScheduler = VirtualTimeScheduler.getOrSet();
-    StepVerifier.setDefaultTimeout(Duration.ofSeconds(20));
-
-    int port = findFreePort();
-    System.setProperty(Properties.HTTP_PORT.getName(), Integer.toString(port));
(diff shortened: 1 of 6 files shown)


sdk/src/test/java/io/dapr/client/DaprClientHttpTest.java
```diff
@@ -107,79 +107,54 @@ public void setUp() {
 
   @Test
   public void waitForSidecarTimeOutHealthCheck() throws Exception {
```
artursouza, line 109: The root cause was global changes performed by these test cases in DaprClientHttpTest and it affected RetryPolicyTest run if it happens after these.

sdk/src/test/java/io/dapr/resiliency/RetryPolicyTest.java
```diff
@@ -19,14 +19,11 @@
...
 
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertSame;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 
 public class RetryPolicyTest {
```
artursouza, line 28: Just did some clean up by using StepVerifier instead - it seems to be better practice but did not affect the fix. On the other hand, the hang reproduces as a timeout and test failure instead of hanging forever.

mukundansundar (approved):

PR #1051, opened 2024-06-15
Remove HTTP client and add gRPC interceptor helper.

artursouza:
# Description

- Removes HTTP client implementation.
- Client is now a mesh of mostly the gRPC client with HTTP for methodInvoke() and waitForSidecar().
- Adds an interceptor helper class for gRPC proxy invoke.
- Adds a method to get the gRPC channel for gRPC proxy invoke.

## Issue reference

Please reference the issue this PR will close: TBD

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


cicoyle (commented):
Its a pretty big diff, but looks good to me so far 👍🏻

examples/src/main/java/io/dapr/examples/configuration/README.md
```diff
@@ -158,7 +158,7 @@ sleep: 10
 -->
 
 ```bash
-dapr run --components-path ./components/configuration --app-id configgrpc --log-level debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.configuration.grpc.ConfigurationClient
```
artursouza, line 161: Configuration API is gRPC only now.

examples/src/main/java/io/dapr/examples/invoke/grpc/HelloWorldClient.java
```diff
@@ -45,22 +44,11 @@ public class HelloWorldClient {
...
-      Metadata headers = new Metadata();
-      headers.put(Metadata.Key.of("dapr-app-id", Metadata.ASCII_STRING_MARSHALLER),
-          "hellogrpc");
-
-      // MetadataUtils.attachHeaders is deprecated.
-      blockingStub = blockingStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(headers));
+    try (DaprClient client = new DaprClientBuilder().build()) {
+      HelloWorldGrpc.HelloWorldBlockingStub blockingStub = HelloWorldGrpc.newBlockingStub(client.getGrpcChannel());
+      // Adds Dapr interceptors to populate gRPC metadata automatically.
+      blockingStub = DaprClientGrpcInterceptors.intercept("hellogrpc", blockingStub);
```
artursouza, line 51: This is the new way to invoke gRPC, now using the gRPC proxy and with the helper methods: `client.getGrpcChannel()` and `DaprClientGrpcInterceptors.intercept(appId, stub);`

artursouza: I was wondering if I should offer a new API that takes a callback function instead.

artursouza: See the new design. I think it is cleaner.

examples/src/main/java/io/dapr/examples/invoke/grpc/HelloWorldClient.java
```diff
@@ -45,22 +41,11 @@ public class HelloWorldClient {
...
-      Metadata headers = new Metadata();
-      headers.put(Metadata.Key.of("dapr-app-id", Metadata.ASCII_STRING_MARSHALLER),
-          "hellogrpc");
-
-      // MetadataUtils.attachHeaders is deprecated.
-      blockingStub = blockingStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(headers));
+    try (DaprClient client = new DaprClientBuilder().build()) {
+      HelloWorldGrpc.HelloWorldBlockingStub blockingStub = client.newGrpcStub(
+          "hellogrpc",
+          channel -> HelloWorldGrpc.newBlockingStub(channel));
```
artursouza, line 48: This is the new(est) design to do service invocation with gRPC proxying in Java.

salaboy: @artursouza I am not sure to understand this, this is just for invoking another service right? Do we need to expose the GrpcStub to the user? can we just not hide that behind the `invoke()` method? Sorry, I feel like I am missing something important and a big design decision here.

artursouza: We cannot because, like in this example, we don't have any idea about their service interface (HelloWorldGrpc in this case). Technically, the user does not even need to use our SDK for it, we are just making it easy for him to reuse the gRPC channel and the metadata interceptors behind the scenes.

examples/src/main/java/io/dapr/examples/bindings/http/README.md
```diff
@@ -54,13 +54,13 @@ Run `dapr init` to initialize Dapr in Self-Hosted Mode if it's not already initi
 
 Before getting into the application code, follow these steps in order to set up a local instance of Kafka. This is needed for the local instances.
```
cicoyle, line 55: Should this example also be deleted? its `bindings/http`

examples/src/main/java/io/dapr/examples/invoke/grpc/HelloWorldClient.java
```diff
@@ -70,10 +55,6 @@ public static void main(String[] args) throws Exception {
       } catch (StatusRuntimeException e) {
         logger.log(Level.WARNING, "RPC failed: {0}", e.getStatus());
       }
-    } finally {
```
cicoyle, line 73: Do we still want this?

cicoyle (dismissed):
added a question, but otherwise lgtm 👍🏻

examples/src/main/java/io/dapr/examples/bindings/http/README.md
artursouza: Great question. I am keeping it because the client talks gRPC but the app's endpoint is HTTP. It is different example from when the app is serving gRPC.

examples/src/main/java/io/dapr/examples/invoke/grpc/HelloWorldClient.java
artursouza: No, because the channel is now encapsulated in the DaprClient, which is closed automatically by the try() statement in the end. The DaprClient itself will then close all downstream resources that need to be closed, including the gRPC channel.

Another example of why using the DaprClient is also useful for those that want to make gRPC proxy invocations in Java.

salaboy:
awesome work folks! let's get this one merged!

cicoyle (approved):

PR #1049, opened 2024-06-10
Adding metadata endpoint 

salaboy:
# Description
Adding metadata endpoint on HTTP and gRPC clients. Please review and advice. 

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1048 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #1054, opened 2024-06-20
Increase Kafka container wait time for HTTP bindings examples

artur-ciocanu:
# Description

Currently when running Java SDK examples for HTTP bindings it fails. It seems that using `sleep: 5` to wait for Kafka container to be up and running is not enough. The change is to increase the sleep value from `5` to `20`. This is not ideal, but it proved to work when running:
```
mm.py ./src/main/java/io/dapr/examples/bindings/http/README.md
```
on local box.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/examples/src/main/java/io/dapr/examples/bindings/http/README.md
+++ b/examples/src/main/java/io/dapr/examples/bindings/http/README.md
@@ -60,7 +60,7 @@ Before getting into the application code, follow these steps in order to set up 
 name: Setup kafka container
 expected_stderr_lines:
   - 'Creating network "http_default" with the default driver'
-sleep: 5
+sleep: 20
 -->
 
 ```bash


artur-ciocanu:
@artursouza and @cicoyle after I have adjusted the sleep for Kafka container the `Auto Validate Examples` workflow passed. Here is a sample run: https://github.com/artur-ciocanu/java-sdk/actions/runs/9599423378/job/26473113165

salaboy:
Increasing the sleep usually highlights an issue, but it might be the case that for this test the initialisation is taking way much more time. I will defer to @artursouza and @cicoyle to validate

cicoyle (approved):
Seems reasonable to me, in looking at the logs I see `duration_seconds: 16.4` for this step: `Setup kafka container`. I'm uncertain this fixes all the issues from [this PR's ](https://github.com/dapr/java-sdk/pull/1049)autovalidate examples CI step, but we shall see

artur-ciocanu:
> Increasing the sleep usually highlights an issue, but it might be the case that for this test the initialisation is taking way much more time. I will defer to @artursouza and @cicoyle to validate

@salaboy I totally agree. I am not sure if `https://github.com/dapr/mechanical-markdown` aka `mm.py` script that is used to run the tests for examples has a way to wait for single or something from the Docker Compose so it doesn't use `sleep`. @cicoyle any thoughts?

artursouza:
The idea would be to have the sidecar to retry the init for some time before crashing. For this scope, timeout bump works.

artursouza (approved):

artur-ciocanu:
@artursouza and @cicoyle it seems that everything is 👍. Could you please merge the PR?

Issue #849, opened 2023-04-25
FEATURE PREVIEW: Add Actor State TTL support


JoshVanL:
Add first class API support for Actor State TTL https://github.com/dapr/dapr/issues/5899

A dedicated function to Actor State should be created. Psudo code: `func (key string, value string, ttl duration)`. It should be made clear in the SDK documentation/comments that users should always use this function, unless they have specifically created some kind of Actor State clean up out of band of Dapr or don't have an issue with the Actor State store keep state and grow "forever".

Please see the [go-sdk](https://github.com/dapr/go-sdk/pull/383/files) for implementation reference.

JoshVanL:
UPDATE: Due to the nature of the current implementation of write through caching of actor state and the unavailability of the real TTL expire time of state keys, SDKs will have an inconsistent view of the world when it has a cold cache and the state store has TTL keys. The TTL functionality has been put behind a feature gate in daprd. See https://github.com/dapr/dapr/pull/6400 for more details. This feature is expected to be moved to GA in 1.12.

cicoyle:
[Inspiration can be derived from this PR adding ttl to actor reminder and timer that was not completed](https://github.com/dapr/java-sdk/pull/658)

cicoyle:
Hey @WhitWaldo - [the PR is not merged yet](https://github.com/dapr/java-sdk/pull/1060), its on my TODO for this upcoming release to fix it up. I think this issue should stay open.

WhitWaldo:
> Hey @WhitWaldo - [the PR is not merged yet](https://github.com/dapr/java-sdk/pull/1060), its on my TODO for this upcoming release to fix it up. I think this issue should stay open.

Hmm... Something must have bugged out somewhere as I don't recall intentionally closing this. Absolutely let's reopen it.

PR #1060, opened 2024-06-26
Add actor state TTL support.

artursouza:
# Description

Add actor state TTL support.

## Issue reference

Please reference the issue this PR will close: #849 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


sdk-tests/components/statestore.yaml
```diff
@@ -10,5 +10,3 @@ spec:
     value: localhost:6379
   - name: redisPassword
     value: ""
-  - name: actorStateStore
```
artursouza, line 13: I had to change the actor state store, so the work left is to use mysql and add all actor apps to the scopes of the new mysql state store.

sdk-tests/components/actorstatestore.yaml
```diff
@@ -0,0 +1,16 @@
...
+  version: v1
+  metadata:
+    - name: connectionString
+      value: "root:test@tcp(127.0.0.1:3306)/?allowNativePasswords=true"
+    - name: cleanupIntervalInSeconds
+      value: "1"
+    - name: actorStateStore
+      value: "true"
+scopes:
+  - actorstateit-statefulactorservice
```
artursouza, line 16: Needs to add the other actor apps too. This way, other tests don't require actor state store.

sdk-tests/deploy/local-test.yml
```diff
@@ -26,3 +26,10 @@ services:
     image: mongo
     ports:
       - "27017:27017"
+
+  mysql:
```
artursouza, line 30: Need to change workflow to install mysql.

cicoyle (approved):
lgtm

sdk-tests/components/actorstatestore.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - activationdeactivationit-demoactorservice
+  - actorexceptionit-myactorservice
+  - actormethodnameit-myactorservice
+  - actorreminderfailoveritone-myactorservice
+  - actorreminderfailoverittwo-myactorservice
+  - actorreminderrecoveryit
+  - actorsdkresiliencytit-demoactorservice
+  - actorstateit-statefulactorservice
+  - actortimerrecoveryit
+  - actorturnbasedconcurrencyit-myactorservice
```
artur-ciocanu, line 26: Nit, add new line

artur-ciocanu (approved):
@artursouza looks great!

I have a more general design question related to ActorState. ActorState uses a generic type, however in ActorStateManager it is used without generic type. I am wondering if we should try to enforce generic type usage across other classes like: AbstractActor, StateChangeMetadata, ActorStateOperation, etc to ensure a stronger type safety.

What are your thoughts?

artursouza:
> @artursouza looks great!
> 
> I have a more general design question related to ActorState. ActorState uses a generic type, however in ActorStateManager it is used without generic type. I am wondering if we should try to enforce generic type usage across other classes like: AbstractActor, StateChangeMetadata, ActorStateOperation, etc to ensure a stronger type safety.
> 
> What are your thoughts?

Let me give it a try. These are not exposed to the user and are for internal use only. There might be a reason I did not propagate the generic type everywhere, but I don't remember now. Again, let me revisit this.

sdk-actors/src/main/java/io/dapr/actors/runtime/ActorStateManager.java
```diff
@@ -309,15 +345,26 @@ private static final class StateChangeMetadata {
      */
     private final Object value;
```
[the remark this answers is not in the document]
artursouza: It does not need because this is only to serialize and not to be deserialized.

artursouza:
> > @artursouza looks great!
> > I have a more general design question related to ActorState. ActorState uses a generic type, however in ActorStateManager it is used without generic type. I am wondering if we should try to enforce generic type usage across other classes like: AbstractActor, StateChangeMetadata, ActorStateOperation, etc to ensure a stronger type safety.
> > What are your thoughts?
> 
> Let me give it a try. These are not exposed to the user and are for internal use only. There might be a reason I did not propagate the generic type everywhere, but I don't remember now. Again, let me revisit this.

See my reply. This is not needed because it is used for serialization and does not interface with the user's API.

artur-ciocanu:
@artursouza could you please check the compilation failures for SB 3.2 and SB 3.3, once those are resolved. I think we are good to merge the PR.

Thank you!

cicoyle (approved):
approving again, assuming CI passes

artur-ciocanu (dismissed):
Looks really, really good!

artur-ciocanu (approved):
Looks great!

artur-ciocanu (approved):
Approving one more time for CI

artur-ciocanu:
@artursouza if you'll have some time could you please check why the tests are failing. I have tried to fix the compilation issues, however I am not sure why the IT is failing.

Thank you!

salaboy:
@cicoyle it might be good if you take this issue, as you were involved.. IMHO the best way to proceed is to create a new PR using the changes from this PR, so we have no DCO issues and then make sure that the tests are green.

cicoyle:
Closing this PR to proceed [with this PR](https://github.com/dapr/java-sdk/pull/1338)

Issue #995, opened 2024-01-24
Java 8 date/time type `java.time.ZonedDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling


Qoooooooooooo:
version 1.10.0
When use daprClient.invokeMethod and the dto includes ZonedDateTime、LocalDateTime.
error:
Java 8 date/time type `java.time.ZonedDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling

Qoooooooooooo:
how to enable jackson-datatype-jsr310 in dapr

Qoooooooooooo:
This issue still exists after adding dependencies to pom.xml

artursouza:
OK. I accept a PR fixing this.

BrayanMnz:
Do we have a sample project to replicate this? @Qoooooooooooo @artursouza

Shubhdeep02:
Hey @artursouza is it still considered as `good-first-issue`?

mcruzdev:
I opened a draft pull request but must test it correctly

Manuel-Moya:
Any progress on this?

salaboy:
@mcruzdev the PR looks good to me.. Can others verify this work? I am curious to see if @Manuel-Moya had the same issue and the PR fixes this for his projects.

cicoyle:
I believe this issue can be closed since the linked PR was merged

zikahk:
I'm a little bit late but this issue also still exists with Java 8 date/time type  --> `java.time.Duration` not supported by default:
add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
It is a little bit strange considering I'm using jackson 3 API

PR #1065, opened 2024-07-02
Add JavaTimeModule

mcruzdev:
# Description

## Issue reference

#995 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


cicoyle:
Ping @mcruzdev were you still working on this?

mcruzdev:
Hi @cicoyle to be sincerous no, but I would like to continue, do you think to get this one?

mcruzdev:
I need to add tests and probably to upgrade the dependency version

mcruzdev:
Hi @cicoyle, sorry for delay... I think this one solve the issue.

I tested calling the `invokeMethod` withouth JavaTimeModule and the error was reproduced.

mcruzdev:
Hi @cicoyle @salaboy, am I in the right path, WDYT?

salaboy:
@artursouza can you check this? I am not sure to have the full context to evaluate these changes.. but looking at the code it looks ok to me. Do we want to have that new dependency in the SDK? I am ok with it.

sdk/src/test/java/io/dapr/serializer/DefaultObjectSerializerTest.java
```diff
@@ -71,6 +74,8 @@ public static class MyObjectTestToSerialize implements Serializable {
     @JsonDeserialize(using = OffsetDateTimeDeserializer.class)
     private OffsetDateTime timeValue;
```
artur-ciocanu, line 76: Could we please add `private` similar to other fields.

artur-ciocanu (approved):
@mcruzdev thank you for your contribution, it looks really good!

artur-ciocanu:
@mcruzdev could you please make sure you rebase your branch.

artur-ciocanu:
@mcruzdev as I was thinking about your PR, it occurred to me that there could be a lot of other instances where we would like to register a custom Jackson module. My suggestion is to see how can we expand the `ObjectSerializer` API a little bit and maybe adding a `ObjectSerializerFactory` with two methods:
- `ObjectSerializerFactory.createDefaultSerializer()` - this will create the default serializer that is currently used everywhere
- `ObjectSerializerFactory.createJacksonSerializer(ObjectMapper om)` - this will be used to create a Jackson based serialized where we can customize the `ObjectMapper` and ensure we can add or remove all the necessary modules

CC: @cicoyle @artursouza

mcruzdev:
Good suggestions @artur-ciocanu TY, I will update the PR.

mcruzdev:
Hi @artur-ciocanu, could you take a new review?

sdk/pom.xml
```diff
@@ -39,6 +39,11 @@
       <artifactId>jackson-databind</artifactId>
       <version>2.15.1</version>
     </dependency>
```
artur-ciocanu, line 41: Could you please remove this dependency. We don't want to enforce anything beyond Jackson Core and Jackson Databind.

Dapr Java SDK users can decide to use whatever Jackson module they need, we just need to have a way to allow them to customize the serializer.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
```diff
@@ -35,16 +38,29 @@ public class ObjectSerializer {
    */
   protected static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
```
artur-ciocanu, line 39: Could you please revert this, we don't want to change the default ObjectMapper instance.

sdk/src/main/java/io/dapr/client/ObjectSerializer.java
```diff
@@ -35,16 +38,29 @@ public class ObjectSerializer {
...
 
+  protected ObjectMapper objectMapper;
+
   /**
    * Default constructor to avoid class from being instantiated outside package but still inherited.
    */
   protected ObjectSerializer() {
+    this(OBJECT_MAPPER);
+  }
+
```
artur-ciocanu, line 53: I would propose the following change, instead of a protected constructor we can have something like this:

```java
protected ObjectMapper getObjectMapper() {
  return OBJECT_MAPPER;
}
```
And in all the places we use `OBJECT_MAPPER` directly we would replace it with `getObjectMapper()`.

Now the beautiful part about it is that we can create our own class like `CustomizableDaprObjectSerializer` that can look like this:

```java
public class CustomizableObjectSerializer extends ObjectSerializer implements DaprObjectSerializer {

  private final ObjectMapper objectMapper;

  public CustomizableObjectSerializer(ObjectMapper objectMapper) {
    this.objectMapper = objectMapper;
  }

  @Override
  protected ObjectMapper getObjectMapper() {
    return objectMapper;
  }
  
  @Override
  public String getContentType() {
    return "application/json";
  }
}
```

This way we reuse most of `ObjectSerializer` functionality and customize just the ObjectMapper.

sdk/src/main/java/io/dapr/serializer/CustomizableDaprObjectSerializer.java
```diff
@@ -0,0 +1,33 @@
...
+import io.dapr.client.ObjectSerializer;
+import io.dapr.utils.TypeRef;
+
+import java.io.IOException;
+import java.util.Objects;
+
+public class CustomizableDaprObjectSerializer implements DaprObjectSerializer {
+
+  private final ObjectSerializer delegate;
+
```
artur-ciocanu, line 13: I would rename this one to `CustomizableObjectSerializer` and as I already mentioned I think we can use an implementation like this:
```java
public class CustomizableObjectSerializer extends ObjectSerializer implements DaprObjectSerializer {

  private final ObjectMapper objectMapper;

  public CustomizableObjectSerializer(ObjectMapper objectMapper) {
    this.objectMapper = objectMapper;
  }

  @Override
  protected ObjectMapper getObjectMapper() {
    return objectMapper;
  }
  
  @Override
  public String getContentType() {
    return "application/json";
  }
}
```

artur-ciocanu (commented):
@mcruzdev thanks a lot for the updates, I have provided some additional comments, also I think it would make sense to provide some examples on how anyone could customize the serializers.

I think it would make sense to have a separate unit test for `CustomizableObjectSerializer` where we could show how one could customize the serializer and how ZonedDateTime would be serialized using Jackson JavaTimeModule

sdk/src/main/java/io/dapr/serializer/ObjectSerializerFactory.java
```diff
@@ -0,0 +1,10 @@
+package io.dapr.serializer;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
```
artur-ciocanu, line 4: It seems that this is used only in the test, up to you, you can keep it or remove it and add it in the future, where we see a real need for this factory.

artur-ciocanu:
@cicoyle and @salaboy could you also please take a look? Thank you.

artur-ciocanu (approved):
Awesome job! 👍

artur-ciocanu:
@cicoyle I think this PR looks great and it is a good example on how anyone can extend the `DaprObjectSerializer` to add other Jackson modules, this was a huge pain in the past.

Could you please review and approve. Thank you!

artur-ciocanu (approved):
Approving for CI

artur-ciocanu:
@cicoyle could you please approve this PR. Thank you.

salaboy (approved):
LGTM ! thanks @mcruzdev !

artur-ciocanu:
@cicoyle could you please approve and merge this PR. Thank you.

PR #1097, opened 2024-08-09
Support static Properties overrides on DaprClientBuilder

salaboy:
# Description

After discussing this with @cicoyle and @artursouza I've created this PR to provide a way to override variables that are set using System properties and environments. 

## Issue reference

Fixes #1096

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Extended the documentation

PR #1098, opened 2024-08-13
PR-1089 plus Dapr Client props override

artur-ciocanu:
# Description

This is a PR similar to #1089 but it also incorporates changes from #1097 primarily to ensure that integration tests based on Testcontainers are working as expected.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


sdk-tests/src/test/java/io/dapr/it/spring/data/MySQLDaprKeyValueTemplateIT.java
```diff
@@ -0,0 +1,246 @@
...
+
+  @Autowired
+  private DaprClient daprClient;
+
+  @Autowired
+  private DaprKeyValueTemplate keyValueTemplate;
+
+  @BeforeEach
+  public void waitSetup() {
+    daprClient.waitForSidecar(1000).block();
```
salaboy, line 118: @artursouza we are doing this now, but I am seriously thinking to move this down into the DaprKeyValueTemplate so the user doesn't need to worry about handling the `daprClient` would it make sense to run `daprClient.waitForSidecar(1000).block()` before each Dapr API operation? Is that the intended use for all the operations?

sdk-tests/src/test/java/io/dapr/it/spring/data/PostgreSQLDaprKeyValueTemplateIT.java
```diff
@@ -0,0 +1,232 @@
...
+  @Container
+  private static final PostgreSQLContainer<?> POSTGRE_SQL_CONTAINER = new PostgreSQLContainer<>("postgres:16-alpine")
+      .withNetworkAliases("postgres")
+      .withDatabaseName("dapr_db")
+      .withUsername("postgres")
+      .withPassword("password")
+      .withExposedPorts(5432)
+      .withNetwork(DAPR_NETWORK);
+
+  private static final WaitStrategy DAPR_CONTAINER_WAIT_STRATEGY = Wait.forHttp("/v1.0/healthz")
```
salaboy, line 71: @artur-ciocanu maybe we can make this internal.. meaning that this can be the default strategy. If I remember correctly, depending the parameters that we set we might run into a loop of `daprd` waiting for the app port to be exposed and the test waiting for `daprd` to be ready. But as a user I don't want to be dealing with this.

sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprContainerTest.java
```diff
@@ -97,19 +93,23 @@ private static void configStub() {
...
   }
 
   @Test
   public void testStateStore() throws Exception {
-    try (DaprClient client = (new DaprClientBuilder()).build()) {
+    DaprClientBuilder builder = new DaprClientBuilder()
+        .withPropertyOverride(Properties.HTTP_PORT, String.valueOf(DAPR_CONTAINER.getHttpPort()))
+        .withPropertyOverride(Properties.GRPC_PORT, String.valueOf(DAPR_CONTAINER.getGrpcPort()));
+    try (DaprClient client = (builder).build()) {
       client.waitForSidecar(5000).block();
```
salaboy, line 113: @artur-ciocanu can we just use 1s all over the place? If we are waiting for more than 1 second then something is wrong somewhere else.

sdk-tests/src/test/java/io/dapr/it/spring/messaging/TestApplication.java
```diff
@@ -0,0 +1,51 @@
...
+
+@SpringBootApplication
+public class TestApplication {
+  public static void main(String[] args) {
+    SpringApplication.run(TestApplication.class, args);
+  }
+
+  @Configuration
+  @EnableConfigurationProperties(DaprPubSubProperties.class)
+  static class DartSpringMessagingConfiguration {
```
salaboy, line 36: @artur-ciocanu fix typo in the class name

dapr-spring/dapr-spring-boot-autoconfigure/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+      <groupId>io.dapr.spring</groupId>
+      <artifactId>dapr-spring-core</artifactId>
+      <version>${project.parent.version}</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.dapr.spring</groupId>
+      <artifactId>dapr-spring-data</artifactId>
+      <version>${project.parent.version}</version>
+      <optional>true</optional>
```
salaboy, line 28: @artur-ciocanu if these are optional, then, as a user we don't get these dependencies if we include the `dapr-spring-boot-starter` which means that we need to separately add this dependencies manually.. was there a reason why this needs to be optional?

sdk-tests/src/test/java/io/dapr/it/spring/data/DaprKeyValueRepositoryIT.java
```diff
@@ -0,0 +1,154 @@
...
+import org.testcontainers.containers.wait.strategy.WaitStrategy;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static io.dapr.it.spring.data.DaprSpringDataConstants.BINDING_NAME;
```
salaboy, line 41: @artur-ciocanu do we want to make this user facing constants? This should be part of the DaprPubSub and DaprStatestore properties no?

salaboy: @artur-ciocanu the binding and the statestore name should have default values here: DaprStateStoreProperties

cicoyle:
Mind resolving the conflicts here :)

artur-ciocanu:
This is no longer required. All the changes have been ported to #1089.

artur-ciocanu:
This is no longer required. All the changes have been ported to #1089.

Issue #1112, opened 2024-08-30
DaprWorkflowClient and WorkflowRuntimeBuilder doesn't allow properties overrides


artur-ciocanu:
## Expected Behavior

Developers should be able to override GRPC port, GRPC endpoint for `DaprWorkflowClient` and `WorkflowRuntimeBuilder` via properties override mechanism similar to this one #1097.

## Actual Behavior

Currently it is impossible to override configurations like GRPC port, for workflow related classes unless we use `System.setProperty()` or env variables.

## Steps to Reproduce the Problem

Check the code for `DaprWorkflowClient` and `WorkflowRuntimeBuilder` and see that there is no way to provide custom properties.

## Release Note

RELEASE NOTE:

PR #1113, opened 2024-08-30
Add configuration overrides for DaprWorkflowClient and WorkflowRuntimeBuilder

artur-ciocanu:
# Description

This PR addresses a small limitation, it ensures that DaprWorkflowClient and WorkflowRuntimeBuilder allows configuration overrides via properties override mechanism.

## Issue reference

Please reference the issue this PR will close: #1112 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


artur-ciocanu:
@artursouza @salaboy as I was working on this PR I have discovered that the workflow tests weren't picked up by Maven Fail Safe plugin since these were ending in `Test` instead of `IT`, so I had to rename the test files.

Please take a look and let me know if everything is looking good.

artur-ciocanu:
@artursouza @cicoyle and @salaboy the build 🟢, could you please review and approve. Thank you! 🙇

salaboy:
Great catch.. I will review and approve later today

salaboy (approved):

salaboy:
@caniszczyk @artursouza This is ready to be merged.. please take a look, I've approved it already, but that is not enough :)

artur-ciocanu:
@salaboy as discussed I have simplified the `TestcontainersDaprClientCustomizer` and I also added the necessary tests to have proper coverage.

Could you please take another look?

sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprWorkflowsIT.java
```diff
@@ -24,66 +28,96 @@
...
 @Tag("testcontainers")
-public class DaprWorkflowsTests {
+public class DaprWorkflowsIT {
+
+  private static final Network DAPR_NETWORK = Network.newNetwork();
+
+  @Container
+  private static final DaprContainer DAPR_CONTAINER = new DaprContainer("daprio/daprd:1.13.2")
+      .withAppName("workflow-dapr-app")
+      .withNetwork(DAPR_NETWORK)
```
stephane-dereppe, line 60: Hi.

How do you pass the WebEnvironment.RANDOM_PORT to dapr? I don't see any `DAPR_CONTAINER.withAppPort(XXX)`

salaboy: @stephane-dereppe there is an environment variable that you can inject :

```
@Value("${server.port}")
private int serverPort;
```

But I don't think this will work as the Dapr Container definition is done in a static field.

stephane-dereppe: Indeed.

```
@Value("${server.port}") 
private int serverPort;
```

Is equivallent to 

```
@LocalServerPort
private Integer server;
```

Which isn't available in a static context.

The only way I managed to make it work is to start the dapr container just before the test (in a non static context) when the @LocalServerPort annotation has be processed by spring boot to inject the port actually used by the test.

```
    @LocalServerPort
    private Integer randomServerPort;

    @BeforeEach
    void setup() {
        daprContainer.withAppPort(randomServerPort);
        daprContainer.start();

        System.setProperty("dapr.http.port", Integer.toString(daprContainer.getHTTPPort()));
        System.setProperty("dapr.grpc.port", Integer.toString(daprContainer.getGRPCPort()));
        System.setProperty("dapr.sidecar.ip", daprContainer.getContainerIpAddress());
        Testcontainers.exposeHostPorts(randomServerPort);
    }
```

But I'm not totally satisfied because it means dapr (test)container is initialized differently than my other containers... 

see my question/remark in the discord channel: https://discord.com/channels/778680217417809931/778749797242765342/1273159006663475201

So I'm wondering how the integration test in this PR can work... I guess dapr doesn't start until it can reach the app... Am I wrong?

salaboy: Yeah this is not good.. but the problem is that Dapr (the sidecar) needs to wait for the application to be up, in order to finish the initialization and because the way that containers are started by Testcontainers (before the app context starts) the App will never be ready before the containers. @artur-ciocanu provided a wait strategy for the Dapr container, but I am not sure if it covers all the use cases.

stephane-dereppe: I don't see any workaround if --app-port cannot be set after dapr container startup

salaboy: We need to find other containers in the same situation or change Dapr to support a different mode to start up that fits better this model

artur-ciocanu: @stephane-dereppe thanks for chiming in. We are working on adding proper support for this. @salaboy is working on adding `@ServiceConnection` for Dapr container. However I don't think we will be able to have a `static` container that can be reused.

salaboy: @artur-ciocanu @stephane-dereppe I need to try out things.. but instead of using static containers we can use beans.. that doesn't change the sequencing, but it gives us some more flexibility for the testing.

artursouza (approved):

Issue #1072, opened 2024-07-09
 Add Streaming Subscription Support 


JoshVanL:
Dapr v1.14 adds support for [streaming subscriptions](https://github.com/dapr/proposals/blob/main/0013-RS-pubsub-subscription-streaming.md) which allows for dynamically subscribing to pubsub topics over a bi-directional gRPC stream. This API is exposed via the [SubscribeTopicEventsAlpha1](https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L64) gRPC API.

On first call, the client sends an [InitialRequest](https://github.com/dapr/dapr/blob/7ea7016c0150e6a7b7bc7c30e34d56f0001c5e90/dapr/proto/runtime/v1/dapr.proto#L434) containing the pubsub, topic name, etc. From then on, topic event messages are sent from daprd to the client over the stream. The client reports the processing status of this message back to daprd. A message will not be considered processed until the client has responded with this status. There can be multiple in-flight topic messages. Daprd unsubscribes from the topic when the stream is closed.

https://github.com/dapr/go-sdk/pull/578 is a reference implementation for the go-sdk.

PR #1142, opened 2024-10-10
Fix bi-di subscription to support dapr-api-token

artursouza:
# Description

Fix bi-di subscription to support dapr-api-token

## Issue reference

Please reference the issue this PR will close: #1072 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
@@ -59,7 +59,7 @@ public class ActorClient implements AutoCloseable {
    * @param overrideProperties Override properties.
    */
   public ActorClient(Properties overrideProperties) {
-    this(buildManagedChannel(overrideProperties), null);
+    this(buildManagedChannel(overrideProperties), null, overrideProperties.getValue(Properties.API_TOKEN));
   }
 
   /**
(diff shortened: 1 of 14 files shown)


sdk-actors/src/main/java/io/dapr/actors/client/ActorClient.java
```diff
@@ -59,7 +59,7 @@ public ActorClient(ResiliencyOptions resiliencyOptions) {
    * @param overrideProperties Override properties.
    */
   public ActorClient(Properties overrideProperties) {
-    this(buildManagedChannel(overrideProperties), null);
+    this(buildManagedChannel(overrideProperties), null, overrideProperties.getValue(Properties.API_TOKEN));
```
artursouza, line 62: I also needed to fix how properties override work because it was not handling it correctly for Dapr API Token. Now, it works too and all ITs require sidecar calls to have dapr-api-token header.

sdk/src/main/java/io/dapr/client/DaprClientImpl.java
```diff
@@ -425,7 +474,8 @@ private <T> Subscription<T> buildSubscription(
       SubscriptionListener<T> listener,
       TypeRef<T> type,
       DaprProtos.SubscribeTopicEventsRequestAlpha1 request) {
-    Subscription<T> subscription = new Subscription<>(this.asyncStub, request, listener, response -> {
+    var interceptedStub = this.grpcInterceptors.intercept(this.asyncStub);
```
artursouza, line 477: This is the fix for streaming subscription to handle Dapr API token header.

cicoyle (approved):

cicoyle:
Changes lgtm, but I see some build failures - might be flaky

artursouza:
> Changes lgtm, but I see some build failures - might be flaky

I have been fixing those for the past few days. Almost there.

artursouza:
> > Changes lgtm, but I see some build failures - might be flaky
> 
> I have been fixing those for the past few days. Almost there.

I think that was the last one. Waiting for the run now.

Issue #1145, opened 2024-10-14
waitForSidecar should respect timeoutInMilliseconds


artursouza:
## Expected Behavior

When calling `waitForSidecar(timeoutInMilliseconds)`, the expected behavior is that it will wait for the sidecar up to `timeoutInMilliseconds`, it does not matter how many retries are done behind the scenes.

## Actual Behavior

The current implementation uses timeoutInMilliseconds but it does not apply after 5 retries of 500ms interval. See https://github.com/dapr/java-sdk/blob/b1d0be76e698b4cbf9d78850741c7a76b960d967/sdk/src/main/java/io/dapr/client/DaprClientImpl.java#L224-L278

## Steps to Reproduce the Problem

Use the method with a value larger than `2.5s` (the larger the better to visually identify the problem) without having a sidecar to connect to, forcing the timeout.

## Release Note

RELEASE NOTE:  **FIXED** handling of timeout in `waitForSidecar`

artursouza:
Ideally, the integration test should repro this bug, so we avoid future regressions.

olitomlinson:
It would be great to release a patch version of the SDK for this as it does cause us daily issues in production.

olitomlinson:
@artursouza are you still planning to tackle this by the end of the week? If so that would be great because we're starting code freeze in my org on Monday and it would be awesome if we could have this resolved. Much appreciated!

olitomlinson:
@artursouza please can we get this released under a patch release 1.12.3?

PR #1146, opened 2024-10-17
Fix waitForSidecar to respect timeout.

artursouza:
# Description

Fix waitForSidecar to respect timeout.

## Issue reference

Please reference the issue this PR will close: #1145 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprClientImpl.java
+++ b/sdk/src/main/java/io/dapr/client/DaprClientImpl.java
@@ -273,53 +273,13 @@ public class DaprClientImpl extends AbstractDaprClient {
   @Override
   public Mono<Void> waitForSidecar(int timeoutInMilliseconds) {
     String[] pathSegments = new String[] { DaprHttp.API_VERSION, "healthz", "outbound"};
-    int maxRetries = 5;
-
-    Retry retrySpec = Retry
-        .fixedDelay(maxRetries, Duration.ofMillis(500))
-        .doBeforeRetry(retrySignal -> {
-          System.out.println("Retrying component health check...");
-        });
-
-    /*
-    NOTE: (Cassie) Uncomment this once it actually gets implemented:
-    https://github.com/grpc/grpc-java/issues/4359
-
-    int maxChannelStateRetries = 5;
-
-    // Retry logic for checking the channel state
-    Retry channelStateRetrySpec = Retry
-            .fixedDelay(maxChannelStateRetries, Duration.ofMillis(500))
-            .doBeforeRetry(retrySignal -> {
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/io/dapr/client/DaprClientImpl.java
```diff
@@ -273,53 +273,13 @@ public <T extends AbstractStub<T>> T newGrpcStub(String appId, Function<Channel,
...
-              // Return true if the status code is in the 2xx range
-              if (statusCode >= 200 && statusCode < 300) {
-                return Mono.empty(); // Continue with the flow
-              }
-            }
-            return Mono.error(new RuntimeException("Health check failed"));
-          }).retryWhen(channelStateRetrySpec);
-        })
-        */
+        .retryWhen(Retry.indefinitely())
```
olitomlinson, line 282: This is great.

However, I would prefer if we left the fixed duration wait between attempts and the log line.

```
 Retry retrySpec = Retry
        .fixedDelay(maxRetries, Duration.ofMillis(500))
        .doBeforeRetry(retrySignal -> {
          System.out.println("Retrying component health check...");
        });
```

This log line is helpful to us because it gives me as an Operator some logging insight into how long its taking the sidecar to become responsive. Also, having no `fixedDelay` between attempts could generate lots of unnecessary log spam which wouldn't be great.

It would also maintain parity with the dotnet SDK

```
public override async Task WaitForSidecarAsync(CancellationToken cancellationToken = default)
        {
            while (true)
            {
                var response = await CheckOutboundHealthAsync(cancellationToken);
                if (response)
                {
                    break;
                }
                await Task.Delay(TimeSpan.FromMilliseconds(500), cancellationToken);
            }
        }
```

olitomlinson (changes requested):

sdk/src/main/java/io/dapr/client/DaprClientImpl.java
artursouza: No problem. I can make that happen. Regarding the log line, println is not the best way to log, so I will change to SLF4J.

artur-ciocanu (approved):
Love it! Looks really good!

sdk/src/main/java/io/dapr/client/DaprClientImpl.java
olitomlinson: Agree with SLF4J. It was kinda painful not having any event attributes, such as timestamps/thread id that kinda thing.

cicoyle (approved):

olitomlinson (approved):

PR #1160, opened 2024-11-05
Simplify OTEL integration to reduce deps

artur-ciocanu:
# Description

This PR has been created after looking at Micrometer Observation integration in `PulsarTemplate` here: https://github.com/spring-projects/spring-pulsar/blob/main/spring-pulsar/src/main/java/org/springframework/pulsar/core/PulsarTemplate.java#L298. Many thanks to @salaboy for pointing this out.

The main idea was to remove explicit OTEL dependency from Dapr Messaging since Micrometer Observation should be able to handle `traceparent` generation behind the scenes and OOTB. 

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


artur-ciocanu:
@artursouza and @cicoyle this is a tiny PR to get rid of extra dependencies and ensure that we stay in the realm of Spring and Micrometer without having to bring in OTEL explicitly.

artursouza (approved):

Issue #1155, opened 2024-10-29
[Workflows] Rename `callSubWorkflow` to `callChildWorkflow`


olitomlinson:
The Java SDK uses the method name `callSubWorkflow` rather than `callChildWorkflow` in the [workflow context](https://github.com/dapr/java-sdk/blob/0b7a051b79c7a394e9bd4f57bd40778fb5f29897/sdk-workflows/src/main/java/io/dapr/workflows/WorkflowContext.java#L376)

The Dapr documentation refers to [child workflows](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-features-concepts/#child-workflows) , so I would recommend that the Java SDK change the method name to `callChildWorkflow` for consistency.

salaboy:
@olitomlinson ok.. PR created.. I will follow the tests to see if I am missing something. 
@artursouza do you know if we need to deprecate the old methods? Do you know how to create a release note for this change? It is a simple replace to upgrade to this new signature.

PR #1168, opened 2024-12-04
Replacing subWorkflow to childWorkflow to align with other SDKs

salaboy:
# Description

using ChildWorkflows instead of SubWorkflows to align with other SDKs and docs

## Issue reference

Please reference the issue this PR will close: #1155 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


cicoyle (approved):
Rename lgtm, I believe we will need to make note of this being a breaking change since it'll break users with `ctx.callSubWorkflow` since its now: `ctx.callChildWorkflow`. Otherwise, we should mark it as a deprecated alias and leave it for 2 releases then remove it. cc/ @artursouza for thoughts?

artursouza:
workflows is alpha, so we can break it.

Issue #1212, opened 2025-02-08
Add app health checks support to Dapr Testcontainer


artur-ciocanu:
## Describe the proposal
Currently there is no way to configure app health check when using Dapr Testcontainer. The proposal is to add all the necessary command line flags to support app health checks.

This should resolve a long standing issue that we have with Dapr Testcontainer and Spring Boot Test race condition.

PR #1213, opened 2025-02-08
Add app health check support to Dapr Testcontainer

artur-ciocanu:
# Description

Although we added support for Dapr Testcontainer sometimes we see race conditions between Dapr daemon and the application that should be run by Dapr. This leads to flaky tests and poor user experience.

The purpose of this PR is to add support for Dapr app health checks. The `DaprContainer` has been enhanced with a new property named `appHealthCheckPath`. If the `withAppHealthCheckPath(...)` builder method is invoked we will add the following CLI args to Dapr daemon:
```
--enable-app-health-check
--app-health-check-path <path to health endpoint>
```
App health checks CLI args are not enough though. What we have observed is that both the app and the Dapr daemon are up and running, but not all the components are setup or ready. We have tried multiple strategies and none of them work correctly. Currently instead of using "dumb" `Thread.sleep(...)` the proposal is to leverage Testcontainer `WaitStrategy` to lookout for Dapr logs that tells us when things have been properly bootstrapped. Here is a sample:
```java
Wait.forLogMessage(SUBSCRIPTION_MESSAGE_PATTERN, 1).waitUntilReady(DAPR_CONTAINER);
```
The downside of this approach is that users will have to add this wait strategy to their tests, but it is definitely better than `Thread.sleep`. 

## Issue reference

Please reference the issue this PR will close: #1212 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/spring/messaging/DaprSpringMessagingIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/spring/messaging/DaprSpringMessagingIT.java
@@ -20,6 +20,7 @@ import io.dapr.testcontainers.Component;
 import io.dapr.testcontainers.DaprContainer;
 import io.dapr.testcontainers.DaprLogLevel;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


artur-ciocanu:
@artursouza and @cicoyle the build is 🟢 could you please review and approve.

@artursouza and @salaboy thanks for the hint related to app check health!

salaboy:
@artur-ciocanu amazing stuff
@artursouza @cicoyle can you please review and merge ASAP?

artursouza (approved):

Issue #1217, opened 2025-02-16
Replace OkHttpClient with Java 11 HttpClient


artur-ciocanu:
## Describe the proposal
A few months ago we have upgraded Dapr Java SDK to use Java 11. Java 11 has native support for HTTP client. The proposal is to use the Java 11 HttpClient instead of OkHttpClient. While OkHttpClient is a nice library it adds yet another dependency that we have to manage and upgrade. Also the fact that OkHttpClient uses Kotlin adds another layer of complications since the Kotlin version in OkHttpClient might conflict with our customers Kotlin stdlib.

PR #1218, opened 2025-02-16
Replacing OkHttpClient with Java 11 HttpClient

artur-ciocanu:
# Description
A few months ago we have upgraded Dapr Java SDK to use Java 11. Java 11 has native support for HTTP client. The proposal is to use the Java 11 HttpClient instead of OkHttpClient. While OkHttpClient is a nice library it adds yet another dependency that we have to manage and upgrade. Also the fact that OkHttpClient uses Kotlin adds another layer of complications since the Kotlin version in OkHttpClient might conflict with our customers Kotlin stdlib.

## Issue reference

Please reference the issue this PR will close: #1217 
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


artur-ciocanu:
@artursouza @cicoyle and @salaboy please take a look and let me know your thoughts. We are still using `OkHttpClient` in ITs, but I will take care of those in a separate PR, so this is more focused.

artur-ciocanu:
@artursouza I might need your help, as far as I can see the only failing test `WaitForSidecarIT.waitSlow()` this test uses `ToxiProxy` I am not sure why it fails. Is `OkHttpClient` something smart behind the scenes compare to Java 11 HttpClient?

sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java
```diff
@@ -68,38 +55,30 @@ public DaprHttp build(Properties properties) {
...
-          ConnectionPool pool = new ConnectionPool(properties.getValue(HTTP_CLIENT_MAX_IDLE_CONNECTIONS),
-                  KEEP_ALIVE_DURATION, TimeUnit.SECONDS);
-          builder.connectionPool(pool);
-
-          OK_HTTP_CLIENT = builder.build();
+        if (HTTP_CLIENT == null) {
+          int maxRequests = properties.getValue(HTTP_CLIENT_MAX_REQUESTS);
+          Executor executor = Executors.newFixedThreadPool(maxRequests);
+          HTTP_CLIENT = HttpClient.newBuilder()
+              .executor(executor)
```
artur-ciocanu, line 64: When connecting to `localhost` `OkHttpClient` use `HTTP 1.1` so we need to use the same HTTP protocol version.

artur-ciocanu:
@artursouza and @cicoyle with @ThomasVitale's help I have managed to fix the ITs. Everything is 🟢.

salaboy:
@artur-ciocanu this looks good to me.. just the license header is messing on that one file

sdk/src/test/java/io/dapr/client/MockHttpResponse.java
```diff
@@ -0,0 +1,68 @@
+package io.dapr.client;
```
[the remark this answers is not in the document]
artur-ciocanu: Good catch. Fixed

cicoyle (approved):
lgtm. Thanks for adding the `READ_TIMEOUT` for consistency

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
```diff
@@ -265,61 +461,4 @@ public void validateExceptionParsing() {
...
-   *   <li>Execute client getState for Key=key2</li>
-   *   <li>Execute client deleteState for Key=key2</li>
-   *   <li>Block for deleteState call.</li>
-   *   <li>Block for getState for Key=key2 and Assert they 2 was not found.</li>
-   * </ol>
-   *
-   * @throws IOException - Test will fail if any unexpected exception is being thrown
-   */
-  @Test()
-  public void testCallbackCalledAtTheExpectedTimeTest() throws IOException {
```
artursouza, line 289: This test is very important. Why was it deleted?

artursouza (changes requested):

artursouza:
Thanks a lot @artur-ciocanu. I have two questions above. The point about the test is that in Project Reactor, all actions need to be lazily resolved - we validate it by making sure nothing is invoked until `.block()` is called. This is basic expectations.

Now, there is a major piece of work that we can consider (another PR), which is to move the SDK away from Project Reactor and use CompletableFuture only (https://www.baeldung.com/java-asynchronous-programming#completableFuture). I am not in favor of using another 3rd party async API anymore :)

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -244,13 +247,13 @@ public Mono<Response> invokeApi(
...
     // fromCallable() is needed so the invocation does not happen early, causing a hot mono.
-    return Mono.fromCallable(() -> doInvokeApi(method, pathSegments, urlParameters, content, headers, context))
-        .flatMap(f -> Mono.fromFuture(f));
+    return Mono.fromCallable(() -> doInvokeApi(method, headers, pathSegments, urlParameters, content, context))
+        .flatMap(Mono::fromFuture);
   }
 
   /**
-   * Shutdown call is not necessary for OkHttpClient.
-   * @see OkHttpClient
```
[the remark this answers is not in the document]
artur-ciocanu: Yes, this is still true for HttpClient, HttpClient doesn't an explicit `close()` method.

sdk/src/test/java/io/dapr/client/DaprHttpTest.java
artur-ciocanu: Since we are using `HttpClient` and `CompletableFuture`, I couldn't really mimic the OkHttpClient MockInterceptor behavior. I will give another try.

Thanks for feedback!

artur-ciocanu:
@artursouza I have added back the test related to multiple Monos, could you please take another look. Thank you!

salaboy (approved):

artur-ciocanu:
@artursouza could you please approve and merge. Thank you!

artursouza (approved):

Issue #1219, opened 2025-02-20
Upgrade Spring Boot version to make sure it is OSS supported


salaboy:
## Expected Behavior

Use an OSS supported Spring Boot version: https://spring.io/projects/spring-boot#support 

## Actual Behavior

We are using a version that is not longer OSS Supported.

## Release Note

RELEASE NOTE:

PR #1220, opened 2025-02-20
Upgrading spring boot version to 3.4.3

salaboy:
# Description

Upgrading Spring Boot version to keep aligned with supported OSS version: https://spring.io/projects/spring-boot#support

I think we are ok to remove 3.2.x from the validation pipeline. 

## Issue reference

Please reference the issue this PR will close: #1219

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/build.yml
```diff
@@ -32,6 +32,10 @@ jobs:
             spring-boot-version: 3.3.0
             spring-boot-display-version: 3.3.x
             experimental: false
```
artur-ciocanu, line 34: I this file shouldn't we remove the 3.2.x from the GitHub action since it is no longer supported.

pom.xml
```diff
@@ -45,7 +45,7 @@
     <junit-bom.version>5.8.2</junit-bom.version>
     <snakeyaml.version>2.0</snakeyaml.version>
     <testcontainers.version>1.20.0</testcontainers.version>
```
artur-ciocanu, line 47: Shouldn't we use 3.3.x here instead of the latest and greatest. It is better to have the previous supported version for all our code, instead of relying on the latest Spring Boot version.

artur-ciocanu (approved):
@salaboy overall looks good, but please take a look at my comments.

.github/workflows/build.yml
salaboy: I think we should

pom.xml
salaboy: I thought about this too.. but the end of support for is 2025-06-30. That only gives us 3 months of support. I would rather start testing 3.4.x and get ready to move to 3.5.x when it is final. I think if we are wise enough, we will align with their release cadence..

salaboy:
@artursouza @cicoyle can you please retrigger the pipelines? the issue here is not related to the changes done in spring boot.

artur-ciocanu:
I have retriggered the failing job

salaboy:
I am running the failing test locally, and it works as expected. I do no know why this is failing here, as it is not related to Spring Boot: https://gist.github.com/salaboy/9cb1066a6e8df7521251e6865f035f1e

salaboy:
@cicoyle @artur-ciocanu @artursouza I found the issue.. fingers crossed now!

snicoll:
> this is definitely a spring boot issue..

What was the issue?

salaboy:
@snicoll super strange an offending dependency in the classpath was breaking the spring boot app bootstrap face between versions 3.3.7 and 3.3.8+. So everything worked correctly in 3.3.7  and in 3.3.8+ the booting event listeners were getting stuck for some reason. 
Once I found the offending dependency that who knows why was there.. everything worked as expected even with 3.4.3.

snicoll:
>  the offending dependency 

What offending dependency? 

>  the booting event listeners were getting stuck for some reason.

Can you take a thread dump?

salaboy:
@snicoll I don't think we need to worry about this.. it was a very nasty dependency that shouldn't been there -> https://github.com/dapr/java-sdk/pull/1220/files#diff-f473a1c5f096a7eca3a87df5cc735319907cb7cf860a8cfd65207c869c93c6c8L79 I am currently rushing to get this release out, so I will not have much time to investigate deeper into this. 

But for me, this was interesting -> https://github.com/dapr/java-sdk/pull/1220/files#diff-f473a1c5f096a7eca3a87df5cc735319907cb7cf860a8cfd65207c869c93c6c8R31
- otel upgrade 
- logback upgrade

salaboy:
@artur-ciocanu @artursouza I need some help here.. I found an inconsistency between 1.15.0 and 1.14.4 about how the status about the connection with the placement service is happening.. this was breaking the `DaprContainerIT.testPlacement()` test. As you can see now the test is using `1.14.4` as by default with latest all the test moved on to `1.15.0` yesterday when the new tag was created. If we use 1.15.0 this test will fail, as placement is not connected unless the actor state store is configured (as per my understanding)

artursouza:
> @artur-ciocanu @artursouza I need some help here.. I found an inconsistency between 1.15.0 and 1.14.4 about how the status about the connection with the placement service is happening.. this was breaking the `DaprContainerIT.testPlacement()` test. As you can see now the test is using `1.14.4` as by default with latest all the test moved on to `1.15.0` yesterday when the new tag was created. If we use 1.15.0 this test will fail, as placement is not connected unless the actor state store is configured (as per my understanding)

If this is the case, then it is a bug.

sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprContainerIT.java
```diff
@@ -59,9 +65,10 @@ public class DaprContainerIT {
   private static final String KEY = "my-key";
   private static final String PUBSUB_NAME = "pubsub";
   private static final String PUBSUB_TOPIC_NAME = "topic";
+  private static final String APP_FOUND_MESSAGE_PATTERN = ".*application discovered on port 8081.*";
 
   @Container
-  private static final DaprContainer DAPR_CONTAINER = new DaprContainer("daprio/daprd")
+  private static final DaprContainer DAPR_CONTAINER = new DaprContainer("daprio/daprd:1.14.4")
```
artursouza, line 71: This is better since it makes the test not break because of a bug in runtime.

artursouza (dismissed):

sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprContainerIT.java
artursouza: Ideally, this is coming from a centralized configuration that is updated once in the repository.

artur-ciocanu: This has been fixed in a previous PR. The image tag for Dapr Container has been extracted to a constant.

artur-ciocanu:
@artursouza and @cicoyle we need your review one more time. Thank you!

sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprContainerIT.java
```diff
@@ -129,9 +136,26 @@ public void testStateStore() throws Exception {
 
   @Test
   public void testPlacement() throws Exception {
```
artur-ciocanu, line 138: @salaboy I don't think this is required. I have ensured that DaprContainer uses a constant `IMAGE_TAG` and it is set to `1.14.1` for now. We can have a separate PR to upgrade to `1.15.1`.

salaboy:
@artursouza we need this to get in before releasing

artur-ciocanu:
@artursouza and @cicoyle it seems that some checks are stuck since we have removed the SB 3.2 from the the GitHub Actions. All the other build steps are 🟢. Could you please take a look and approve and merge.

Thank you!

artur-ciocanu (approved):
LGTM

artursouza (approved):

PR #1234, opened 2025-03-06
Skipping docs and deploy on release

salaboy:
# Description

There is no need to deploy to maven central the sb examples

## Issue reference

Please reference the issue this PR will close: #

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/spring-boot-examples/consumer-app/pom.xml
+++ b/spring-boot-examples/consumer-app/pom.xml
@@ -93,6 +93,22 @@
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


spring-boot-examples/consumer-app/pom.xml
```diff
@@ -93,6 +93,22 @@
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
cicoyle, line 97: Is this configuration needed in the `consumer-app` and `producer-app` poms if we have `maven.deploy.skip` set in the parent `spring-boot-examples/pom.xml`? The parent configuration should be inherited by all child modules automatically.

cicoyle (approved):

PR #1241, opened 2025-03-07
Fix orphaned channel

cicoyle:
Fix orphaned channel for dapr client builder. [Noticed in the failed publish step here](https://github.com/dapr/java-sdk/actions/runs/13682584960/job/38330466331#step:8:348)


salaboy:
@cicoyle but this looks ok with me

PR #1242, opened 2025-03-07
Update dapr v1.15

cicoyle:
want to ensure the release doesnt encounter any issues. Going to bump versions bc I was seeing the actor timer issue that runtime v1.15.2 fixed. Saw v1.15.3 is out, so going to use latest.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
       GOPROXY: https://proxy.golang.org
       JDK_VER: ${{ matrix.java }}
       DAPR_CLI_VER: 1.15.0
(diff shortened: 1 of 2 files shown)


artur-ciocanu (approved):
LGTM, let's see if all the tests are 🟢

artur-ciocanu:
@cicoyle I have run the "Auto Validate Examples" multiple times, but it keeps failing, I am not sure what has changed in the 1.15 but apparently the output is no longer the same since the validation fails.

artur-ciocanu:
@cicoyle this PR seems to be 🟢 as well, I think it is safe to merge it. Thank you.

cicoyle:
I re-ran it like twice, so I think we are good here as well. Will merge

Issue #1248, opened 2025-03-10
Fix Flaky CI Build Test Failure: ConsumerApp


cicoyle:
Fix flaky test showing in CI:

https://github.com/dapr/java-sdk/actions/runs/13773015167/job/38515789368?pr=1246

```
Error:  Errors: 
[...]

PR #1273, opened 2025-03-21
Use Network.newNetwork() to create a Docker Network

siri-varma:
# Description

From the logs it is not able to find "dapr-network"
The other nonflaky tests inside testcontainers just use Network.newNetwork(). So changed this to use the same pattern

## Issue reference

Please reference the issue this PR will close: #1248 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


siri-varma:
@artur-ciocanu , @salaboy , @cicoyle I just saw other non flaky tests use this pattern of creating a network. So changed this test to match that pattern as well. Let me know your thoughts

salaboy:
@siri-varma we cannot use `newNetwork()` because we should be able to reuse the network if it already exists. This work locally, but for some reason it does fail in CI and it is becoming annoying, so I am investigating with @eddumelendez from testcontainers.

salaboy (changes requested):
I dont think we can do this.

spring-boot-examples/consumer-app/src/test/java/io/dapr/springboot/examples/consumer/DaprTestContainersConfig.java
```diff
@@ -36,31 +36,7 @@ public class DaprTestContainersConfig {
...
-
-    List<com.github.dockerjava.api.model.Network> networks = DockerClientFactory.instance().client().listNetworksCmd()
-            .withNameFilter("dapr-network").exec();
-    if (networks.isEmpty()) {
-      Network.builder().createNetworkCmdModifier(cmd -> cmd.withName("dapr-network")).build().getId();
-      return defaultDaprNetwork;
-    } else {
-      return defaultDaprNetwork;
-    }
+    return Network.newNetwork();
```
salaboy, line 39: @siri-varma this would have been so much easier.. but we cannot do that if we want to run both applications locally to reuse the network.

salaboy:
@siri-varma can you please close this one?

siri-varma:
Synced up with @salaboy and looks like he is already working on it. Will close this PR.

PR #1274, opened 2025-03-23
Support Podman Desktop on Spring Boot examples

salaboy:
# Description

Adding extra instrucitons if you are running with Podman

## Issue reference

Please reference the issue this PR will close: #1275

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/spring-boot-examples/kubernetes/README.md
+++ b/spring-boot-examples/kubernetes/README.md
@@ -13,6 +13,17 @@ with a local container registry, so we can push our container images to it. This
 ./kind-with-registry.sh
 ```
 
+**Note**: If you are using Podman Desktop, instead of Docker you need to run the following command to enable insecure registries:
+
+```
+read -r -d '' registry_conf <<EOF
+[[registry]]
+location = "localhost:5001"
+insecure = true
+EOF
+podman machine ssh --username=root sh -c 'cat > /etc/containers/registries.conf.d/local.conf' <<<$registry_conf
+```
+
 Once you have the cluster up and running you can install Dapr: 
 
 ```bash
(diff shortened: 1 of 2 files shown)


mcruzdev (approved):

artur-ciocanu (approved):
@salaboy it looks good to me, but the broader question is, is this Spring Boot Examples specific or Podman can be used for other examples as well.

cicoyle (changes requested):
Thanks for adding this @salaboy -As @artur-ciocanu mentioned, I think this applies to other examples as well for a consistent experience. 

Can you make a note in the [root README](https://github.com/dapr/java-sdk/blob/master/README.md#pre-requisites-1) too about Podman support? Im fine if you want to link to the podman commands here from the root README, but we should at least mention it as an option I think.

salaboy:
@artur-ciocanu @cicoyle we can create a separate issue for that.. there is no way that I will test other examples with Podman, as that will be a much larger work than this. 
I don't think we need to reference this on the Java SDK, as if Dapr works with Podman is more about dapr/dapr (runtime) itself. 
My recommendation is to leave that for another time/issue.

cicoyle:
> @artur-ciocanu @cicoyle we can create a separate issue for that.. there is no way that I will test other examples with Podman, as that will be a much larger work than this. I don't think we need to reference this on the Java SDK, as if Dapr works with Podman is more about dapr/dapr (runtime) itself. My recommendation is to leave that for another time/issue.

Can you open an issue for podman support beyond springboot?

cicoyle (approved):

PR #1288, opened 2025-03-27
Remove flaky test using Testcontainers and a more specific test approach

mcruzdev:
# Description

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/dapr-spring/dapr-spring-boot-autoconfigure/src/test/java/io/dapr/spring/boot/autoconfigure/client/DaprClientPropertiesTest.java
+++ b/dapr-spring/dapr-spring-boot-autoconfigure/src/test/java/io/dapr/spring/boot/autoconfigure/client/DaprClientPropertiesTest.java
@@ -1,3 +1,16 @@
+/*
+ * Copyright 2024 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package io.dapr.spring.boot.autoconfigure.client;
 
 import org.assertj.core.api.SoftAssertions;
--- a/dapr-spring/dapr-spring-boot-autoconfigure/src/test/java/io/dapr/spring/boot/autoconfigure/client/DaprClientPropertiesTest.java
+++ b/dapr-spring/dapr-spring-boot-autoconfigure/src/test/java/io/dapr/spring/boot/autoconfigure/client/DaprClientPropertiesTest.java
@@ -1,0 +1,76 @@
+package io.dapr.spring.boot.autoconfigure.client;
+
+import org.assertj.core.api.SoftAssertions;
+import org.junit.Test;
+import org.junit.jupiter.api.DisplayName;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+
+public class DaprClientPropertiesTest {
+
+
+  private final ApplicationContextRunner runner = new ApplicationContextRunner()
+      .withUserConfiguration(EnableDaprClientProperties.class);
+
+  @Test
+  @DisplayName("Should create DaprClientProperties correctly through constructor")
+  public void shouldCreateDaprClientPropertiesCorrectly() {
+
+    DaprClientProperties properties = new DaprClientProperties(
+        "http://localhost", "localhost", 3500, 50001
+    );
+
+    SoftAssertions.assertSoftly(softAssertions -> {
+      softAssertions.assertThat(properties.getGrpcEndpoint()).isEqualTo("localhost");
+      softAssertions.assertThat(properties.getHttpEndpoint()).isEqualTo("http://localhost");
+      softAssertions.assertThat(properties.getHttpPort()).isEqualTo(3500);
+      softAssertions.assertThat(properties.getGrpcPort()).isEqualTo(50001);
+    });
+  }
+
+  @Test
+  @DisplayName("Should create DaprClientProperties correctly through setters")
+  public void shouldSetDaprClientPropertiesCorrectly() {
+
+    DaprClientProperties properties = new DaprClientProperties();
+
+    properties.setGrpcEndpoint("localhost");
+    properties.setGrpcPort(50001);
+    properties.setHttpEndpoint("http://localhost");
+    properties.setHttpPort(3500);
+
+    SoftAssertions.assertSoftly(softAssertions -> {
+      softAssertions.assertThat(properties.getGrpcEndpoint()).isEqualTo("localhost");
+      softAssertions.assertThat(properties.getHttpEndpoint()).isEqualTo("http://localhost");
+      softAssertions.assertThat(properties.getHttpPort()).isEqualTo(3500);
+      softAssertions.assertThat(properties.getGrpcPort()).isEqualTo(50001);
+    });
+  }
+
+  @Test
+  @DisplayName("Should map DaprClient properties correctly")
+  public void shouldMapDaprClientProperties() {
+
+    runner.withSystemProperties(
+        "dapr.client.http-endpoint=http://localhost",
+        "dapr.client.http-port=3500",
+        "dapr.client.grpc-endpoint=localhost",
+        "dapr.client.grpc-port=50001"
+    ).run(context -> {
+      DaprClientProperties properties = context.getBean(DaprClientProperties.class);
+      SoftAssertions.assertSoftly(softAssertions -> {
+        softAssertions.assertThat(properties.getGrpcEndpoint()).isEqualTo("localhost");
+        softAssertions.assertThat(properties.getHttpEndpoint()).isEqualTo("http://localhost");
+        softAssertions.assertThat(properties.getHttpPort()).isEqualTo(3500);
+        softAssertions.assertThat(properties.getGrpcPort()).isEqualTo(50001);
+      });
+
+    });
+
+  }
+
+  @EnableConfigurationProperties(DaprClientProperties.class)
+  static class EnableDaprClientProperties {
+
+  }
+}
--- a/sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
@@ -13,153 +13,204 @@ limitations under the License.
 
 package io.dapr.it.resiliency;
 
+import com.github.tomakehurst.wiremock.client.WireMock;
+import com.github.tomakehurst.wiremock.junit5.WireMockTest;
+import eu.rekawek.toxiproxy.Proxy;
+import eu.rekawek.toxiproxy.ToxiproxyClient;
+import eu.rekawek.toxiproxy.model.ToxicDirection;
+import eu.rekawek.toxiproxy.model.toxic.Latency;
+import eu.rekawek.toxiproxy.model.toxic.Timeout;
 import io.dapr.client.DaprClient;
 import io.dapr.client.DaprClientBuilder;
-import io.dapr.client.DaprClientImpl;
 import io.dapr.client.resiliency.ResiliencyOptions;
-import io.dapr.it.BaseIT;
-import io.dapr.it.DaprRun;
-import io.dapr.it.ToxiProxyRun;
+import io.dapr.config.Properties;
+import io.dapr.exceptions.DaprException;
+import io.dapr.testcontainers.DaprContainer;
+import io.dapr.testcontainers.DaprLogLevel;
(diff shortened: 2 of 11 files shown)


salaboy:
@mcruzdev what's the idea here? just to replicate the same tests (the flaky one) but with Testcontainers and Toxiproxy managed by testcontainers?

mcruzdev:
> @mcruzdev what's the idea here? just to replicate the same tests (the flaky one) but with Testcontainers and Toxiproxy managed by testcontainers?

The idea is to use Testcontainers and to test in a different way.

salaboy:
As soon as it test the same thing, we can use this to create a plan to move all tests to use test containers

mcruzdev:
> As soon as it test the same thing, we can use this to create a plan to move all tests to use test containers

I changed the test to be more specific, testing each configuration separated.

salaboy:
@artur-ciocanu did you had time to check this out? This is definitely our first step to move our integration tests to testcontainers .. and it is looking really good IMO. Do you have any concerns about this ?

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+
+    stubFor(any(urlMatching("/([a-z1-9]*)"))
+        .willReturn(aResponse().withBody("[]").withStatus(200)));
+
+    configureFor("localhost", WIREMOCK_PORT);
+  }
 
-  private static final int MAX_RETRIES = -1;  // Infinity
+  @BeforeAll
+  static void configure() throws IOException {
```
cicoyle, line 113: Shall we move the Toxiproxy container setup to the `DaprContainerIT` file since it's common infrastructure logic? This would make it available for all integration tests.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+    ToxiproxyClient toxiproxyClient = new ToxiproxyClient(TOXIPROXY.getHost(), TOXIPROXY.getControlPort());
+    proxy =
+        toxiproxyClient.createProxy("dapr", "0.0.0.0:8666", "dapr:3500");
```
cicoyle, line 116: We should also consider randomizing the ports to avoid conflicts between tests.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+        .willReturn(aResponse().withStatus(204).withFixedDelay(ms1000)));
+
+    stubFor(any(urlMatching("/([a-z1-9]*)"))
+        .willReturn(aResponse().withBody("[]").withStatus(200)));
+
+    configureFor("localhost", WIREMOCK_PORT);
+  }
 
-  private static final int MAX_RETRIES = -1;  // Infinity
+  @BeforeAll
```
cicoyle, line 112: Should we have an `@AfterAll` for cleanup?

cicoyle:
> @artur-ciocanu did you had time to check this out? This is definitely our first step to move our integration tests to testcontainers .. and it is looking really good IMO. Do you have any concerns about this ?

I agree, this is a great start and will be great in the long run for standardization across IT tests.

@salaboy - Please cc the listed java approvers + maintainers for input as opposed to individual GH handles, as that way, when we add more approvers and maintainers we don't leave anyone out.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
mcruzdev: @cicoyle the `8666` port is the internal port of Toxiproxy container and the `3500` is the default configured by DaprContainer class. I think we can't to do it for Toxiproxy and to Dapr we need a change on container, am I right @salaboy?

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
mcruzdev: Good catch! For TestContainers I saw that the annotation `@TestContainers` take care about it for us, and for [Wiremock](https://wiremock.org/docs/junit-jupiter/#wiremock-server-lifecycle) too.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
mcruzdev: Maybe we can have:

* `BaseResiliencyContainers` with Dapr and Toxiproxy containers
* `BaseDaprContainer` with Dapr only

Both having Wiremock configurations, WDYT? IMHO we do not need an abstraction right now, but for `ActorSdkResiliencyIT` we will ned.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.configureFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
+import static com.github.tomakehurst.wiremock.client.WireMock.verify;
+import static io.dapr.it.resiliency.SdkResiliencyIT.WIREMOCK_PORT;
+import static io.dapr.it.testcontainers.DaprContainerConstants.IMAGE_TAG;
```
artur-ciocanu, line 60: Could we use a simpler import like:
```
import static org.assertj.core.api.Assertions.assertThat;
```

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+      .withAppName("dapr-app")
+      .withAppPort(WIREMOCK_PORT)
+      .withDaprLogLevel(DaprLogLevel.DEBUG)
+      .withLogConsumer(new Slf4jLogConsumer(LoggerFactory.getLogger("dapr-logs")))
+      .withAppHealthCheckPath("/actuator/health")
+      .withAppChannelAddress("host.testcontainers.internal")
+      .withNetworkAliases(List.of("dapr"))
+      .withNetwork(NETWORK);
+
+  @Container
```
artur-ciocanu, line 85: I have created a class named `DaprContainerConstants`, but I think we can rename it to `ContainerConstants` and add have Dapr container image tag and ToxiProxy and anything else we might need for our Testcontainers based tests.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+    stubFor(any(urlMatching("/actuator/health"))
+        .willReturn(aResponse().withBody("[]").withStatus(200)));
+
+    stubFor(any(urlMatching("/dapr/subscribe"))
+        .willReturn(aResponse().withBody("[]").withStatus(200)));
+
+    stubFor(get(urlMatching("/dapr/config"))
+        .willReturn(aResponse().withBody("[]").withStatus(200)));
+
+    // create a stub for simulating dapr sidecar with timeout of 1000 ms
```
artur-ciocanu, line 101: We try to avoid using `var`. Also I saw that you already use `Duration`, so I would recommend using `Duration.ofMillis(..)` and when a raw values is required use `...toMillis()`, this way it is more readable.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
artur-ciocanu: @mcruzdev and @cicoyle this is just the first test, so I think the best we can do is to make sure that we can extract the container tag to a file with constants representing different containers. I already suggested that in a previous comment.

Once we migrate more ITs we will see if a pattern emerges and we can decide what we want to pull into a common base class.

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
-  private static final int MAX_RETRIES = -1;  // Infinity
+  @BeforeAll
+  static void configure() throws IOException {
+    ToxiproxyClient toxiproxyClient = new ToxiproxyClient(TOXIPROXY.getHost(), TOXIPROXY.getControlPort());
+    proxy =
+        toxiproxyClient.createProxy("dapr", "0.0.0.0:8666", "dapr:3500");
+  }
 
-  private static DaprClient daprClient;
+  @BeforeEach
```
artur-ciocanu, line 119: I think this method name is misleading, I would name it `beforeEach`

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+  public void setDaprProperties() {
+    configStub();
+    org.testcontainers.Testcontainers.exposeHostPorts(WIREMOCK_PORT);
+  }
 
-  private static ToxiProxyRun toxiProxyRun;
+  @Test
```
artur-ciocanu, line 125: I love it, the `@DisplayName` is awesome!

sdk-tests/src/test/java/io/dapr/it/resiliency/SdkResiliencyIT.java
```diff
@@ -13,153 +13,213 @@
...
+  @DisplayName("should throw exception when the configured timeout exceeding waitForSidecar's timeout")
+  public void testSidecarWithoutTimeout() {
```
artur-ciocanu, line 127: As I mentioned we avoid using `var` in our codebase. Also a similar comment, could we please use `Duration` so it is clear that we are dealing with millis.

artur-ciocanu: Minor nitpick, I don't think comments like `act`, `arrange`, `assert` add any value. `I have seen it in multiple places.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -70,9 +72,11 @@ public class DaprContainer extends GenericContainer<DaprContainer> {
   private Integer appPort;
   private String appHealthCheckPath;
   private boolean shouldReusePlacement;
```
artur-ciocanu, line 74: It is not clear, why we need this change. As far as I saw in the IT tests, the network alias is not required. I am not against enhancing `DaprContainer` with network alias, but I think it can be done in a separate PR where we could explain what is the use case and probably and a few unit tests to improve coverage.

artur-ciocanu (changes requested):
@mcruzdev awesome work, love it! I have left some comments could you please address them and I'll take another look.

Thank you ❤️

salaboy:
@cicoyle how do you do that ? Meaning mentioning all maintainers ?

cicoyle:
> @cicoyle how do you do that ? Meaning mentioning all maintainers ?

yes, as I want to ensure the appropriate folk are not left out. Here are the tags for the appropriate community members in this SDK: @dapr/approvers-java-sdk and @dapr/maintainers-java-sdk.

But tagging is not necessary immediately after opening PRs and issues bc those groups already get emails about it bc approvers and maintainers are already listed as reviewers automatically
image

I would say tagging is more for after a day or more goes by without review or input.

salaboy:
@mcruzdev ok.. it seems that only codecov is missing, this is quite important so we can maintain the coverage over time. Can you check the report and see if you can cover more paths to get over the line?

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
artur-ciocanu: @mcruzdev I see that you are using `.withNetwork(NETWORK)` for `ToxiproxyContainer`, hence my question regarding network alias.

As I mentioned we can add support for network alias in a separate PR and add the necessary code coverage to `DaprContainer`. Once we have support for network alias we can revisit this IT and use network alias instead of network.

artur-ciocanu (approved):
@mcruzdev thanks for addressing all the comments, it looks great!

dapr-spring/dapr-spring-boot-autoconfigure/src/test/java/io/dapr/spring/boot/autoconfigure/client/DaprClientPropertiesTest.java
```diff
@@ -0,0 +1,76 @@
+package io.dapr.spring.boot.autoconfigure.client;
```
cicoyle, line 1: One last thing, mind adding the header at the top of this file? Then lgtm and Ill approve and merge

cicoyle (approved):

Issue #1305, opened 2025-04-11
Solve ContainerLaunch Timed out exception


mcruzdev:
## Release Note

RELEASE NOTE:

cicoyle:
@mcruzdev - mind providing additional information like links to CI logs or logs from local runs when opening issues? Any additional information would be super helpful for others to understand the issue here, be able to reproduce it, and have context on it 🙏🏻

mathecruz:
Hi @cicoyle, I think we can close this one

PR #1306, opened 2025-04-11
Update DaprActorsIT wait strategy

mcruzdev:
# Description

## Issue reference

Please reference the issue this PR will close: #1305

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprActorsIT.java
```diff
@@ -53,7 +53,7 @@ public class DaprActorsIT {
   private static final Random RANDOM = new Random();
   private static final int PORT = RANDOM.nextInt(1000) + 8000;
 
-  private static final String ACTORS_MESSAGE_PATTERN = ".*Actor API level in the cluster has been updated to 10.*";
+  private static final String ACTORS_MESSAGE_PATTERN = ".*Found Subscription.*";
```
salaboy, line 56: @mcruzdev I agree for now.. but this is getting very inconsistent.. I keep finding edge cases..

mcruzdev: What edge cases @salaboy? I tried here n times and the tests are passing.

salaboy: @mcruzdev the problem is that if there are no subscriptions registered by the application this test way of waiting will not work. But for this, if you are creating a subscription for sure its ok, but it is not generic enough for all the test.

salaboy:
@mcruzdev can you please check the codecov and resolve the conflicts?

sdk-tests/src/test/java/io/dapr/it/testcontainers/DaprActorsIT.java
mcruzdev: I agree with you, for this case we really need to wait the subscription (see liine 91). But we can not use this approach as generic as well.

cicoyle (changes requested):
Can you elaborate as to why this change was made? Why change the message here from `actor runtime started` to `found subscription` message pattern for actors?

mcruzdev:
> Can you elaborate as to why this change was made? Why change the message here from `actor runtime started` to `found subscription` message pattern for actors?

I tested locally here some times and I think we can close this pull request.

Issue #1314, opened 2025-04-18
Dapr SDK 1.14.0 Issue with Slash "\" in URI


rahulpoddar-fyndna:
**Setup:**
dapr-sdk: 1.14.0
dapr runtime: 1.15.4

**Facing issue with leading slash in url.While it worked with earlier SDK versions, it's failing with new sdk 1.14.0.**

We observed that including a leading slash (/) in the endpoint causes the call to fail with an HTTP 400 error.

Example:
When ServiceA invokes ServiceB with the endpoint defined as:
udirBaseURI: "/udir/translate" → The call fails with HTTP 400.
udirBaseURI: "udir/translate" → The call succeeds.
Please note this is observed when side is mocked and using dapr sdk as is for service to service invocation.

Another observation is as follows
**Analysis:**

Dapr has recently changed the implementation of service invocation replacing OkHttpClient with Java's HttpClient. Please refer to the commit: **https://github.com/dapr/java-sdk/pull/1218/commits/cdda128626af0fe14eed23c7c35a9057eff19cec**

**The newly added createQuery method tries to encode the query parameters before trying to create java URI**

eg uri=/serviceB-point/serviceB-address/fetch-by-id gets encoded to uri=%2FserviceB-point%2FserviceB-address%2Ffetch-by-id gets

**Java URI creation, apparently encodes the query parameters again before sending the http request** 

eg uri=%2FserviceB-point%2FserviceB-address%2Fetch-by-id gets encoded again to uri=%252FserviceB-point%252FserviceB-address%252Ffetch-by-id

**On the server side, Spring decodes the query parameter just once which is causing the problem.**

eg uri=%252F2FserviceB-point%252F2FserviceB-address%252Ffetch-by-id gets decoded to uri=%2FserviceB-point%2FserviceB-address%2Ffetch-by-id which should have been /serviceB-point/serviceB-address/fetch-by-id

**The impact is also seen in path parameter scenarios.**

vikram-mahadevkar:
- In our application we have code which uses DaprClient.invokeMethod method provided by Dapr SDK to make remote call to other apps.
- As during JUnit run sidecar is not available, we create mock HTTP endpoints using WireMock library which mimics endpoints provides by Dapr sidecar and return configured mock response.
- In our application we use leading slash(/) in all endpoints while calling DaprClient.invokeMethod method.
- These JUnits were working properly with Dapr SDK version 1.13.3.
[...]

PR #1320, opened 2025-04-21
Fix URL building logic

artur-ciocanu:
# Description

This is a tiny PR that fixes the URL building logic in the DaprHttp functionality. This is trying to fix an issue where we tried to replace the OkHttp URL builder with something that uses JDK URI.

## Issue reference

Please reference the issue this PR will close: #1314 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


artur-ciocanu:
@rahulpoddar-fyndna and @vikram-mahadevkar this is the fix for #1314. We hope to release it as soon as possible.

artur-ciocanu:
@cicoyle @mcruzdev @salaboy could you please take a look. This might need a hot fix.

sdk-tests/src/test/java/io/dapr/it/methodinvoke/http/GH_1314_IT.java
```diff
@@ -0,0 +1,87 @@
+package io.dapr.it.methodinvoke.http;
```
cicoyle, line 1: Can we add a header to this file and change the name to something more intuitive like query_encoding_IT or query_param_encoding_IT or similar? We don't typically name files based on the GH issue bc future developers wont know what is in the test without an extra step of pulling up GH and looking up the issue, although I appreciate you trying to reference it that way.

sdk/src/main/java/io/dapr/client/DaprHttp.java
```diff
@@ -363,6 +373,11 @@ private static String createQuery(Map<String, List<String>> urlParameters) {
 
     for (Map.Entry<String, List<String>> entry : urlParameters.entrySet()) {
       String key = entry.getKey();
+
+      if (key == null || key.isEmpty()) {
+        continue; // Skip empty keys
```
siri-varma, line 378: Quick question, when would urlParameters/pathSegments have empty or null values ?

sdk-tests/src/test/java/io/dapr/it/methodinvoke/http/GH_1314_IT.java
```diff
@@ -0,0 +1,87 @@
...
+          requestData,
+          HttpExtension.POST,
+          String.class
+      ).block();
+
+      assertEquals("Hello from say-hello app!", response);
+    } catch (Exception e) {
+      fail("Exception occurred: " + e.getMessage());
+    } finally {
+      try {
```
siri-varma, line 79: assertDoesNotThrow(() -> client.close());

Maybe this could be simplified to assertDoesNotThrow(() -> client.close());

sdk-tests/src/test/java/io/dapr/it/methodinvoke/http/GH_1314_IT.java
artur-ciocanu: Done

siri-varma (approved):
LGTM

artur-ciocanu:
@cicoyle @salaboy and @mcruzdev I have figured what was the problem with my setup. Here is the breakdown:
- To be able to use named parameters in Spring we have to either adjust the Java compiler arguments or make sure that we use `@RequestParam("uri") String uri`, otherwise this results in runtime exceptions and eventually ends up as HTTP 500.
- On my local box `mDNS` is not working, hence service to service was failing. I have no idea how to fix it, the way I managed to overcome this is by using SQLite name resolution.

After I have went through all this hurdles, I was able to confirm that my fix works and there are no more double encodings.

CC: @rdosifyndna

artur-ciocanu:
@cicoyle and @salaboy since the name resolution based on mDNS can be such a pain should we use SQLite name resolution in our ITs. I think this should make our tests a little bit more predictable and reliable.

Any thoughts?

cicoyle (approved):

PR #1332, opened 2025-05-05
Users/svegiraju/pom cleanup 3

siri-varma:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


artur-ciocanu:
@siri-varma it seems that there are so missing classes after your POM changes and there some `NoSuchMethodError` messages in logs.

siri-varma:
@artur-ciocanu Yes. Still working on fixing this

siri-varma:
@artur-ciocanu @cicoyle This PR is ready to be reviewed. Thank you

salaboy:
@siri-varma to be clear.. I am not against this.. I am just thinking a bit long term how this will work out.

siri-varma:
> @artur-ciocanu do you have any comments about this? Because the Dapr SDK is split into two big things (modules and sdk-tests), I believe that it was ok to do dependency management separately, as dependencies needed in the sdk-tests are not needed in the SDK itself (like toxy proxy). But if we are moving towards a world where we have all the SDK tests as testcontainer tests, this makes sense.

@salaboy quick comment, my intention of moving all of them to the base pom is so that we can have one single place to manage all the dependencies. However, this need not affect the SDK/testcontainer test choices is what I feel.

Let me know if I am missing something.

examples/pom.xml
```diff
@@ -82,6 +82,16 @@
       <artifactId>opentelemetry-exporter-zipkin</artifactId>
       <version>${opentelemetry.version}</version>
     </dependency>
```
artur-ciocanu, line 84: Could we make sure that OpenTelemetry deps are listed one after the other. We should use version properties consistently and avoid inline version. Also we have the same list of deps in sdk-tests, could you please make sure that these are in sync, otherwise tests have a set of deps and examples a different set of deps, I am referring to OTel and Zipkin reporter.

pom.xml
```diff
@@ -230,6 +235,132 @@
         <artifactId>spring-boot-testcontainers</artifactId>
         <version>${springboot.version}</version>
       </dependency>
```
artur-ciocanu, line 237: As far as I can see some dependencies are using version properties, some are using hard coded versions. Could we make sure that we use ONLY version properties.

Also could we please make sure that we use the right dependency type like `test` for dependencies that are not intended to be used in the final distribution of Java SDK.

artur-ciocanu: BTW the same comment related to version properties applies to Maven plugins, not just dependencies.

artur-ciocanu (changes requested):
@siri-varma the PR looks great, I have left some comments, could you please take a look.

Also could you please provide a high level overview what are you trying to achieve with dependency cleanup PRs. I see the value in extracting version properties, however I am not following why we have to move them from respective places to parent POM.

examples/pom.xml
siri-varma: I investigated a bit more into the OTel version and Zipkin versions. 

The code in the examples is using an older version of OTel and Zipkin and some of these methods have been deprecated in the newer versions.

So, we will have to migrate the code and then only we can upgrade the version.

pom.xml
siri-varma: Yep. The test scopes are already declared in the dapr-tests/pom.xml.

The best practice is to declare the dependency in the parent pom and let the user of the dependency decide the scope (sdk-tests in this case).

siri-varma:
@artur-ciocanu/ @salaboy / @dapr/maintainers-java-sdk addressed all the comments. Thank you

siri-varma:
@artur-ciocanu / @dapr-java-sdk-reviewer / @salaboy  This PR is ready to be reviewed

siri-varma:
@artur-ciocanu / @dapr-java-sdk-reviewer / @salaboy 

Just a gentle reminder about this PR

PR #1333, opened 2025-05-05
Compensation example for Workflows

cicoyle:
Add a compensation workflow example


salaboy:
@cicoyle I was looking at this PR. The example doesn't seem to show that compensations are executed, right? Was this already implemented in dotnet? Do we have any reference of how it was implemented in dotnet? Did you used that implementation as inspiration for this PR?

cicoyle:
> @cicoyle I was looking at this PR. The example doesn't seem to show that compensations are executed, right? Was this already implemented in dotnet? Do we have any reference of how it was implemented in dotnet? Did you used that implementation as inspiration for this PR?

This is a new pattern entirely. There are no examples in any other sdk or in the dapr docs. Its a simple example to illustrate how users can use the compensation workflow pattern.

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -540,3 +541,116 @@ The log from client:
...
+  - "Flight booking completed: Flight booked successfully"
+  - "Starting Activity: io.dapr.examples.workflows.compensation.BookHotelActivity"
+  - "Simulating hotel booking process..."
+  - "Activity completed with result: Hotel booked successfully"
+  - "Hotel booking completed: Hotel booked successfully"
+  - "Starting Activity: io.dapr.examples.workflows.compensation.BookCarActivity"
+  - "Forcing Failure to trigger compensation for activity: io.dapr.examples.workflows.compensation.BookCarActivity"
+  - "******** executing compensation logic ********"
+  - "Activity failed: Task 'io.dapr.examples.workflows.compensation.BookCarActivity' (#2) failed with an unhandled exception: Failed to book car"
+  - "Error during compensation: The orchestrator is blocked and waiting for new inputs. This Throwable should never be caught by user code."
```
siri-varma, line 624: Would it be better to catch the OrchestratorException separately. Because that is expected and is not an `error during compensation`

siri-varma (approved):
Overall LGTM.

I think most of the files are missing a new line. If we could add that, the PR is good to go.

examples/src/main/java/io/dapr/examples/workflows/compensation/BookTripWorkflow.java
```diff
@@ -0,0 +1,75 @@
...
+                            case "CancelHotel":
+                                String hotelCancelResult = ctx.callActivity(CancelHotelActivity.class.getName(), null, String.class).await();
+                                ctx.getLogger().info("Hotel cancellation completed: {}", hotelCancelResult);
+                                break;
+                            case "CancelFlight":
+                                String flightCancelResult = ctx.callActivity(CancelFlightActivity.class.getName(), null, String.class).await();
+                                ctx.getLogger().info("Flight cancellation completed: {}", flightCancelResult);
+                                break;
+                        }
+                    } catch (Exception ex) {
```
javier-aliaga, line 67: Can we use retries in the Compensation activities? If so would catching the Exception break the retry mechanism?

artur-ciocanu (dismissed):
@cicoyle awesome job! I love the docs related to Mechanical Markdown.

My only concern is the comment from @javier-aliaga regarding retries. Either we mention this is the code comment or add some note, so developers are aware how handling the exception might affect the retry behavior.

examples/src/main/java/io/dapr/examples/workflows/README.md
cicoyle: good point, let me change that. It would be cleaner to not catch [that exception since its part of the normal workflow runtime control flow](https://github.com/dapr/durabletask-java/blob/main/client/src/main/java/io/dapr/durabletask/TaskOrchestrationExecutor.java#L1307). Ive switched to catching the TaskFailedException.

cicoyle:
I will push a commit or 2 addressing the comments ^

examples/src/main/java/io/dapr/examples/workflows/compensation/BookTripWorkflow.java
cicoyle: yes, retries work for compensation activities. I have locally 2 new IT tests Ill clean up and push for confirmation. Catching `TaskFailedExceptions` works, but users must avoid catching a generic exception as that will interfere with compensation logic execution

cicoyle:
Will circle back to the build failures

salaboy (dismissed):
not 100% sure about this but here is the label

Issue #1353, opened 2025-05-13
SSL CA support for the Java SDK


salaboy:
## Expected Behavior

Similar to temporal we should be able to set the CA key and path:

```
 String clientKeyPath = getenv("TEMPORAL_MTLS_PRIVATE_KEY_PATH");
        String clientCertPath = getenv("TEMPORAL_MTLS_CERT_PATH");

        try {
            // Generate an SSL context
            InputStream clientCertInputStream = new FileInputStream(clientCertPath);
            InputStream clientKeyInputStream = new FileInputStream(clientKeyPath);
            SslContext sslContext = SimpleSslContextBuilder.forPKCS8(clientCertInputStream, clientKeyInputStream).build();

            // Set the Service Stub options (SSL context and gRPC endpoint)
            WorkflowServiceStubsOptions stubsOptions = WorkflowServiceStubsOptions
                .newBuilder()
                .setSslContext(sslContext)
                .setTarget(gRPCEndpoint)
                .build();
```

Example: https://github.com/temporalio/documentation/blob/main/sample-apps/java/client/cloudserver-client-sample/src/main/java/clientsample/YourCallerApp.java 

## Actual Behavior

If the app is behind a proxy with their own trust store the SDK needs to support setting the SSL context. 

## Release Note

RELEASE NOTE:

cicoyle:
https://github.com/dapr/java-sdk/pull/1361

PR #1360, opened 2025-05-14
feat: Support for GRPC ssl

javier-aliaga:
# Description

## Issue reference

Please reference the issue this PR will close: #1353

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


famarting:
this adds client cert and key, which I think is for MTLs

can we add another option to just set a CA, i.e if you were going to just expose tls on selfsigned certs, so we can still use tls

salaboy:
> this adds client cert and key, which I think is for MTLs
> 
> can we add another option to just set a CA, i.e if you were going to just expose tls with selfsigned certs, so we can still use tls

We were looking at what Temporal is adding, as the customer shows us that code described in this issue: https://github.com/dapr/java-sdk/issues/1353

salaboy:
I am looking at the APIS available now to see what else can be set.

salaboy:
I am looking at the APIs it looks like we just need to set: `trustManager()`
```
 ChannelCredentials credentials = TlsChannelCredentials.newBuilder()
                .trustManager()
                .build();
```

PR #1382, opened 2025-05-21
chore: Add grpc keepalives

javier-aliaga:
# Description

_Please explain the changes you've made_

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #1383, opened 2025-05-21
[1.15] cherrypick grpc keepalives (#1382)

cicoyle:
[cherrypick this PR](https://github.com/dapr/java-sdk/pull/1382)

### DIFF
--- a/sdk/src/main/java/io/dapr/config/Properties.java
+++ b/sdk/src/main/java/io/dapr/config/Properties.java
@@ -146,6 +146,38 @@ public class Properties {
       null);
 
   /**
+   * GRPC enable keep alive.
+   */
+  public static final Property<Boolean> GRPC_ENABLE_KEEP_ALIVE = new BooleanProperty(
+      "dapr.grpc.enableKeepAlive",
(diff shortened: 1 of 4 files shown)


artur-ciocanu (approved):
@cicoyle as far as I understand this is a cherry pick, however the Java properties are a little bit inconsistent with other Java property names.

Here we are using a mix of dot notation and camel case, while for other GRPC properties we use dot notation like:
- `dapr.grpc.tls.insecure`
- `dapr.grpc.tls.ca.path`

We should try to be consistent, it helps with documentation and long term maintenance.

cicoyle:
> @cicoyle as far as I understand this is a cherry pick, however the Java properties are a little bit inconsistent with other Java property names.
> 
> Here we are using a mix of dot notation and camel case, while for other GRPC properties we use dot notation like:
> 
> * `dapr.grpc.tls.insecure`
> * `dapr.grpc.tls.ca.path`
> 
> We should try to be consistent, it helps with documentation and long term maintenance.

I'll open follow up PRs to address the inconsistency there, good catch 👍🏻 since you approved this, Ill proceed with this since its a cherrypick then open 2 followup PRs to fix it and add docs👍🏻

PR #1384, opened 2025-05-21
System Properties + Env Var Docs

cicoyle:
Updates gRPC keepalive property names to use consistent naming convention: `DAPR_GRPC_ENABLE_KEEP_ALIVE` for environment variables and `dapr.grpc.enable.keep.alive` for system properties. 

Add docs for users to know what they can set


artur-ciocanu (approved):
Love it, really, really nice, especially the docs part.

PR #1413, opened 2025-06-11
Adding examples to Spring Boot (duration, zoneddatetime and suspend/resume)

salaboy:
# Description

Adding examples to validate that ZonedDateTime works 

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #1409

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #1419, opened 2025-06-11
Pull out createTimer logic

cicoyle:
[Pull out logic from this PR to add it to the release quicker](https://github.com/dapr/java-sdk/pull/1413). Examples should still be added in the original PR, but the logic should be rebased since it will go thru in this PR


siri-varma (approved):
LGTM

Issue #1422, opened 2025-06-13
Add more details to WorkflowTaskRetryHandler


TheForbiddenAi:
## Describe the proposal
Add task name and task input to WorkflowTaskRetryContext to allow for additional error-handling capabilities in the RetryHandler.

siri-varma:
@TheForbiddenAi thank you for the issue and pr. Could you pleae clarify what "additional error-handling capabilities" mean ?

TheForbiddenAi:
@siri-varma Yea so let's say you want to control the retry handlers flow based off of a field in the input object for the activity. As of now you can only access the input values of workflow and not the activities or child workflows. 

Additionally, if you want to wanted to be able to have one central retry handler but still want to have some small amount of activity specific logic. There is no way to know what activity is currently executing. The workaround for this now would be to create a whole new retry handler and chain it to the original retry handler. But that will lead to duplicate code if you have code in the original retry handler that has to be executed before the activity specific handler.

siri-varma:
Thank you for the explanation. If my understanding is right, you want to change the retry handler based on input to the activity.

Within a workflow, when activities are triggered, the input to the activity is something we send isn't it ? So, my assumption is we already have access to the input or am I missing something

TheForbiddenAi:
Yes, that is right. However, we don't have access to the activity/child workflows input because we pass the parent workflow context not the activity context. We do have access to the parent workflow input via the context.

siri-varma:
If you don't mind, can you add an example of your use case in the PR ? That will make things more clear.

May be here https://github.com/dapr/java-sdk/tree/master/examples/src/main/java/io/dapr/examples/workflows

TheForbiddenAi:
@siri-varma I added an example. Another thing to note is that having access to the input at retry handle time would allow the input to be extracted and sent elsewhere (i.e. an incident report). I still need to update the readme and will do that in a little bit

TheForbiddenAi:
readme has been updated

PR #1423, opened 2025-06-13
Add taskName and input to retry handler

TheForbiddenAi:
# Description

Add `taskName` and `input` to `WorkflowTaskRetryContext`
Changed `workflowContext` type in `WorkflowTaskRetryContext` to `WorkflowContext`

## Issue reference

Please reference the issue this PR will close: #1422 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-workflows/src/main/java/io/dapr/workflows/WorkflowTaskRetryContext.java
+++ b/sdk-workflows/src/main/java/io/dapr/workflows/WorkflowTaskRetryContext.java
@@ -14,16 +14,17 @@ limitations under the License.
 package io.dapr.workflows;
 
 import io.dapr.workflows.client.WorkflowFailureDetails;
-import io.dapr.workflows.runtime.DefaultWorkflowContext;
 
 import java.time.Duration;
 
 public class WorkflowTaskRetryContext {
 
-  private final DefaultWorkflowContext workflowContext;
+  private final WorkflowContext workflowContext;
   private final int lastAttemptNumber;
   private final WorkflowFailureDetails lastFailure;
   private final Duration totalRetryTime;
+  private final String taskName;
+  private final Object input;
 
   /**
    * Constructor for WorkflowTaskRetryContext.
@@ -32,28 +33,34 @@ public class WorkflowTaskRetryContext {
    * @param lastAttemptNumber The number of the previous attempt
    * @param lastFailure The failure details from the most recent failure
    * @param totalRetryTime The amount of time spent retrying
+   * @param taskName The name of the task
+   * @param input The input of the task
    */
   public WorkflowTaskRetryContext(
-          DefaultWorkflowContext workflowContext,
+          WorkflowContext workflowContext,
           int lastAttemptNumber,
           WorkflowFailureDetails lastFailure,
-          Duration totalRetryTime) {
+          Duration totalRetryTime,
+          String taskName,
+          Object input) {
     this.workflowContext = workflowContext;
     this.lastAttemptNumber = lastAttemptNumber;
     this.lastFailure = lastFailure;
     this.totalRetryTime = totalRetryTime;
+    this.taskName = taskName;
+    this.input = input;
   }
 
   /**
    * Gets the context of the current workflow.
    *
    * <p>The workflow context can be used in retry handlers to schedule timers (via the
-   * {@link DefaultWorkflowContext#createTimer} methods) for implementing delays between retries. It can also be
-   * used to implement time-based retry logic by using the {@link DefaultWorkflowContext#getCurrentInstant} method.
+   * {@link WorkflowContext#createTimer} methods) for implementing delays between retries. It can also be
+   * used to implement time-based retry logic by using the {@link WorkflowContext#getCurrentInstant} method.
    *
    * @return the context of the parent workflow
    */
-  public DefaultWorkflowContext getWorkflowContext() {
+  public WorkflowContext getWorkflowContext() {
     return this.workflowContext;
   }
 
@@ -85,4 +92,21 @@ public class WorkflowTaskRetryContext {
     return this.totalRetryTime;
   }
 
+  /**
+   * Gets the name of the task.
+   *
+   * @return the name of the task
+   */
+  public String getTaskName() {
+    return taskName;
+  }
+
+  /**
+   * Gets the input of the task.
+   *
+   * @return the task's input
+   */
+  public Object getInput() {
+    return input;
+  }
 }
--- a/sdk-workflows/src/main/java/io/dapr/workflows/runtime/DefaultWorkflowContext.java
+++ b/sdk-workflows/src/main/java/io/dapr/workflows/runtime/DefaultWorkflowContext.java
@@ -164,7 +164,7 @@ public class DefaultWorkflowContext implements WorkflowContext {
    * {@inheritDoc}
    */
   public <V> Task<V> callActivity(String name, Object input, WorkflowTaskOptions options, Class<V> returnType) {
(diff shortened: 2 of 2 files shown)


examples/src/main/java/io/dapr/examples/workflows/retryhandler/DemoRetryHandler.java
```diff
@@ -0,0 +1,48 @@
...
+public class DemoRetryHandler implements WorkflowTaskRetryHandler {
+
+  @Override
+  public boolean handle(WorkflowTaskRetryContext retryContext) {
+    WorkflowContext workflowContext = retryContext.getWorkflowContext();
+    Logger logger = retryContext.getWorkflowContext().getLogger();
+    Object input = retryContext.getInput();
+    String taskName = retryContext.getTaskName();
+
+    if(taskName.equalsIgnoreCase(FailureActivity.class.getName())) {
```
siri-varma, line 34: Interesting, so we can have multiple if blocks

if (taskName.equals..(ChildWorkf1.class)) {
  // do something
}

if (taskName.equals..(ChildWorkf2.class)) {
  // do something
}

if (taskName.equals..(ChildActivity1.class)) {
  // do something
}

Wouldn't it be more maintainable to colocate the workflow or activity with its corresponding retry logic? While I understand the code duplication part, centralizing all retry handling in a single utility class does not seem optimal

@cicoyle , @artur-ciocanu , @salaboy I would like to know your thoughts too.

TheForbiddenAi: So in the case of multiple activites, I would recommend breaking them up into separate retry handlers that call each other (See [chain of responsibility design pattern](https://refactoring.guru/design-patterns/chain-of-responsibility)). The main benefit of chaining handlers is that you can reuse logic from prior handlers to reduce code duplication. For example

```java
class HandlerOne implements WorkflowTaskRetryHandler {

  private WorkflowTaskRetryHandler nextHandler;
  
  public boolean handle(WorkflowTaskRetryContextcontext) {
    // handler logic here
    if(nextHandler != null) {
      return nextHandler.handle(context);
    }
    return true;
  }

  void setNextHandler(WorkflowTaskRetryHandler handler) {
     this.nextHandler = handler;
  }
}
```

In a workflow for instance, it would look like this (HandlerTwo follows same layout as HandlerOne above)
```java
public class TestWorkflow implements Workflow {

  @Override
  public WorkflowStub create() {
    return context -> {
      Logger logger = context.getLogger();
      logger.info("Starting RetryWorkflow: {}", context.getName());

      HandlerOne handlerOne = new HandlerOne();
      HandlerTwo handlerOne = new HandlerTwo();

      handlerOne.setNextHandler(handlerTwo);
      WorkflowTaskOptions taskOptions = new WorkflowTaskOptions(handlerOne);

      logger.info("RetryWorkflow is calling Activity: {}", FailureActivity.class.getName());
      Instant currentTime = context.getCurrentInstant();
      Instant result = context.callActivity(FailureActivity.class.getName(), currentTime, taskOptions, Instant.class).await();

      logger.info("RetryWorkflow finished with: {}",  result);
      context.complete(result);
    };
  }

}
```
Ideally, you would define an interface or abstract class that defines any common methods/logic (i.e. setNextHandler) when using this pattern, but for the sake of simplicity, I left it out in the example above

TheForbiddenAi: > Additionally, I'm curious about scenarios where retries are driven solely by the names of workflow/activities. Could you share an example where that pattern would be more effective?

Perhaps, you want to have variable backoff times depending on what each activity is doing. Additionally, if you know the name of the activity, you know the type of its input. Also, having the name is useful when it comes to tracking what task is being retried in the logs, especially if you are running multiple activities concurrently.

Another possible situation is: let's say you have Task A. This task makes a request to a service under the assumption that this service has access to other data. If this service doesn't have this data, then Task A fails. To recover from this failure, you may want to spin off a new activity, we'll call this Task B, to do some action that creates that data. 

In this hypothetical situation, you would only want Task B to be called if Task A fails to prevent duplicating data. Once Task B completes, we could then return to our retry handler to retry Task A automatically.

siri-varma:
@TheForbiddenAi thank you for the insightful explanation. The PR looks good to me. Thank you

siri-varma (approved):
LGTM

TheForbiddenAi:
Workflow failure is not related to these changes. They should pass on rerun

salaboy:
@TheForbiddenAi @siri-varma I have some issues with this idea. I believe that the task name needs to be propagated to the runtime to keep track of the task name. Not having this at the runtime level, will make it impossible from the orchestrator (the dapr sidecar) to keep track of what is executed, as this is only being added on the SDK side. 
This also brings inconsistency with other SDKs.

TheForbiddenAi:
I'm confused at what you mean by the task name having to be propagated to the runtime. This wouldn't change at. I don't see how this affects the knowledge of the orchestrator either. The orchestrator is what calls the retry handler in the first place. We just wrap the objects to not expose the durable task ones. This is just adding extra information, which we already have available to us when the context objects get instantiated at runtime, to that object.

As far as inconsistencies go, I don't think other SDKs have added support for durabletask's retry handler.

TheForbiddenAi:
@salaboy I'm confused at what you mean by the task name having to be propagated to the runtime. This wouldn't change at. I don't see how this affects the knowledge of the orchestrator either. The orchestrator is what calls the retry handler in the first place. We just wrap the objects to not expose the durable task ones. This is just adding extra information, which we already have available to us when the context objects get instantiated at runtime, to that object.

As far as inconsistencies go, I don't think other SDKs have added support for durabletask's retry handler.

examples/src/main/java/io/dapr/examples/workflows/retryhandler/DemoRetryHandler.java
siri-varma: @TheForbiddenAi I was revisiting this comment earlier this morning and had a couple of thoughts:

Regarding logs: if you include a logger statement inside each activity, it should help you identify which activity is currently running, right?

As for the second scenario, one approach could be to structure the workflow so that it starts with Activity A. If Activity A fails due to missing data, the workflow can then trigger Activity B. Once B completes successfully, you can start Activity A within the same workflow.

Also, the way I see retries, they are for transient failures but in your case Activity A fails with permanent failure

Let me know your thoughts.

TheForbiddenAi: So for logs, if you have multiple workflows running or multiple activities running asynchronously. It would be very difficult to differentiate logs in the retry handler from one another. 

As for second scenario, yes that is a valid approach. However it would require a try catch and then manually calling the activity again, meaning if you had multiple of these scenarios for the same activity, you would have to chain try catches which would not be great from a readability and maintainability standpoint

salaboy:
What I meant is that the task name is something more general than just the
retry mechanism, hence just adding the task name for the retry mechanism
feels wrong to me.

 - Blog: http://salaboy.com <http://salaboy.wordpress.com>
 - Github user: http://github.com/salaboy
 - Twitter: http://twitter.com/salaboy

 - Mauricio "Salaboy" Salatino -

TheForbiddenAi:
I still don't really see the problem. You could say the same thing about the orchestrator context that the original durable task retry context passes. I don't see the issue with providing the user more information when retrying to allow for more complex logic. However, if you truly think this is a blocker for this PR I will take out the task name from the context.

PR #1439, opened 2025-07-04
Update DaprClientGrpcTest.java

love12yadav:
This PR fixes a bug in the `getStateThenDelete` test method.

Previously, the test incorrectly assumed that `getState(...).block()` would return the state value even after the state had been deleted. However, the `Mono` was not blocked until *after* the delete operation, meaning the value was no longer present in the state store.

### Fix Summary:
- Delayed the `.block()` on `getState` for key2 until after `deleteState` is executed.
- Updated the mock to simulate that the state is removed after deletion.
- Changed the test to correctly assert that the value is `null` after deletion.

This aligns the test with the correct reactive execution behavior in Dapr's Java SDK.

cc @salaboy @siri-varma

### DIFF
--- a/sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
+++ b/sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
@@ -1376,51 +1376,68 @@ public class DaprClientGrpcTest {
    *
    * @throws Exception - Test will fail if any unexpected exception is being thrown.
    */
+@Test
+public void getStateThenDelete() throws Exception {
+  String etag = "ETag1";
+  String key1 = "key1";
+  String expectedValue1 = "Expected state 1";
+  String key2 = "key2";
+  String expectedValue2 = "Expected state 2";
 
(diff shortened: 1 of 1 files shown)


sdk/src/test/java/io/dapr/client/DaprClientGrpcTest.java
```diff
@@ -1376,51 +1376,68 @@ private <T> State<T> buildStateKey(T value, String key, String etag, Map<String,
    *
    * @throws Exception - Test will fail if any unexpected exception is being thrown.
    */
```
artur-ciocanu, line 1378: @love12yadav could you please use the right formatting. Currently it seem that the whole `getStateThenDelete` is incorrectly formatted.

artur-ciocanu (changes requested):
@love12yadav thanks a lot for your contribution the PR looks good, but the formatting is a little bit off. Could you please fix the formatting and I'll approve the PR.

Thank you.

PR #1468, opened 2025-07-25
Feat Cross App CallActivity

cicoyle:
Add support for cross app call activity calls.

### DIFF
--- a/examples/src/main/java/io/dapr/examples/workflows/README.md
+++ b/examples/src/main/java/io/dapr/examples/workflows/README.md
@@ -53,6 +53,7 @@ Those examples contain the following workflow patterns:
 4. [External Event Pattern](#external-event-pattern)
 5. [Child-workflow Pattern](#child-workflow-pattern)
 6. [Compensation Pattern](#compensation-pattern)
+7. [Cross-App Pattern](#cross-app-pattern)
 
 ### Chaining Pattern
 In the chaining pattern, a sequence of activities executes in a specific order.
@@ -653,6 +654,212 @@ Key Points:
 3. The workflow ensures that all resources are properly cleaned up even if the process fails
 4. Each activity simulates work with a short delay for demonstration purposes
 
+
+### Cross-App Pattern
+
+The cross-app pattern allows workflows to call activities that are hosted in different Dapr applications. This is useful for microservices architectures where activities are distributed across multiple services, service mesh scenarios, or multi-tenant applications where activities are isolated by app ID.
+
+The `CrossAppWorkflow` class defines the workflow. It demonstrates calling activities in different apps using the `appId` parameter in `WorkflowTaskOptions`. See the code snippet below:
+```java
+public class CrossAppWorkflow implements Workflow {
+  @Override
+  public WorkflowStub create() {
+    return ctx -> {
+      System.out.println("=== WORKFLOW STARTING ===");
+      ctx.getLogger().info("Starting CrossAppWorkflow: " + ctx.getName());
+      System.out.println("Workflow name: " + ctx.getName());
+      System.out.println("Workflow instance ID: " + ctx.getInstanceId());
+
+      String input = ctx.getInput(String.class);
+      ctx.getLogger().info("CrossAppWorkflow received input: " + input);
+      System.out.println("Workflow input: " + input);
+
+      // Call an activity in another app by passing in an active appID to the WorkflowTaskOptions
+      ctx.getLogger().info("Calling cross-app activity in 'app2'...");
+      System.out.println("About to call cross-app activity in app2...");
+      String crossAppResult = ctx.callActivity(
+          App2TransformActivity.class.getName(),
+          input,
+          new WorkflowTaskOptions("app2"),
+          String.class
+      ).await();
+
+      // Call another activity in a different app
+      ctx.getLogger().info("Calling cross-app activity in 'app3'...");
+      System.out.println("About to call cross-app activity in app3...");
+      String finalResult = ctx.callActivity(
+          App3FinalizeActivity.class.getName(),
+          crossAppResult,
+          new WorkflowTaskOptions("app3"),
+          String.class
+      ).await();
+      ctx.getLogger().info("Final cross-app activity result: " + finalResult);
+      System.out.println("Final cross-app activity result: " + finalResult);
+      
+      ctx.getLogger().info("CrossAppWorkflow finished with: " + finalResult);
+      System.out.println("=== WORKFLOW COMPLETING WITH: " + finalResult + " ===");
+      ctx.complete(finalResult);
+    };
+  }
+}
+```
+
+The `App2TransformActivity` class defines an activity in app2 that transforms the input string. See the code snippet below:
+```java
+public class App2TransformActivity implements WorkflowActivity {
+  @Override
+  public Object run(WorkflowActivityContext ctx) {
+    System.out.println("=== App2: TransformActivity called ===");
+    String input = ctx.getInput(String.class);
+    System.out.println("Input: " + input);
+    
+    // Transform the input
+    String result = input.toUpperCase() + " [TRANSFORMED BY APP2]";
+    
+    System.out.println("Output: " + result);
+    return result;
+  }
+}
+```
+
+The `App3FinalizeActivity` class defines an activity in app3 that finalizes the processing. See the code snippet below:
+```java
+public class App3FinalizeActivity implements WorkflowActivity {
+  @Override
+  public Object run(WorkflowActivityContext ctx) {
+    System.out.println("=== App3: FinalizeActivity called ===");
+    String input = ctx.getInput(String.class);
+    System.out.println("Input: " + input);
+    
+    // Finalize the processing
+    String result = input + " [FINALIZED BY APP3]";
+    
+    System.out.println("Output: " + result);
+    return result;
+  }
+}
+```
+
+**WorkflowTaskOptions Constructors for Cross-App Calls:**
+
+The `WorkflowTaskOptions` class supports several constructors for cross-app calls:
+
+```java
+// App ID only
+new WorkflowTaskOptions("app2")
+
+// Retry policy + app ID
+new WorkflowTaskOptions(retryPolicy, "app2")
+
+// Retry handler + app ID
+new WorkflowTaskOptions(retryHandler, "app2")
+
+// All parameters
+new WorkflowTaskOptions(retryPolicy, retryHandler, "app2")
+```
+
+**Key Features:**
+- **Cross-app activity calls**: Call activities in different Dapr applications specifying the appID in the WorkflowTaskOptions
+- **WorkflowTaskOptions with appId**: Specify which app should handle the activity
+- **Combined with retry policies**: Use app ID along with retry policies and handlers
+- **Error handling**: Works the same as local activity calls
+
+**Requirements:**
+- Multiple Dapr applications running with different app IDs
+- Activities registered in the target applications
+- Proper Dapr workflow runtime configuration
+
+**Important Limitations:**
+- **Cross-app calls are currently supported for activities only**
+- **Child workflow cross-app calls (suborchestration) are NOT supported**
+- The app ID must match the Dapr application ID of the target service
+
+**Running the Cross-App Example:**
+
+This example requires running multiple Dapr applications simultaneously. You'll need to run the following commands in separate terminals:
+
+1. **Start the main workflow worker (crossapp-worker):**
+```sh
+dapr run --app-id crossapp-worker --resources-path ./components/workflows --dapr-grpc-port 50001 --log-level=debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.CrossAppWorker
+```
+
+2. **Start app2 worker (handles App2TransformActivity):**
+```sh
+dapr run --app-id app2 --resources-path ./components/workflows --dapr-grpc-port 50002 --log-level=debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.App2Worker
+```
+
+3. **Start app3 worker (handles App3FinalizeActivity):**
+```sh
+dapr run --app-id app3 --resources-path ./components/workflows --dapr-grpc-port 50003 --log-level=debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.App3Worker
+```
+
+4. **Run the workflow client:**
+```sh
+java -Djava.util.logging.ConsoleHandler.level=FINE -Dio.dapr.durabletask.level=FINE -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.CrossAppWorkflowClient "Hello World"
+```
+
+<!-- STEP
+name: Run Cross-App Pattern workflow
+match_order: none
+output_match_mode: substring
+expected_stdout_lines:
+  - "=== Starting Cross-App Workflow Client ==="
+  - "Input: Hello World"
+  - "Created DaprWorkflowClient successfully"
+  - "Attempting to start new workflow..."
+  - "Started a new cross-app workflow with instance ID:"
+  - "Waiting for workflow completion..."
+  - "Workflow instance with ID:"
+  - "completed with result: HELLO WORLD [TRANSFORMED BY APP2] [FINALIZED BY APP3]"
+  - "=== WORKFLOW STARTING ==="
+  - "Starting CrossAppWorkflow:"
+  - "Workflow name:"
+  - "Workflow instance ID:"
+  - "CrossAppWorkflow received input: Hello World"
+  - "Workflow input: Hello World"
+  - "Calling cross-app activity in 'app2'..."
+  - "About to call cross-app activity in app2..."
+  - "=== App2: TransformActivity called ==="
+  - "Input: Hello World"
+  - "Output: HELLO WORLD [TRANSFORMED BY APP2]"
+  - "Calling cross-app activity in 'app3'..."
+  - "About to call cross-app activity in app3..."
+  - "=== App3: FinalizeActivity called ==="
+  - "Input: HELLO WORLD [TRANSFORMED BY APP2]"
+  - "Output: HELLO WORLD [TRANSFORMED BY APP2] [FINALIZED BY APP3]"
+  - "Final cross-app activity result:"
+  - "CrossAppWorkflow finished with:"
+  - "=== WORKFLOW COMPLETING WITH:"
+background: true
+sleep: 60
+timeout_seconds: 60
+-->
+
+<!-- END_STEP -->
+
+**Expected Output:**
+
+The client will show:
+```text
+=== Starting Cross-App Workflow Client ===
+Input: Hello World
+Created DaprWorkflowClient successfully
+Attempting to start new workflow...
+Started a new cross-app workflow with instance ID: 001113f3-b9d9-438c-932a-a9a9b70b9460
+Waiting for workflow completion...
+Workflow instance with ID: 001113f3-b9d9-438c-932a-a9a9b70b9460 completed with result: HELLO WORLD [TRANSFORMED BY APP2] [FINALIZED BY APP3]
+```
+
+The workflow demonstrates:
+1. The workflow starts in the main worker (crossapp-worker)
+2. Calls an activity in 'app2' using cross-app functionality
+3. Calls an activity in 'app3' using cross-app functionality
+4. The workflow completes with the final result from all activities
+
+This pattern is particularly useful for:
+- Microservices architectures where activities are distributed across multiple services
+- Service mesh scenarios where different apps handle different types of activities
+- Multi-tenant applications where activities are isolated by app ID
 
 ### Suspend/Resume Pattern
 
--- a/examples/src/main/java/io/dapr/examples/workflows/crossapp/App2TransformActivity.java
+++ b/examples/src/main/java/io/dapr/examples/workflows/crossapp/App2TransformActivity.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2025 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.examples.workflows.crossapp;
+
+import io.dapr.workflows.WorkflowActivity;
+import io.dapr.workflows.WorkflowActivityContext;
+
+/**
+ * TransformActivity for App2 - transforms input to uppercase.
+ * This activity is called cross-app from the main workflow.
+ */
+public class App2TransformActivity implements WorkflowActivity {
+    @Override
+    public Object run(WorkflowActivityContext context) {
+        String input = context.getInput(String.class);
+        System.out.println("=== App2: TransformActivity called ===");
+        System.out.println("Input: " + input);
+        String result = input.toUpperCase() + " [TRANSFORMED BY APP2]";
+        System.out.println("Output: " + result);
+        return result;
+    }
+}
--- a/examples/src/main/java/io/dapr/examples/workflows/crossapp/App3FinalizeActivity.java
+++ b/examples/src/main/java/io/dapr/examples/workflows/crossapp/App3FinalizeActivity.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2025 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.examples.workflows.crossapp;
+
+import io.dapr.workflows.WorkflowActivity;
+import io.dapr.workflows.WorkflowActivityContext;
+
+/**
+ * FinalizeActivity for App3 - adds final processing.
+ * This activity is called cross-app from the main workflow.
+ */
+public class App3FinalizeActivity implements WorkflowActivity {
+    @Override
+    public Object run(WorkflowActivityContext context) {
+        String input = context.getInput(String.class);
+        System.out.println("=== App3: FinalizeActivity called ===");
+        System.out.println("Input: " + input);
+        String result = input + " [FINALIZED BY APP3]";
+        System.out.println("Output: " + result);
+        return result;
+    }
+}
--- a/examples/src/main/java/io/dapr/examples/workflows/crossapp/CrossAppWorkflow.java
+++ b/examples/src/main/java/io/dapr/examples/workflows/crossapp/CrossAppWorkflow.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2025 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.examples.workflows.crossapp;
+
+import io.dapr.workflows.Workflow;
+import io.dapr.workflows.WorkflowStub;
+import io.dapr.workflows.WorkflowTaskOptions;
+
+/**
+ * Example workflow that demonstrates cross-app activity calls.
+ * This workflow calls activities in different apps using the appId parameter.
+ */
+public class CrossAppWorkflow implements Workflow {
+  @Override
+  public WorkflowStub create() {
+    return ctx -> {
+      System.out.println("=== WORKFLOW STARTING ===");
+      ctx.getLogger().info("Starting CrossAppWorkflow: " + ctx.getName());
+      System.out.println("Workflow name: " + ctx.getName());
+      System.out.println("Workflow instance ID: " + ctx.getInstanceId());
+
+      String input = ctx.getInput(String.class);
+      ctx.getLogger().info("CrossAppWorkflow received input: " + input);
+      System.out.println("Workflow input: " + input);
+
+      // Call an activity in another app by passing in an active appID to the WorkflowTaskOptions
+      ctx.getLogger().info("Calling cross-app activity in 'app2'...");
+      System.out.println("About to call cross-app activity in app2...");
+      String crossAppResult = ctx.callActivity(
+          App2TransformActivity.class.getName(),
+          input,
+          new WorkflowTaskOptions("app2"),
+          String.class
+      ).await();
+
+      // Call another activity in a different app
+      ctx.getLogger().info("Calling cross-app activity in 'app3'...");
+      System.out.println("About to call cross-app activity in app3...");
+      String finalResult = ctx.callActivity(
+          App3FinalizeActivity.class.getName(),
+          crossAppResult,
+          new WorkflowTaskOptions("app3"),
+          String.class
+      ).await();
+      ctx.getLogger().info("Final cross-app activity result: " + finalResult);
+      System.out.println("Final cross-app activity result: " + finalResult);
+      
+      ctx.getLogger().info("CrossAppWorkflow finished with: " + finalResult);
+      System.out.println("=== WORKFLOW COMPLETING WITH: " + finalResult + " ===");
+      ctx.complete(finalResult);
+    };
+  }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,6 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <grpc.version>1.69.0</grpc.version>
     <protobuf.version>3.25.5</protobuf.version>
-    <protocCommand>protoc</protocCommand>
     <dapr.proto.baseurl>https://raw.githubusercontent.com/dapr/dapr/v1.15.4/dapr/proto</dapr.proto.baseurl>
     <dapr.sdk.version>1.16.0-SNAPSHOT</dapr.sdk.version>
     <dapr.sdk.alpha.version>0.16.0-SNAPSHOT</dapr.sdk.alpha.version>
--- a/sdk-autogen/pom.xml
+++ b/sdk-autogen/pom.xml
@@ -22,7 +22,7 @@
     <protobuf.input.directory>${project.build.directory}/proto</protobuf.input.directory>
     <maven.deploy.skip>false</maven.deploy.skip>
     <grpc.version>1.69.0</grpc.version>
-    <protocCommand>protoc</protocCommand>
+    <protocCommand>cassie</protocCommand>
     <protobuf.version>3.25.5</protobuf.version>
   </properties>
 
@@ -132,7 +132,6 @@
               <goal>run</goal>
             </goals>
             <configuration>
-              <protocCommand>protoc</protocCommand>
               <protocVersion>${protobuf.version}</protocVersion>
               <protocArtifact>com.google.protobuf:protoc:3.25.5</protocArtifact>
               <addProtoSources>inputs</addProtoSources>
--- a/sdk-workflows/pom.xml
+++ b/sdk-workflows/pom.xml
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>io.dapr</groupId>
       <artifactId>durabletask-client</artifactId>
-      <version>1.5.6</version>
+      <version>1.5.7-SNAPSHOT</version>
     </dependency>
     <!--
     manually declare durabletask-client's jackson dependencies
--- a/examples/src/main/java/io/dapr/examples/workflows/crossapp/App2Worker.java
+++ b/examples/src/main/java/io/dapr/examples/workflows/crossapp/App2Worker.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2025 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.examples.workflows.crossapp;
+
+import io.dapr.workflows.runtime.WorkflowRuntime;
+import io.dapr.workflows.runtime.WorkflowRuntimeBuilder;
+
+/**
+ * App2 Worker - registers only the TransformActivity.
+ * This app will handle cross-app activity calls from the main workflow.
+ */
+public class App2Worker {
+  
+  public static void main(String[] args) throws Exception {
+    System.out.println("=== Starting App2Worker ===");
+    // Register the Workflow with the builder
+    WorkflowRuntimeBuilder builder = new WorkflowRuntimeBuilder()
+            .registerActivity(App2TransformActivity.class);
+
+      // Build and start the workflow runtime
+      try (WorkflowRuntime runtime = builder.build()) {
+        System.out.println("App2 is ready to receive cross-app activity calls...");
+        runtime.start();
+      }
+    }
+}
--- a/examples/src/main/java/io/dapr/examples/workflows/crossapp/App3Worker.java
+++ b/examples/src/main/java/io/dapr/examples/workflows/crossapp/App3Worker.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2025 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.examples.workflows.crossapp;
+
+import io.dapr.workflows.runtime.WorkflowRuntime;
+import io.dapr.workflows.runtime.WorkflowRuntimeBuilder;
+
+/**
+ * App3 Worker - registers only the FinalizeActivity.
+ * This app will handle cross-app activity calls from the main workflow.
+ */
+public class App3Worker {
+  
+  public static void main(String[] args) throws Exception {
+    System.out.println("=== Starting App3Worker ===");
+    // Register the Workflow with the builder
+    WorkflowRuntimeBuilder builder = new WorkflowRuntimeBuilder()
+            .registerActivity(App3FinalizeActivity.class);
+
+      // Build and start the workflow runtime
+      try (WorkflowRuntime runtime = builder.build()) {
+        System.out.println("App3 is ready to receive cross-app activity calls...");
+        runtime.start();
+      }
+    }
+}
--- a/examples/src/main/java/io/dapr/examples/workflows/crossapp/CrossAppWorker.java
+++ b/examples/src/main/java/io/dapr/examples/workflows/crossapp/CrossAppWorker.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2025 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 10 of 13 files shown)


sdk-workflows/pom.xml
```diff
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>io.dapr</groupId>
       <artifactId>durabletask-client</artifactId>
-      <version>1.5.6</version>
+      <version>1.5.7-SNAPSHOT</version>
```
cicoyle: This will be updated to `1.5.8` once [this durabletask-java PR is approved & merged.](https://github.com/dapr/durabletask-java/pull/34)

sdk-autogen/pom.xml
```diff
@@ -132,7 +132,7 @@
               <goal>run</goal>
             </goals>
             <configuration>
-              <protocCommand>protoc</protocCommand>
+              <protocCommand>dapr/java-sdk-protoc</protocCommand>
```
cicoyle: Updated a few `protocCommand` lines in the `pom.xml` files due to very obscure errors I ran into locally due to having different versions which resulted in flaky weird errors.

cicoyle:
Note, this PR should be cherrypicked into the release branch once merged.

cicoyle:
This PR is also dependent on the dapr release for 1.16. [Javi has a PR to update to the latest rc version here](https://github.com/dapr/java-sdk/pull/1463). We will need both merged in, then cherrypick both to the release branch

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -654,6 +655,212 @@ Key Points:
...
+
+The workflow demonstrates:
+1. The workflow starts in the main worker (crossapp-worker)
+2. Calls an activity in 'app2' using cross-app functionality
+3. Calls an activity in 'app3' using cross-app functionality
+4. The workflow completes with the final result from all activities
+
+This pattern is particularly useful for:
+- Microservices architectures where activities are distributed across multiple services
+- Service mesh scenarios where different apps handle different types of activities
```
salaboy: @cicoyle Service Mesh scenarios? What does that mean?

cicoyle: I think I just mean the first bullet- I meant the first bullet idea with the second one but its not clear and redundant, so Ill rm it

cicoyle:
I need to release durabletask-java with [this PR merged](https://github.com/dapr/durabletask-java/pull/40) in before this current PR will be green

salaboy:
@cicoyle I was looking at this PR, but it doesn't have anything related to cross app call activity.

cicoyle:
> @cicoyle I was looking at this PR, but it doesn't have anything related to cross app call activity.

Did you by chance read the thorough readme updates I added?

cicoyle:
I just got a workflow IT test locally 🎉  Need to cleanup and push, will do that + fix the conflict in my morning. Please feel free to review everything else and know I will update the pom.xml in my morning and add an IT test. Ill remove the mechanical markdown since Ill be adding the IT test.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,327 @@
...
+
+  // App2 container for App2TransformActivity - using GenericContainer for custom ports
+  @Container
+  private static final GenericContainer<?> APP2_CONTAINER = new GenericContainer<>(DAPR_RUNTIME_IMAGE_TAG)
+      .withNetwork(DAPR_NETWORK)
+      .withNetworkAliases("app2-sidecar")
+      .withExposedPorts(3501, 50002)
+      .withCommand("./daprd", 
+                   "--app-id", "app2",
+                   "--dapr-listen-addresses=0.0.0.0",
```
cicoyle: Im tweaking this bit, so dont review on these specific lines. will update the `DaprContainer.java` to clean this container definition up.

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,172 @@ Key Points:
 4. Each activity simulates work with a short delay for demonstration purposes
 
 
+### Cross-App Pattern
+
+The cross-app pattern allows workflows to call activities that are hosted in different Dapr applications. This is useful for microservices architectures where activities are distributed across multiple services, or multi-tenant applications where activities are isolated by app ID.
```
salaboy: I would remove the multi-tenant side of things because it is confusing. Something like: 
```
The cross-app pattern allows workflows to call activities that are hosted in different Dapr applications. This is useful for microservices architectures allowing multiple applications to host activities that can be orchestrated by Dapr Workflows.
```

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,172 @@ Key Points:
...
+
+The `CrossAppWorkflow` class defines the workflow. It demonstrates calling activities in different apps using the `appId` parameter in `WorkflowTaskOptions`. See the code snippet below:
+```java
+public class CrossAppWorkflow implements Workflow {
+  @Override
+  public WorkflowStub create() {
+    return ctx -> {
+      System.out.println("=== WORKFLOW STARTING ===");
+      ctx.getLogger().info("Starting CrossAppWorkflow: " + ctx.getName());
+      System.out.println("Workflow name: " + ctx.getName());
```
salaboy: @cicoyle `System.out.println()` should never be used inside a Dapr Workflow. You should always use `ctx.getLogger()` as this logger takes into account the replay workflow mechanism.

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,172 @@ Key Points:
...
+    return result;
+  }
+}
+```
+
+**WorkflowTaskOptions Constructors for Cross-App Calls:**
+
+The `WorkflowTaskOptions` class supports several constructors for cross-app calls:
+
+```java
```
salaboy: This is not wrong, but all this information is already available in the JavaDocs.

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,172 @@ Key Points:
...
+- **Child workflow cross-app calls (suborchestration) are NOT supported**
+- The app ID must match the Dapr application ID of the target service
+
+**Running the Cross-App Example:**
+
+This example requires running multiple Dapr applications simultaneously. You'll need to run the following commands in separate terminals:
+
+1. **Start the main workflow worker (crossapp-worker):**
+```sh
+dapr run --app-id crossapp-worker --resources-path ./components/workflows --dapr-grpc-port 50001 --log-level=debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.CrossAppWorker
```
salaboy: @cicoyle I would avoid using `--log-level=debug` for this example

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,172 @@ Key Points:
...
+```
+
+2. **Start app2 worker (handles App2TransformActivity):**
+```sh
+dapr run --app-id app2 --resources-path ./components/workflows --dapr-grpc-port 50002 --log-level=debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.App2Worker
```
salaboy: @cicoyle I would avoid using `--log-level=debug` for this example

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,172 @@ Key Points:
...
+```
+
+3. **Start app3 worker (handles App3FinalizeActivity):**
+```sh
+dapr run --app-id app3 --resources-path ./components/workflows --dapr-grpc-port 50003 --log-level=debug -- java -jar target/dapr-java-sdk-examples-exec.jar io.dapr.examples.workflows.crossapp.App3Worker
```
salaboy: @cicoyle I would avoid using `--log-level=debug` for this example

examples/src/main/java/io/dapr/examples/workflows/crossapp/CrossAppWorkflow.java
```diff
@@ -0,0 +1,64 @@
...
+ * Example workflow that demonstrates cross-app activity calls.
+ * This workflow calls activities in different apps using the appId parameter.
+ */
+public class CrossAppWorkflow implements Workflow {
+  @Override
+  public WorkflowStub create() {
+    return ctx -> {
+      System.out.println("=== WORKFLOW STARTING ===");
+      ctx.getLogger().info("Starting CrossAppWorkflow: " + ctx.getName());
+      System.out.println("Workflow name: " + ctx.getName());
```
salaboy: @cicoyle never use `System.out.println()` inside a Dapr Workflow class

pom.xml
```diff
@@ -16,7 +16,6 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <grpc.version>1.69.0</grpc.version>
     <protobuf.version>3.25.5</protobuf.version>
-    <protocCommand>protoc</protocCommand>
```
salaboy: @cicoyle I wouldn't remove this.

sdk-autogen/pom.xml
```diff
@@ -22,7 +22,6 @@
     <protobuf.input.directory>${project.build.directory}/proto</protobuf.input.directory>
     <maven.deploy.skip>false</maven.deploy.skip>
     <grpc.version>1.69.0</grpc.version>
-    <protocCommand>protoc</protocCommand>
```
salaboy: @cicoyle I wouldn't remove this.

sdk-autogen/pom.xml
```diff
@@ -132,7 +131,6 @@
               <goal>run</goal>
             </goals>
             <configuration>
-              <protocCommand>protoc</protocCommand>
```
salaboy: @cicoyle I wouldn't remove this.

spring-boot-examples/consumer-app/src/test/java/io/dapr/springboot/examples/consumer/DaprTestContainersConfig.java
```diff
@@ -93,7 +93,7 @@ public DaprContainer daprContainer(Network daprNetwork, RabbitMQContainer rabbit
         .withAppName("consumer-app")
         .withNetwork(daprNetwork).withComponent(new Component("pubsub",
             "pubsub.rabbitmq", "v1", rabbitMqProperties))
-        .withDaprLogLevel(DaprLogLevel.INFO)
+        .withDaprLogLevel(DaprLogLevel.DEBUG)
```
salaboy: @cicoyle I definitely wouldn't change this. It has nothing to do with the purpose of this PR

spring-boot-examples/workflows/src/test/java/io/dapr/springboot/examples/wfp/DaprTestContainersConfig.java
```diff
@@ -44,7 +55,9 @@ public DaprContainer daprContainer(Network network) {
             .withAppPort(8080)
             .withNetwork(network)
             .withAppHealthCheckPath("/actuator/health")
-            .withAppChannelAddress("host.testcontainers.internal");
+            .withAppChannelAddress("host.testcontainers.internal")
```
salaboy: @cicoyle this change has nothing to do with the Cross App CallActivity feature.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -166,6 +170,16 @@ public DaprContainer withSchedulerService(String schedulerService) {
     return this;
   }
 
+  public DaprContainer withCustomHttpPort(Integer port) {
```
salaboy: @cicoyle I don't fully understand the reason why this is needed. I would suggest against adding this change. Testcontainers by default will start Dapr containers with random ports, the user shouldn't be setting those ports manually.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -166,6 +170,16 @@ public DaprContainer withSchedulerService(String schedulerService) {
...
+    this.customHttpPort = port;
+    return this;
+  }
+
+  public DaprContainer withCustomGrpcPort(Integer port) {
```
salaboy: @cicoyle I don't fully understand the reason why this is needed. I would suggest against adding this change. Testcontainers by default will start Dapr containers with random ports, the user shouldn't be setting those ports manually.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -303,11 +353,33 @@ protected void configure() {
...
 
+    dependsOn(placementContainer, schedulerContainer);
+  }
+
+  @Override
+  public void start() {
     List<String> cmds = new ArrayList<>();
     cmds.add("./daprd");
+    
+    String finalAppName = (appName != null && !appName.trim().isEmpty())
```
salaboy: @cicoyle I like this, but I think before the `appName` was mandatory, as if it was null it should break the container to start.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -303,11 +353,33 @@ protected void configure() {
...
+        ? appName 
+        : "dapr-app-" + UUID.randomUUID().toString().substring(0, 8);
+    
     cmds.add("--app-id");
-    cmds.add(appName);
+    cmds.add(finalAppName);
+    
     cmds.add("--dapr-listen-addresses=0.0.0.0");
+    
+    if (customHttpPort != null) {
```
salaboy: @cicoyle I really think we shouldn't do this. As we are relying on Testcontainers to map the ports, there is no need to change the default ones.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -303,11 +353,33 @@ protected void configure() {
...
+      cmds.add("--dapr-http-port");
+      cmds.add(Integer.toString(customHttpPort));
+    }
+    
+    if (customGrpcPort != null) {
```
salaboy: @cicoyle I really think we shouldn't do this. As we are relying on Testcontainers to map the ports, there is no need to change the default ones.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -287,6 +301,42 @@ protected void configure() {
       withNetwork(Network.newNetwork());
     }
 
+    if (configuration != null) {
```
salaboy: @cicoyle this block of code seems to be moved. If there is no change in the logic of the code, please revert back to its original place.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+  private static DaprContainer APP3_CONTAINER;
+
+  // TestContainers for each app
+  private static GenericContainer<?> crossappWorker;
+  private static GenericContainer<?> app2Worker;
+  private static GenericContainer<?> app3Worker;
+
+  @BeforeAll
+  public static void setUp() throws Exception {
+    sharedPlacementContainer = new DaprPlacementContainer("daprio/placement:1.16.0-rc.3")
```
salaboy: @cicoyle please don't hardcode versions, use constants from https://github.com/dapr/java-sdk/blob/master/testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainerConstants.java#L16

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+    MAIN_WORKFLOW_CONTAINER = new DaprContainer(DAPR_RUNTIME_IMAGE_TAG)
+        .withAppName("crossapp-worker")
+        .withNetwork(DAPR_NETWORK)
+        .withNetworkAliases("main-workflow-sidecar")
+        .withPlacementContainer(sharedPlacementContainer)
+        .withSchedulerContainer(sharedSchedulerContainer)
+        .withComponent(new Component("kvstore", "state.in-memory", "v1", Map.of("actorStateStore", "true")))
+        .withComponent(new Component("pubsub", "pubsub.in-memory", "v1", Collections.emptyMap()))
+        .withSubscription(new Subscription("local", "pubsub", "topic", "/events"))
+        .withDaprLogLevel(DaprLogLevel.DEBUG)
```
salaboy: @cicoyle there is no need for Debug level if this is just an integration test.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withLogConsumer(outputFrame -> System.out.println("MAIN_WORKFLOW: " + outputFrame.getUtf8String()))
```
salaboy: @cicoyle there is no need to output the logs to the console if this is a test that will assert on results.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withAppChannelAddress("host.testcontainers.internal")
+        .waitingFor(Wait.forHttp("/v1.0/healthz/outbound")
```
salaboy: @cicoyle this is already built in in the DaprContainer, if a condition was missing, it will be better to add it to the DaprContainer logic that adding more every time a user creates a new instance.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+            .forStatusCode(200)
+            .forStatusCode(204));
+    
+    APP2_CONTAINER = new DaprContainer(DAPR_RUNTIME_IMAGE_TAG)
+        .withAppName("app2")
+        .withNetwork(DAPR_NETWORK)
+        .withNetworkAliases("app2-sidecar")
+        .withPlacementContainer(sharedPlacementContainer)
+        .withSchedulerContainer(sharedSchedulerContainer)
+        .withCustomHttpPort(3501)
```
salaboy: @cicoyle I don't think we need this. We should avoid doing this, as Testcontainers is already mapping the container ports.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withCustomGrpcPort(50002)
```
salaboy: @cicoyle I don't think we need this. We should avoid doing this, as Testcontainers is already mapping the container ports.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withAppChannelAddress("main-workflow-sidecar:3500")
+        .withDaprLogLevel(DaprLogLevel.DEBUG)
```
salaboy: @cicoyle there is no need for Debug level on a test.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withComponent(new Component("kvstore", "state.in-memory", "v1",
+            Map.of("actorStateStore", "true")))
+        .withComponent(new Component("pubsub", "pubsub.in-memory", "v1", Collections.emptyMap()))
+        .withSubscription(new Subscription("local", "pubsub", "topic", "/events"))
+        .withLogConsumer(outputFrame -> System.out.println("APP2: " + outputFrame.getUtf8String()))
```
salaboy: @cicoyle there is no need to output logs for the sidecar on a test.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .waitingFor(Wait.forHttp("/v1.0/healthz/outbound")
```
salaboy: @cicoyle as stated before, this is already built in, there shouldn't be a need to define it here.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withLogConsumer(outputFrame -> System.out.println("CrossAppWorker: " + outputFrame.getUtf8String()));
+    
+    // Start app2 worker (connects to APP2_CONTAINER)
+    app2Worker = new GenericContainer<>("openjdk:17-jdk-slim")
+        .withCopyFileToContainer(MountableFile.forHostPath("target/test-classes"), "/app/classes")
+        .withCopyFileToContainer(MountableFile.forHostPath("target/dependency"), "/app/libs")
+        .withWorkingDirectory("/app")
+        .withCommand("java", "-cp", "/app/classes:/app/libs/*", 
+                     "-Ddapr.app.id=app2",
+                     "-Ddapr.grpc.endpoint=app2-sidecar:50002",
```
salaboy: @cicoyle here you can use the Testcontainers mapped (random) port instead of relying on a custom port number.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+    
+    // Start app3 worker (connects to APP3_CONTAINER)
+    app3Worker = new GenericContainer<>("openjdk:17-jdk-slim")
+        .withCopyFileToContainer(MountableFile.forHostPath("target/test-classes"), "/app/classes")
+        .withCopyFileToContainer(MountableFile.forHostPath("target/dependency"), "/app/libs")
+        .withWorkingDirectory("/app")
+        .withCommand("java", "-cp", "/app/classes:/app/libs/*", 
+                     "-Ddapr.app.id=app3",
+                     "-Ddapr.grpc.endpoint=app3-sidecar:50003",
+                     "-Ddapr.http.endpoint=app3-sidecar:3502",
```
salaboy: @cicoyle here you can use the Testcontainers mapped (random) port instead of relying on a custom port number.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+    
+    Properties clientProperties = new Properties(propertyOverrides);
+    DaprWorkflowClient workflowClient = new DaprWorkflowClient(clientProperties);
+    
+    try {
+      String instanceId = workflowClient.scheduleNewWorkflow(CrossAppWorkflow.class, input);
+      assertNotNull(instanceId, "Workflow instance ID should not be null");
+      workflowClient.waitForInstanceStart(instanceId, Duration.ofSeconds(30), false);
+
+      Duration timeout = Duration.ofMinutes(2);
```
salaboy: @cicoyle Is this really needed? doesn't the workflow starts and finishes quite quickly?

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+        .withNetworkAliases("app3-sidecar")
+        .withPlacementContainer(sharedPlacementContainer)
+        .withSchedulerContainer(sharedSchedulerContainer)
+        .withCustomHttpPort(3502)
+        .withCustomGrpcPort(50003)
+        .withAppChannelAddress("main-workflow-sidecar:3500")
+        .withDaprLogLevel(DaprLogLevel.DEBUG)
+        .withComponent(new Component("kvstore", "state.in-memory", "v1",
+            Map.of("actorStateStore", "true")))
+        .withComponent(new Component("pubsub", "pubsub.in-memory", "v1", Collections.emptyMap()))
```
salaboy: @cicoyle is the pubsub and subscription needed here?

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,267 @@
...
+
+    // Start crossapp worker (connects to MAIN_WORKFLOW_CONTAINER)
+    crossappWorker = new GenericContainer<>("openjdk:17-jdk-slim")
+        .withCopyFileToContainer(MountableFile.forHostPath("target/test-classes"), "/app/classes")
+        .withCopyFileToContainer(MountableFile.forHostPath("target/dependency"), "/app/libs")
+        .withWorkingDirectory("/app")
+        .withCommand("java", "-cp", "/app/classes:/app/libs/*", 
+                     "-Ddapr.app.id=crossapp-worker",
+                     "-Ddapr.grpc.endpoint=main-workflow-sidecar:50001",
+                     "-Ddapr.http.endpoint=main-workflow-sidecar:3500",
```
salaboy: @cicoyle this should use Testcontainers mapped ports instead of hardcoded ports.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
cicoyle: Im aligning with dapr for using debug for tests. This is helpful for debugging when issues arise, as it will be in debug mode with logs that might be useful if regressions or the like arise. I feel strongly to leave tests in debug mode. Otherwise its an extra step later if an issue arises to try to debug it

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
cicoyle: I was using it for local development purposes. However, I still firmly believe we should keep the `withLogConsumer` bc the log output without the log consumers shows very little useful information. The current output only shows container startup messages and YAML configurations, but nothing about the actual workflow execution, cross-app communication, or any potential issues, etc.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
cicoyle: I believe it is useful per my above comments. This aligns with other dapr repos

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
cicoyle: please see my other comments as to why I believe we should keep this

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
cicoyle: Actually, this is using internal container communication within the Docker network, not host communication. The worker containers are all running in the same Docker network and communicate directly with each other using network aliases like `main-workflow-sidecar:50001`. 

For internal container-to-container communication, I believe we need to use the internal network names and ports, not mapped host ports. The mapped ports are only needed when communicating from the host (test code) to containers.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
cicoyle: I moved this code from `start()` to `configure()` because I was experiencing container startup failures where Dapr couldn't find the required YAML resources (components, subscriptions, etc.) when they were configured in `start()`. The `configure()` method runs before container startup, ensuring all resources are properly prepared and copied to the container before Dapr begins execution versus the `start()` method runs during container startup. It is best practice to have configuration in a `configure()` method. The code block was moved to fix actual issues, not just organizational preferences.

examples/src/main/java/io/dapr/examples/workflows/README.md
cicoyle: I can remove it if you prefer, seemed relevant at the time to note it.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
salaboy: @cicoyle I think that now is good, because at least testcontainers is starting the containers in random ports.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
salaboy: @cicoyle I am not entirely sure to understand this comment. If you find problems with the code sequence, then we should create a test to validate the problem. But now, the code was moved and I am not sure which problems were fixed by moving the code.

testcontainers-dapr/src/main/java/io/dapr/testcontainers/DaprContainer.java
```diff
@@ -357,52 +399,9 @@ protected void configure() {
...
-      String endpointYaml = HTTPENDPOINT_CONVERTER.convert(endpoint);
-
-      LOGGER.info("> HTTPEndpoint YAML: \n");
-      LOGGER.info("\t\n" + endpointYaml + "\n");
-
-      withCopyToContainer(Transferable.of(endpointYaml), "/dapr-resources/" + endpoint.getName() + ".yaml");
-    }
-
-    dependsOn(placementContainer, schedulerContainer);
+    super.start();
```
salaboy: @cicoyle this is definitely not good, be careful with this. If you add a `super.start()` here testcontainers will not be able to manage the lifecycle of the container. Why was this added? I don't agree with this change.

salaboy:
@cicoyle codecov is still failing and sb 3.3.x was cancelled for some reason, I think the changes in the PR are ok if the pipelines and codecov are green.

cicoyle:
> @cicoyle codecov is still failing and sb 3.3.x was cancelled for some reason, I think the changes in the PR are ok if the pipelines and codecov are green.

I pushed a test that should appease codecov. I also upped the test timeout to unblock this PR. I will open an issue to parallelize the test suite a bit bc it should not take this long.

cicoyle:
[created this issue](https://github.com/dapr/java-sdk/issues/1522)

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   build:
     name: "Build jdk:${{ matrix.java }} sb:${{ matrix.spring-boot-display-version }} exp:${{ matrix.experimental }}"
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
```
salaboy: @cicoyle Is this really needed? How long is the build taking to increase this? If for some reason the pipeline hangs, now we will need to wait for 60 mins for this to be cancelled.

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -678,6 +679,155 @@ Key Points:
...
+- **Error handling**: Works the same as local activity calls
+
+**Requirements:**
+- Multiple Dapr applications running with different app IDs
+- Activities registered in the target applications
+- Proper Dapr workflow runtime configuration
+
+**Important Limitations:**
+- **Cross-app calls are currently supported for activities only**
+- **Child workflow cross-app calls (suborchestration) are NOT supported**
```
kendallroden, line 787: I believe this was added to dapr after this doc was created - may want to remove this limitation before merging - just happened to catch

cicoyle: Hey @kendallroden thanks for the review!

This was intentional. At the moment, only `callactivity` logic is exposed and fully supported in the java sdk. More work is required in the durabletask-java repo for `suborchestration` to be supported. That will come after the 1.16 version for java. Only `callactivity` cross app is supported at the moment in the sdk.

.github/workflows/build.yml
```diff
@@ -47,7 +47,7 @@ jobs:
   build:
     name: "Build jdk:${{ matrix.java }} sb:${{ matrix.spring-boot-display-version }} exp:${{ matrix.experimental }}"
     runs-on: ubuntu-latest
```
artur-ciocanu: Do we need a timeout of an hour? Even 30 mins seems a lot, but I am wondering why we need to increase the timeout.

examples/src/main/java/io/dapr/examples/workflows/crossapp/CrossAppWorkflow.java
```diff
@@ -0,0 +1,64 @@
...
+import io.dapr.workflows.WorkflowTaskOptions;
+
+/**
+ * Example workflow that demonstrates cross-app activity calls.
+ * This workflow calls activities in different apps using the appId parameter.
+ */
+public class CrossAppWorkflow implements Workflow {
+  @Override
+  public WorkflowStub create() {
+    return ctx -> {
```
artur-ciocanu, line 27: Could we get a reference to `logger` and avoid the `ctx.getLogger()`. Also when using `slf4j` `Logger` you can use `{}` placeholder and avoid string concatenation.

We can have something like this:
```
Logger logger = ctx.getLogger();
logger.info("=== WORKFLOW STARTING ===");
logger.info("Starting CrossAppWorkflow: {}", ctx.getName());
...      
```

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -681,6 +682,155 @@ Key Points:
...
+### Cross-App Pattern
+
+The cross-app pattern allows workflows to call activities that are hosted in different Dapr applications. This is useful for microservices architectures allowing multiple applications to host activities that can be orchestrated by Dapr Workflows.
+
+The `CrossAppWorkflow` class defines the workflow. It demonstrates calling activities in different apps using the `appId` parameter in `WorkflowTaskOptions`. See the code snippet below:
+```java
+public class CrossAppWorkflow implements Workflow {
+  @Override
+  public WorkflowStub create() {
+      return ctx -> {
```
artur-ciocanu, line 694: We can get a reference to logger and use it instead of repeating `ctx.getLogger()`. Also we can use placeholders in `logger.info()` calls instead of concatenating strings.

examples/src/main/java/io/dapr/examples/workflows/README.md
```diff
@@ -681,6 +682,155 @@ Key Points:
...
+  }
+}
+
+```
+
+The `App2TransformActivity` class defines an activity in app2 that transforms the input string. See the code snippet below:
+```java
+public class App2TransformActivity implements WorkflowActivity {
+  @Override
+  public Object run(WorkflowActivityContext ctx) {
```
artur-ciocanu, line 740: I have opened a PR: https://github.com/dapr/java-sdk/pull/1534 that adds logger to `WorkflowActivityContext` once that is approved I think we could use the logger instead of `System.out.print`

examples/src/main/java/io/dapr/examples/workflows/crossapp/App2TransformActivity.java
```diff
@@ -0,0 +1,33 @@
...
+import io.dapr.workflows.WorkflowActivity;
+import io.dapr.workflows.WorkflowActivityContext;
+
+/**
+ * TransformActivity for App2 - transforms input to uppercase.
+ * This activity is called cross-app from the main workflow.
+ */
+public class App2TransformActivity implements WorkflowActivity {
+    @Override
+    public Object run(WorkflowActivityContext context) {
```
artur-ciocanu: Once we #1534 is approved and merge we can use the `ctx.getLogger()` to log messages instead of `System.out`

examples/src/main/java/io/dapr/examples/workflows/crossapp/App3FinalizeActivity.java
```diff
@@ -0,0 +1,33 @@
...
+import io.dapr.workflows.WorkflowActivity;
+import io.dapr.workflows.WorkflowActivityContext;
+
+/**
+ * FinalizeActivity for App3 - adds final processing.
+ * This activity is called cross-app from the main workflow.
+ */
+public class App3FinalizeActivity implements WorkflowActivity {
+    @Override
+    public Object run(WorkflowActivityContext context) {
```
artur-ciocanu, line 25: Once we https://github.com/dapr/java-sdk/pull/1534 is approved and merge we can use the `ctx.getLogger()` to log messages instead of `System.out`

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/App2TransformActivity.java
```diff
@@ -0,0 +1,36 @@
...
+import io.dapr.workflows.WorkflowActivityContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class App2TransformActivity implements WorkflowActivity {
+
+  private static final Logger logger = LoggerFactory.getLogger(App2TransformActivity.class);
+
+  @Override
+  public Object run(WorkflowActivityContext ctx) {
```
artur-ciocanu, line 26: Once we https://github.com/dapr/java-sdk/pull/1534 is approved and merge we can use the `ctx.getLogger()` to log messages instead of `System.out`

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/App3FinalizeActivity.java
```diff
@@ -0,0 +1,36 @@
...
+import io.dapr.workflows.WorkflowActivityContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class App3FinalizeActivity implements WorkflowActivity {
+
+  private static final Logger logger = LoggerFactory.getLogger(App3FinalizeActivity.class);
+
+  @Override
+  public Object run(WorkflowActivityContext ctx) {
```
artur-ciocanu, line 26: Once we https://github.com/dapr/java-sdk/pull/1534 is approved and merge we can use the `ctx.getLogger()` to log messages instead of `System.out`.

Also you should use `logger.inof(""Input: {}", input)` instead of string concatenation

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/CrossAppWorkflow.java
```diff
@@ -0,0 +1,58 @@
...
+*/
+
+package io.dapr.it.testcontainers.workflows.crossapp;
+
+import io.dapr.workflows.Workflow;
+import io.dapr.workflows.WorkflowStub;
+import io.dapr.workflows.WorkflowTaskOptions;
+import org.slf4j.Logger;
+
+public class CrossAppWorkflow implements Workflow {
```
artur-ciocanu, line 21: nit: too many new lines

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/CrossAppWorkflow.java
```diff
@@ -0,0 +1,58 @@
...
+
+
+
+  @Override
+  public WorkflowStub create() {
+    return ctx -> {
+      Logger logger = ctx.getLogger();
+      String instanceId = ctx.getInstanceId();
```
artur-ciocanu, line 26: We should use `logger.info` with placeholders instead of string concatenation. Like:
```
logger.info("Instance ID: {}", instanceId);
```

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,278 @@
...
+ * 2. Launching Java processes that register workflows/activities in separate apps
+ * 3. Executing a cross-app workflow
+ * 4. Asserting successful completion
+ */
+@Testcontainers
+@Tag("testcontainers")
+public class WorkflowsCrossAppCallActivityIT {
+
+  private static final Network DAPR_NETWORK = Network.newNetwork();
+
```
artur-ciocanu, line 58: I am wondering why can't we use `@Container` annotation for all the containers. As far as I can see in the `setUp` method we are just instantiating the containers. Using the `@Container` annotations allows Testcontainer machinery to take care of bootstrap and tear down without explicitly starting or stoping containers.

Using this approach we can get rid of `setUp` and `tearDown` methods.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,278 @@
...
+    // TestContainers wait strategies ensure all containers are ready before this test runs
+
+    String input = "Hello World";
+    String expectedOutput = "HELLO WORLD [TRANSFORMED BY APP2] [FINALIZED BY APP3]";
+    
+    // Create workflow client connected to the main workflow orchestrator
+    // Use the same endpoint configuration that the workers use
+    // The workers use host.testcontainers.internal:50001, so we need to use the mapped port
+    String grpcEndpoint = "localhost:" + MAIN_WORKFLOW_CONTAINER.getMappedPort(50001);
+    String httpEndpoint = "localhost:" + MAIN_WORKFLOW_CONTAINER.getMappedPort(3500);
```
artur-ciocanu: Could we please remove the `System.setProperty()` calls, this have caused a lot of issues in the past. As far as I remember we are using property overrides everywhere. I think property overrides should be enough.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
```diff
@@ -0,0 +1,278 @@
...
+      assertEquals(expectedOutput, workflowOutput, "Workflow output should match expected result");
+    } finally {
+      workflowClient.close();
+    }
+  }
+  
+  /**
+   * Ensures that dependencies are copied to target/dependency for container classpath.
+   * This is needed because the containers need access to the workflow runtime classes.
+   */
```
artur-ciocanu: I am wondering why is this required? Can't we make sure that all the dependencies are in the right place during build, so the container just copies the files without having to invoke Maven behind the scenes.

artur-ciocanu (commented):
@cicoyle this looks great! I have left some comments. Some of them are nitpicks.

I want to make sure that all the ITs that are using Testcontainers follow the same pattern and we rely on Testcontainer machinery for all the resource bootstrap and cleanup.

I am guessing you have run into issues that's why you added the `ensureDependenciesCopied()` method, but my question is why can't we have everything done during build, so we don't have to invoke Maven during IT execution.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
salaboy: I agree.. this needs to be changed to align with other tests.

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
salaboy: @artur-ciocanu 100% agree

.github/workflows/build.yml
javier-aliaga: we should not need this now, I have improved the CI and now it takes 20 mins

examples/src/main/java/io/dapr/examples/workflows/README.md
javier-aliaga: done

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
javier-aliaga: We are using `@Container` now

javier-aliaga:
@artur-ciocanu @salaboy I fixed all you comments and suggestion, can you take look again?

sdk-tests/src/test/java/io/dapr/it/testcontainers/workflows/crossapp/WorkflowsCrossAppCallActivityIT.java
salaboy: amazing stuff @javier-aliaga

sdk-workflows/pom.xml
```diff
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>io.dapr</groupId>
       <artifactId>durabletask-client</artifactId>
-      <version>1.5.7</version>
+      <version>1.5.10</version>
```
salaboy, line 50: 100+ 💯

salaboy (approved):
LGTM, thanks @javier-aliaga for adding the @Containers to the test.. that will simplify our lives

Issue #1434, opened 2025-06-26
Java Version Mismatch between Core SDK & Spring Modules


cicoyle:
The Java SDK project has a configuration inconsistency where the core modules use Java 11, but the Spring modules were configured to use [Spring Framework 6.1.21 and Spring Boot 3.2.6, and now we are at 6.2.7 && 3.4.6](https://github.com/dapr/java-sdk/pull/1427/files) which require Java 17+.

Core SDK modules (sdk, sdk-workflows, etc.) are configured for Java 11
Spring modules are using Spring Framework 6.2.7 and Spring Boot 3.4.6, which require Java 17+
This creates a compilation error when trying to build the entire project

[See the core sdk config in the pom.xml here](https://github.com/dapr/java-sdk/blob/master/pom.xml#L32-L34)
[See the spring parent pom that has java 11](https://github.com/dapr/java-sdk/blob/master/dapr-spring/pom.xml#L31-L33), but incompatible deps

from master run `mvn compile -q`
```
git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
cassie@Cassandras-MacBook-Pro java-sdk % git pull
Already up to date.
cassie@Cassandras-MacBook-Pro java-sdk % mvn compile -q     
[ERROR] COMPILATION ERROR : 
[ERROR] /Users/cassie/go/src/github.com/java-sdk/spring-boot-examples/workflows/src/main/java/io/dapr/springboot/examples/wfp/WorkflowPatternsApplication.java:[16,32] cannot access org.springframework.boot.SpringApplication
  bad class file: /Users/cassie/.m2/repository/org/springframework/boot/spring-boot/3.4.3/spring-boot-3.4.3.jar(/org/springframework/boot/SpringApplication.class)
    class file has wrong version 61.0, should be 55.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
[ERROR] /Users/cassie/go/src/github.com/java-sdk/spring-boot-examples/workflows/src/main/java/io/dapr/springboot/examples/wfp/WorkflowPatternsApplication.java:[17,46] cannot access org.springframework.boot.autoconfigure.SpringBootApplication
  bad class file: /Users/cassie/.m2/repository/org/springframework/boot/spring-boot-autoconfigure/3.4.3/spring-boot-autoconfigure-3.4.3.jar(/org/springframework/boot/autoconfigure/SpringBootApplication.class)
    class file has wrong version 61.0, should be 55.0
[...]

PR #1498, opened 2025-08-13
rm protoc cmd from pom.xml

cicoyle:
Locally I have a diff protoc version. That results in a crazy bad error. To not use a developers local protoc Im removing it so cmds like `mvn clean install` will run without error.

[I believe it fixes this issue](https://github.com/dapr/java-sdk/issues/1434) or is at least a very similar err. See the err in the attached file.
[protoc-issue.txt](https://github.com/user-attachments/files/21762678/protoc-issue.txt)

[Also noticed CI throwing](https://github.com/dapr/java-sdk/actions/runs/16992501822/job/48175129624):
```
Run ./mvnw compile -B -q
protoc-jar: executing: [protoc, --version]
protoc-jar: caught exception, retrying: java.io.IOException: Cannot run program "protoc": error=2, No such file or directory
protoc-jar: executing: [protoc, --version]
protoc-jar: caught exception, retrying: java.io.IOException: Cannot run program "protoc": error=2, No such file or directory
protoc-jar: executing: [protoc, --version]
protoc-jar: executing: [/tmp/protoc18031117828517050527.exe, --version]
libprotoc 25.5
```


artur-ciocanu (approved):
LGTM, does this mean that the Maven plugin will use its internal protoc version?

PR #1517, opened 2025-08-26
Bump 1.15.10 for 1.15.1

salaboy:
# Description

Bump durable tasks to 1.5.10 and fix conflicts for Dapr SDK 1.15.1

## Issue reference

Please reference the issue this PR will close: #1516 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-workflows/src/main/java/io/dapr/workflows/WorkflowTaskOptions.java
+++ b/sdk-workflows/src/main/java/io/dapr/workflows/WorkflowTaskOptions.java
@@ -17,18 +17,48 @@ public class WorkflowTaskOptions {
 
   private final WorkflowTaskRetryPolicy retryPolicy;
   private final WorkflowTaskRetryHandler retryHandler;
+  private final String appId;
 
   public WorkflowTaskOptions(WorkflowTaskRetryPolicy retryPolicy, WorkflowTaskRetryHandler retryHandler) {
-    this.retryPolicy = retryPolicy;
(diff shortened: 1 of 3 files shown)


sdk-workflows/src/main/java/io/dapr/workflows/WorkflowTaskOptions.java
```diff
@@ -17,18 +17,48 @@ public class WorkflowTaskOptions {
...
+  }
+
+  /**
+   * Constructor for WorkflowTaskOptions with app ID for cross-app calls.
+   *
+   * @param appId the ID of the app to call the activity in
+   */
+  public WorkflowTaskOptions(String appId) {
+    this(null, null, appId);
+  }
```
javier-aliaga: Should be throw an exception here?

salaboy: @javier-aliaga I thought about that.. but then I think it should throw an exception everywhere where it is not null.. meaning that if you are trying to set it with this version, you should be stopped.

javier-aliaga (approved):

sdk-workflows/src/main/java/io/dapr/workflows/WorkflowTaskOptions.java
salaboy: done

javier-aliaga (approved):

PR #1567, opened 2025-09-30
Setting default props for DaprClient

### LINKED ISSUE #1566: DaprClientProperties don't have defaults set
## Expected Behavior

When starting a Spring Boot application without setting DaprClientProperties, these should be instantiated with the default Dapr values.

Http Endpoint should be: "http://localhost"  by default if not overriden
GRPC Endpoint should be: "localhost"  by default if not overriden
HTTP port should be: 3500
GRPC port should be: 50001
API token should be: null (unless specified)

## Actual Behavior

Properties instanciated with empty values for http and grpc endpoints and ports.

## Release Note

RELEASE NOTE:

salaboy:
# Description

Setting the default values for DaprClient Spring Boot properties. 

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature has been discussed prior to implementation.

Please reference the issue this PR will close: #1566

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


mcruzdev (approved):
LGTM

salaboy:
@dapr/maintainers-java-sdk @dapr/approvers-java-sdk without setting these defaults, Spring Boot doesn't know where to find Dapr when running the application. This force the user to set the properties when they shouldn't.

siri-varma (dismissed):
LGTM

dapr-spring/dapr-spring-boot-autoconfigure/src/main/java/io/dapr/spring/boot/autoconfigure/client/DaprClientProperties.java
```diff
@@ -17,10 +17,10 @@
 
 @ConfigurationProperties(prefix = "dapr.client")
 public class DaprClientProperties {
```
artur-ciocanu, line 19: I wondering if an endpoints shouldn't have the port. As far as I remember there is some parsing logic in the network utils.

dapr-spring/dapr-spring-boot-autoconfigure/src/test/java/io/dapr/spring/boot/autoconfigure/client/DaprClientPropertiesTest.java
```diff
@@ -105,6 +120,8 @@ public void shouldMapDaprClientPropertiesCamelCase() {
     });
   }
```
artur-ciocanu, line 121: We could remove unnecessary new lines.

artur-ciocanu (commented):
@salaboy I like the idea, however I think the endpoint syntax should be a little bit different. It should include the port as well.

I have looked in these places:
- GRPC endpoint - https://github.com/dapr/java-sdk/blob/222b9a0fdc3845ee922493d156fa9e29e9ee3c34/sdk/src/main/java/io/dapr/utils/NetworkUtils.java#L277
- HTTP endpoint - https://github.com/dapr/java-sdk/blob/222b9a0fdc3845ee922493d156fa9e29e9ee3c34/sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java#L76

In both of these places the expectation is that we provide the port as part of endpoint.

artur-ciocanu:
@salaboy I see that the build steps are timing out could you please check what might be the issue.

Thank you!

cicoyle (changes requested):
I second @artur-ciocanu's comments here - looks like its causing issues with CI - this should be resolved before its merged.

dapr-spring/dapr-spring-boot-autoconfigure/src/main/java/io/dapr/spring/boot/autoconfigure/client/DaprClientProperties.java
salaboy: @artur-ciocanu yeah.. I have this constant feeling that this is not aligned all over the place.. but I didn't wanted to add more changes to this PR. Maybe we need another issue to track this alignment. Right now, this is just setting the default values.

cicoyle (changes requested):
There looks to be an endpoint misalignment/misconfiguration going on causing CI to hang.

You may need to tweak things after confirming how the `daprworkflowclient` is constructed using the `buildgrpcmanagedchannel`. I think its coming from this file `NetworkUtils.java` as I see we have this check which might be were that `443` is coming from in the logs which would be the wrong port with this configuration:
```
        String parsedPort = matcher.group("port");
        if (parsedPort != null) {
          port = Integer.parseInt(parsedPort);
        } else {
          // This implements default port as 80 for http for backwards compatibility.
          port = http ? 80 : 443;
        }
```

artur-ciocanu:
@salaboy I think this PR needs more investigation. The build timeouts, it seems that default values are conflicting with other overrides and tests try to connect to the wrong ports and build times out.

Issue #1569, opened 2025-10-01
No way to pass state metadata into `saveState` method


alicejgibbons:
## Expected Behavior

Today the only way to pass metadata into a save state call is to use the bulkSaveState method, considering the following method sets the metadata to null:

https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/AbstractDaprClient.java#L503

## Actual Behavior

I would expect there to be a method with a saveState option to pass it metadata like in go: https://github.com/dapr/go-sdk/blob/main/client/state.go#L279

## Release Note

RELEASE NOTE: **FIX** Pass metadata in saveState.

PR #1570, opened 2025-10-01
adding new method signature plus test

salaboy:
# Description

Add new method signature to support metadata map, I've added a test that validates that the custom metadata is added.

## Issue reference

Please reference the issue this PR will close: #1569 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
```diff
@@ -509,6 +510,24 @@ public Mono<Void> saveState(String storeName, String key, String etag, Object va
     return this.saveBulkState(storeName, Collections.singletonList(state));
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Mono<Void> saveState(String storeName, String key, String etag, Object value, Map<String, String> meta,
+                              StateOptions options) {
```
artur-ciocanu, line 518: It is dangerous to modify the incoming method parameter. I would rather make a copy and add the content type if `value` is not null.

artur-ciocanu (commented):
@salaboy nice job! I just think we should be careful not to change the incoming parameter. For example if someone would send an unmodifiable map as parameter, it will result in a runtime exception.

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
```diff
@@ -509,6 +510,24 @@ public Mono<Void> saveState(String storeName, String key, String etag, Object va
...
+    if (meta == null) {
+      meta = new HashMap<>();
+    }
+
+    if (value != null) {
+      meta.put("contentType", stateSerializer.getContentType());
```
javier-aliaga: What if the key already exists? Is it ok to overwrite it?

salaboy: solved

sdk/src/main/java/io/dapr/client/AbstractDaprClient.java
salaboy: done.. thanks for the suggestion @artur-ciocanu

artur-ciocanu (approved):
@salaboy looks great!

Issue #1571, opened 2025-10-13
Move Dapr Durable Tasks Java to a module in the Java SDK


salaboy:
## Describe the proposal

To simplify the release and maintenance of the durable task implementation, we should move the source code (located here: https://github.com/dapr/durabletask-java/ fork), which is using Gradle hosted to a Maven module hosted in the Java SDK repository. 

This will improve: 
1) Releasing the Java SDK
2) Testing the Workflow capabilities with latest changes from Durable Tasks without waiting to be released
3) Aligning tooling and release process

siri-varma:
Closing this because the PR is already merged.

Thank you salaboy for taking care of this.

PR #1529, opened 2025-08-28
Fix CVEs

javier-aliaga:
This pr includes:
- Use stored cert from resources folder to avoid importing bouncy castle library just for tests
- Ci: Split unit test to its own job so we can run them in parallel and shorten all the cycle in around 5 minutes

# Description

_Please explain the changes you've made_

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/1528

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #1575, opened 2025-10-21
Bringing Durable Task Java as a Maven module inside the Java SDK

salaboy:
# Description

- Migrated Durable Task Java from Gradle to Maven
- Migrated proto generation to maven
- Aligned dependencies to make sure we use the same grpc and protobuf versions
- Aligned codestyle checks and javadoc generation
- Updated the build workflow to test using durabletask runtime in a container
- Refactored (internal) methods that don't conform with the Java SDK style (UUID to Uuid)

## Issue reference

Please reference the issue this PR will close: #[1571](https://github.com/dapr/java-sdk/issues/1571)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -356,13 +356,13 @@
       </dependency>
       <dependency>
         <groupId>org.junit.jupiter</groupId>
-        <artifactId>junit-jupiter-params</artifactId>
+        <artifactId>junit-jupiter-engine</artifactId>
         <version>${junit-bom.version}</version>
       </dependency>
       <dependency>
-        <groupId>com.fasterxml.jackson.datatype</groupId>
-        <artifactId>jackson-datatype-jsr310</artifactId>
-        <version>${jackson.version}</version>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-params</artifactId>
+        <version>${junit-bom.version}</version>
       </dependency>
       <dependency>
         <groupId>io.projectreactor</groupId>


salaboy:
Merging this PR will mean that we need to archive https://github.com/dapr/durabletask-java/

salaboy:
@mcruzdev @artur-ciocanu folks I want you to be aware of this, as the PRs to align versions and remove netty-shaded are all collapsed here.

salaboy:
@dapr/approvers-java-sdk @dapr/maintainers-java-sdk please review and provide feedback. We should not merge other PRs in durabletask-java until this is merged here to avoid divergence. 

For now, we will need to leave with low code coverage, as we are brining this module to the SDK, it will be easier to add tests as we move forward.

durabletask-client/src/test/java/io/dapr/durabletask/DurableTaskGrpcClientTlsTest.java
```diff
@@ -0,0 +1,342 @@
+///*
```
salaboy, line 1: If all the other changes are approved I am happy to fix this tests as this follows the same pattern as this PR: https://github.com/dapr/java-sdk/pull/1529 were we remove all these tests and related dependencies in favour of pre made certs. Please leave this file out of the overall review of this PR. 🙏

siri-varma:
@salaboy this PR LGTM. Should we exclude this module from the coverage ? Otherwise codecov will start failing for every PR.

salaboy:
> @salaboy this PR LGTM. Should we exclude this module from the coverage ? Otherwise codecov will start failing for every PR.

Yes, that is something we can do. I really want us to bring the testing level up for this module. 
There are still a couple of more things to sort out for this PR to be mergeable, but we are trying to get agreement on these changes.

salaboy:
> I believe we need to add to our release automation script if we proceed with this PR. Specifically we will need to add to the `update_sdk_version.sh` something like the following:
> 
> ```
> mvn versions:set -DnewVersion=$DAPR_JAVA_SDK_VERSION -f durabletask-client/pom.xml
> ```
> 
> Also, could you verify if we need changes in the `update_docs.sh` script for this new module, as I believe we will need changes there.

I've clean up the script to be simpler to maintain, https://github.com/dapr/java-sdk/pull/1575/files#diff-e0e9e1d7c713596b37959c8e4a8f9f5540b77c1a741e31e8c24d5243f35f113d It was getting confusing and there is really no need to change that for child modules, now that we don't have any alpha modules in the SDK. 

On the docs script, I don't think we need a change.

salaboy:
> @salaboy this PR LGTM. Should we exclude this module from the coverage ? Otherwise codecov will start failing for every PR.

Instaed of excluding this from codecov, what we should do is to set the current coverage number as the base for that moduel, and then improve from there. @siri-varma does tha make sense?

pom.xml
```diff
@@ -496,6 +491,10 @@
                 <goal>check</goal>
               </goals>
               <configuration>
+                <excludes>
+                  <exclude>io/dapr/durabletask/**/*</exclude>
+                  <exclude>io/dapr/springboot/examples/**/*</exclude>
```
cicoyle, line 496: Why exclude springboot examples in the durabletask migration PR?

salaboy: @cicoyle Durable tasks was the the first exclusion that we knew we needed to do, hence when I added durable tasks, we can have this low hanging fruit to improve test coverage numbers that really matter.

durabletask-client/src/test/java/io/dapr/durabletask/DurableTaskGrpcClientTlsTest.java
salaboy: Issue created that explains the migration needed for this test, please never enable this test as is again: https://github.com/dapr/java-sdk/issues/1604

salaboy:
@cicoyle issue created: https://github.com/dapr/java-sdk/issues/1604 notice that we don't need to bring back that test as it is, we need to migrated as explained in the issue.

PR #1577, opened 2025-10-22
If @component has value register activity with that value

javier-aliaga:
# Description
Enable registering activities with custom names. @component value will be use as registration name
## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/dapr-spring/dapr-spring-workflows/src/main/java/io/dapr/spring/workflows/config/DaprWorkflowsConfiguration.java
+++ b/dapr-spring/dapr-spring-workflows/src/main/java/io/dapr/spring/workflows/config/DaprWorkflowsConfiguration.java
@@ -58,13 +58,12 @@ public class DaprWorkflowsConfiguration implements ApplicationContextAware {
     for (WorkflowActivity activity :  workflowActivitiesBeans.values()) {
       // Get the @Component annotation
       Component componentAnnotation = activity.getClass().getAnnotation(Component.class);
-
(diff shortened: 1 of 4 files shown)


salaboy:
@javier-aliaga I would totally add a test to verify that if two Activities with the same name are registered from two different packages we know what happens. 
I do assume that if you just add a `@Component` without specifying the name it will use the FQN, if you add a name it will use that name instead, then leaving the responsibility to the user to assign unique names to the activities.

salaboy:
@kendallroden @javier-aliaga I also think this deteriorates the developer experience forcing the user to assign a name just to avoid displaying something in the UI.

javier-aliaga:
@salaboy We can try to register the activities with `com.longpackage.anotherpackage.lastpackage.MyClass` and `c.l.a.l.MyClass`

The main problem is how to make this not a breaking change

salaboy:
That's what I suggested to implement in the UI. We can do that in the SDK side, but if the user wants to see more info about what the abbreviated notation mean that is completely lost.

javier-aliaga:
> I also think this deteriorates the developer experience forcing the user to assign a name just to avoid displaying something in the UI.

@salaboy I think with the evolution of the workflows the way we call activities may not be possible anymore (it will be possible as long as the class is in the classpath). It will be the case that the class is not present on the client side as it is being to be executed by another app even in a different language

PR #1586, opened 2025-10-24
Add gRPC support to Dapr testcontainer

### LINKED ISSUE #1578: Add withAppProtocol to testcontainer
Hello,

I would like to be able to test my app's interactions with the dapr sidecar in an integration test. However, `io.dapr.testcontainers.DaprContainer` is currently hard-coded to use HTTP as the app protocol. How much effort would it take to add a `DaprContainer.withAppProtocol()` method so we can test with gRPC? 

If it seems like something a newbie like me could handle, let me know.

Thanks!

wlfgang:
# Description

Added a simple `withAppProtocol` method to the DaprContainer test container which allows the user to specify `--app-protocol` either `HTTP` or `GRPC`, allowing projects to exercise gRPC services from integration tests.

## Issue reference

This will close #1578 

## Checklist

Let me know if there's documentation that should be updated, I can add that.

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/testcontainers-dapr/src/test/java/io/dapr/testcontainers/DaprContainerTest.java
+++ b/testcontainers-dapr/src/test/java/io/dapr/testcontainers/DaprContainerTest.java
@@ -6,7 +6,6 @@ import org.testcontainers.utility.DockerImageName;
 import static io.dapr.testcontainers.DaprContainerConstants.DAPR_RUNTIME_IMAGE_TAG;
 import static io.dapr.testcontainers.DaprContainerConstants.DAPR_VERSION;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNull;
 
 public class DaprContainerTest {
 
@@ -81,22 +80,18 @@ public class DaprContainerTest {
   }
 
   @Test
-  public void appPortProtocolDefaultsTest() {
+  public void appProtocolDefaultsTest() {
     try (DaprContainer daprContainer = new DaprContainer(DAPR_RUNTIME_IMAGE_TAG)
             .withAppName("dapr-app")) {
       daprContainer.configure();
-      assertNull(daprContainer.getAppPort());
-      assertNull(daprContainer.getAppProtocol());
+      assertEquals(DaprProtocol.HTTP, daprContainer.getAppProtocol());
     }
 
-    Integer port = 50003;
     DaprProtocol protocol = DaprProtocol.GRPC;
     try (DaprContainer daprContainer = new DaprContainer(DAPR_RUNTIME_IMAGE_TAG)
             .withAppName("dapr-app4")
-            .withAppPort(port)
             .withAppProtocol(protocol)) {
       daprContainer.configure();
-      assertEquals(port, daprContainer.getAppPort());
       assertEquals(protocol, daprContainer.getAppProtocol());
     }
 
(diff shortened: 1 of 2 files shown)


mathecruz:
Hi @wlfgang thanks for this pull request, you need to [sign-off your commit](https://github.com/dapr/java-sdk/blob/master/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work).

wlfgang:
> Hi @wlfgang thanks for this pull request, you need to [sign-off your commit](https://github.com/dapr/java-sdk/blob/master/CONTRIBUTING.md#developer-certificate-of-origin-signing-your-work).

Thanks @mcruzdev -- I've added the DCO to the commit.

artur-ciocanu (approved):
Nicely done! Thanks a lot for your contribution @wlfgang!

artur-ciocanu:
@dapr/maintainers-java-sdk and @dapr/approvers-java-sdk let me know if you have any concerns merging this PR.

testcontainers-dapr/src/test/java/io/dapr/testcontainers/DaprContainerTest.java
```diff
@@ -78,4 +79,26 @@ public void appHealthParametersDefaultsTest(){
     assertEquals(3, dapr2.getAppHealthCheckThreshold());
 
   }
+
+  @Test
+  public void appPortProtocolDefaultsTest() {
+    try (DaprContainer daprContainer = new DaprContainer(DAPR_RUNTIME_IMAGE_TAG)
+            .withAppName("dapr-app")) {
+      daprContainer.configure();
+      assertNull(daprContainer.getAppPort());
```
salaboy: @wlfgang I think for backward compatibility and to make it easier for the user configuring the DaprContainer, HTTP should be still the default, hence if nothing is set you still get HTTP.

wlfgang: @salaboy I agree completely, and this is still true; HTTP will be used by default if the user doesn't call the method. For `withAppProtocol` I mirrored the existing implementation of `withAppPort` - if you don't call these methods, their fields won't be assigned, and the `--app-port` and `--app-protocol` arguments won't be specified to dapr. 

If you like, I can change it to initialize `appProtocol` rather than relying on a null, which may be considered better practice, and the behavior would remain the same.

salaboy: @wlfgang yes.. please let's make sure that by default it returns HTTP, unless GRPC is specified, that will helps us to keep the existing behavior.

artur-ciocanu:
@wlfgang could you please adjust the code based on @salaboy feedback.

Thank you.

testcontainers-dapr/src/test/java/io/dapr/testcontainers/DaprContainerTest.java
salaboy: The more I think about this the more I want to align with the defaults that are set on the Dapr Sidecar itself. For the protocol, by default if you don't set it is `http`, so when we create the container and use the getAppProtocol() it shoudn't return null.. it should return the same value that has been set on the sidecar.

wlfgang: Yeah, I was a little surprised when I wrote the test that `getAppPort` was returning null- I was expecting it to return the value of the default port. IMO getAppPort and getAppProtocol should have the same behavior (and maybe other methods as well).

wlfgang: Ok, it appears from the dapr docs that app port is actually undefined if you don't set it, so maybe null there does make sense. I have made the adjustment to stop using null for the protocol, so hopefully this should be all set.

wlfgang:
> @wlfgang could you please adjust the code based on @salaboy feedback.
> 
> Thank you.

Just made the fix, hopefully we should be good now!

marcduiker:
@holopin-bot @wlfgang Thank you! Here's a digital badge as a small token of appreciation.

Issue #1616, opened 2026-01-07
Flux Streaming subscriptions with generic type


alicejgibbons:
Today using Flux with streaming subscriptions in Java, you can only publish/subscribe using Dapr's CloudEvent type as shown in the example here: https://github.com/dapr/java-sdk/blob/8f119720c89d0c8f29a5d72e17d620f5f622170e/examples/src/main/java/io/dapr/examples/pubsub/stream/Subscriber.java#L46

The latest PR by @artur-ciocanu is an improvement since subscribeToEvents() uses a flux based method but the limitation here is that only the CloudEvent<T> can be sent.

We should also support something like the following as many users will be sending generic types (not cloud event based) on the message broker and this is supported in Dapr Pubsub generally.

```
public <T> Flux<T> subscribeToEvents(...
```

One way to standardize this would be to add an override that uses the `DaprAppCallbackProtos.TopicEventRequest topicEventRequest` type which can be deserialized as per the example here: https://github.com/artur-ciocanu/java-sdk/blob/26d215342bf541135c200ba22fd085b8a4e1c010/examples/src/main/java/io/dapr/examples/pubsub/grpc/SubscriberGrpcService.java#L67

alicejgibbons:
Note even the old `subscribeToEvents()` method that doesn't use Flux seems to convert to CloudEvents: https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/Subscription.java#L94

artur-ciocanu:
@alicejgibbons exactly, I tried to follow the current pattern for anything related to subscription. Since the Flux based API is pretty new, it still marked as "preview", I am wondering if there is any value in exposing the CloudEvent based API or should replace the current API with the generic one. I want to make that we keep the API surface small.

Let me know your thoughts and what customers are looking for.

alicejgibbons:
@artur-ciocanu Replacing the current API could be a good option since users could still use the Dapr Java CloudEvent object as the generic type anyways and get the same result. This also avoids the code around having to [rebuild the CoudEvent](https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/internal/subscription/EventSubscriberStreamObserver.java#L145) object which doesn't take into account *all* of the cloudevent fields [here](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-cloudevents/#dapr-generated-cloudevents-example).

[...]

PR #1617, opened 2026-01-10
Adding raw event subscription alongside CloudEvent subscription

artur-ciocanu:
# Description

Adding raw event subscription alongside CloudEvent subscription

## Issue reference

Please reference the issue this PR will close: #1616

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/pubsub/stream/PubSubStreamIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/pubsub/stream/PubSubStreamIT.java
@@ -43,6 +43,9 @@ public class PubSubStreamIT extends BaseIT {
   // sent by the runtime. When this was first added, the batch size in runtime was set to 10.
   private static final int NUM_MESSAGES = 100;
   private static final String TOPIC_NAME = "stream-topic";
+  private static final String TOPIC_NAME_FLUX = "stream-topic-flux";
+  private static final String TOPIC_NAME_CLOUDEVENT = "stream-topic-cloudevent";
+  private static final String TOPIC_NAME_RAWPAYLOAD = "stream-topic-rawpayload";
   private static final String PUBSUB_NAME = "messagebus";
 
   private final List<DaprRun> runs = new ArrayList<>();
@@ -139,17 +142,17 @@ public class PubSubStreamIT extends BaseIT {
       // Publish messages
       for (int i = 0; i < NUM_MESSAGES; i++) {
         String message = String.format("Flux message #%d for run %s", i, runId);
-        client.publishEvent(PUBSUB_NAME, TOPIC_NAME, message).block();
+        client.publishEvent(PUBSUB_NAME, TOPIC_NAME_FLUX, message).block();
         System.out.println(
-            String.format("Published flux message: '%s' to topic '%s'", message, TOPIC_NAME));
+            String.format("Published flux message: '%s' to topic '%s'", message, TOPIC_NAME_FLUX));
       }
 
-      System.out.println("Starting Flux subscription for " + TOPIC_NAME);
+      System.out.println("Starting Flux subscription for " + TOPIC_NAME_FLUX);
 
       Set<String> messages = Collections.synchronizedSet(new HashSet<>());
 
       // subscribeToEvents now returns Flux<T> directly (raw data)
-      var disposable = previewClient.subscribeToEvents(PUBSUB_NAME, TOPIC_NAME, TypeRef.STRING)
+      var disposable = previewClient.subscribeToEvents(PUBSUB_NAME, TOPIC_NAME_FLUX, TypeRef.STRING)
           .doOnNext(rawMessage -> {
             // rawMessage is String directly
             if (rawMessage.contains(runId)) {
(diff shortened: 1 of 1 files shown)


artur-ciocanu:
@dapr/approvers-java-sdk and @dapr/maintainers-java-sdk could you please take a look and approve. This has been requested by @alicejgibbons.

One thing I want to mention, I am not really fond of the new name `subscribeToEventsData(...)` I couldn't think of a better name.

An alternative would be to rename the old method to `subscribeToCloudEvents(...)` that uses `CloudEvent` and then have `subscribeToEvents(...)` return generic type.

javier-aliaga:
> @dapr/approvers-java-sdk and @dapr/maintainers-java-sdk could you please take a look and approve. This has been requested by @alicejgibbons.
> 
> One thing I want to mention, I am not really fond of the new name `subscribeToEventsData(...)` I couldn't think of a better name.
> 
> An alternative would be to rename the old method to `subscribeToCloudEvents(...)` that uses `CloudEvent` and then have `subscribeToEvents(...)` return generic type.

Another possibility is naming the new one `SubscribeToRawEvent` or similar

javier-aliaga (dismissed):

siri-varma (dismissed):

sdk/src/main/java/io/dapr/client/DaprClientImpl.java
```diff
@@ -514,6 +514,15 @@ public <T> Flux<CloudEvent<T>> subscribeToEvents(String pubsubName, String topic
     }, FluxSink.OverflowStrategy.BUFFER);
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public <T> Flux<T> subscribeToEventsData(String pubsubName, String topic, TypeRef<T> type) {
+    return subscribeToEvents(pubsubName, topic, type)
+        .map(CloudEvent::getData);
```
alicejgibbons: @artur-ciocanu does this line not force the type to deserialize into a CloudEvent?

alicejgibbons:
Regardinging renaming, can we have multiple overloads to subscribeToEvents?

artur-ciocanu:
> Regardinging renaming, can we have multiple overloads to subscribeToEvents?

@alicejgibbons unfortunately we can not have the same overloads that depends just on generic type.

@alicejgibbons as I was thinking more and more. Instead of coming up with different APIs, customers can leverage Flux via a simple `map` operator. It is trivial:
```
Flux<SomeEvent> stream = daprClient.subscribeToEvents(pubsubName, topic, type)
        .map(CloudEvent::getData);
```
It is an additional step, but it is not too bad.

Please let me know your thoughts.

CC: @javier-aliaga @cicoyle

sdk-tests/src/test/java/io/dapr/it/pubsub/stream/PubSubStreamIT.java
```diff
@@ -124,4 +124,48 @@ public void onError(RuntimeException exception) {
...
+        60000));
+
+    var runId = UUID.randomUUID().toString();
+    try (DaprClient client = daprRun.newDaprClient();
+         DaprPreviewClient previewClient = daprRun.newDaprPreviewClient()) {
+
+      // Publish messages
+      for (int i = 0; i < NUM_MESSAGES; i++) {
+        String message = String.format("Raw message #%d for run %s", i, runId);
+        client.publishEvent(PUBSUB_NAME, TOPIC_NAME, message).block();
```
alicejgibbons: This is still publishing a cloudEvent by default since you are not adding the raw payload header. Thus needs to be changed to something like the following:

```
PublishEventRequest publishEventRequest = new PublishEventRequest(pubsubName, topicName, eventType);
publishEventRequest.setContentType(APPLICATION_EVENT_TYPES_JSON);
headers.put("rawPayload", "true"); 
publishEventRequest.setMetadata(headers);
client.publishEvent(publishEventRequest).block();
```

sdk/src/main/java/io/dapr/client/DaprClientImpl.java
artur-ciocanu: I went a different route.

sdk-tests/src/test/java/io/dapr/it/pubsub/stream/PubSubStreamIT.java
artur-ciocanu: I went a different route

artur-ciocanu:
@alicejgibbons @cicoyle @javier-aliaga I went a different route, since this is not released yet, I think a breaking change is OK.

The final API is the following:
```
<T> Flux<T> subscribeToEvents(String pubsubName, String topic, TypeRef<T> type);

<T> Flux<T> subscribeToEvents(String pubsubName, String topic, TypeRef<T> type, Map<String, String> metadata);
```

There are different examples in the integrations tests for raw payload and CloudEvent.

Let me know your thoughts.

javier-aliaga:
> @alicejgibbons @cicoyle @javier-aliaga I went a different route, since this is not released yet, I think a breaking change is OK.
> 
> The final API is the following:
> 
> ```
> <T> Flux<T> subscribeToEvents(String pubsubName, String topic, TypeRef<T> type);
> 
> <T> Flux<T> subscribeToEvents(String pubsubName, String topic, TypeRef<T> type, Map<String, String> metadata);
> ```
> 
> There are different examples in the integrations tests for raw payload and CloudEvent.
> 
> Let me know your thoughts.

Looks good to me, it feels more simple.

cicoyle (commented):
Can you review the tests bc CI is failing with:
```
Error:  Errors: 
Error:    PubSubStreamIT.testPubSub:115 » Runtime org.opentest4j.AssertionFailedError: expected: <100> but was: <95>
Error:    PubSubStreamIT.testPubSubCloudEvent:207 » Runtime org.opentest4j.AssertionFailedError: expected: <100> but was: <0>
```

artur-ciocanu:
@cicoyle @alicejgibbons and @javier-aliaga the PR is looking good and the build is 🟢.

Could you please take another look. Thank you.

PR #1623, opened 2026-01-21
Add PATCH to HttpMethods enum for service invocation

### LINKED ISSUE #1622: Add PATCH to HttpMethods enum for service invocation
## Problem

The `HttpMethods` enum in `DaprHttp` does not include `PATCH`, which prevents using `DaprClient.invokeMethod()` with PATCH requests through HTTPEndpoints.

## Use Case

When calling external REST APIs via Dapr HTTPEndpoint and service invocation, PATCH is commonly required for partial updates. For example, Salesforce REST API requires PATCH for updating records.

## Current Workaround

Developers must bypass the SDK and make direct HTTP calls to the Dapr sidecar:

```java
var url = "http://localhost:" + daprPort + "/v1.0/invoke/" + appId + "/method/" + path;
var request = HttpRequest.newBuilder()
    .uri(URI.create(url))
    .method("PATCH", BodyPublishers.ofString(body))
    .build();
httpClient.send(request, BodyHandlers.ofString());
```

This loses the benefits of the SDK (type safety, error handling, reactive patterns).

## Proposed Solution

Add `PATCH` to the `HttpMethods` enum in `DaprHttp.java`:

```java
public enum HttpMethods {
  NONE,
  GET,
  PUT,
  POST,
  DELETE,
  HEAD,
  CONNECT,
  OPTIONS,
  TRACE,
  PATCH  // Add this
}
```

## Environment

- Java SDK version: 1.12.0
- Dapr runtime: 1.14.x

lindner:
## Summary
- Adds `PATCH` to the `HttpMethods` enum to enable using `DaprClient.invokeMethod()` with PATCH requests through HTTPEndpoints

## Test plan
- [x] SDK compiles successfully
- [x] Verify PATCH method works with `invokeMethod()` for HTTPEndpoint service invocation

Fixes #1622

### DIFF
--- a/sdk-tests/src/test/java/io/dapr/it/spring/messaging/DaprSpringMessagingIT.java
+++ b/sdk-tests/src/test/java/io/dapr/it/spring/messaging/DaprSpringMessagingIT.java
@@ -19,6 +19,7 @@ import io.dapr.spring.messaging.DaprMessagingTemplate;
 import io.dapr.testcontainers.Component;
 import io.dapr.testcontainers.DaprContainer;
 import io.dapr.testcontainers.DaprLogLevel;
+import io.dapr.testcontainers.wait.strategy.DaprWait;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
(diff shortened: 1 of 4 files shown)


salaboy:
@lindner thanks for the contribution. We will definitely need a test for this. I would recommend you an integration test too using test container to validate that the change works using the sidecar.

lindner:
> @lindner thanks for the contribution. We will definitely need a test for this. I would recommend you an integration test too using test container to validate that the change works using the sidecar.

Can you help me identify the correct place for this?  It's just another HTTP verb, so I don't think it's going to need much...

salaboy:
@lindner I think adding a test here will be enough: https://github.com/dapr/java-sdk/blob/master/sdk/src/test/java/io/dapr/client/DaprHttpTest.java#L247

Issue #1631, opened 2026-01-29
Allow to use shutdown now then closing  a `WorkflowRuntime`


mcruzdev:
# Proposal

It would be great if we have a way to close "now" all `WorkflowRuntime` resources. 

It is intended to be used in tests or live reload approaches.

javier-aliaga:
@mcruzdev @siri-varma I think this PR https://github.com/dapr/java-sdk/pull/1727 will help to mitigate the problem as the close should be much faster now

PR #1632, opened 2026-01-29
Introduce closeNow() to allow a "now" shutdown

mcruzdev:
# Description

Add a new method in `WorkflowRuntime` class, allowing a shutdown "now" approach.

## Issue reference

Please reference the issue this PR will close: #1631 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


sdk-workflows/src/main/java/io/dapr/workflows/runtime/WorkflowRuntime.java
```diff
@@ -72,6 +73,14 @@ public void close() {
     this.closeSideCarChannel();
   }
 
+  /**
+   * Immediately shuts down all resources associated with this instance.
+   */
+  public void closeNow() {
```
siri-varma, line 80: @mcruzdev When would one use close vs closeNow ?

mathecruz: Hi @siri-varma, when we have a I live reload feature or when I am running tests.

siri-varma: Should we be handling the interrupted exception in this method ?

Apart from that everything looks good

PR #1640, opened 2026-02-05
feat: Add new fields to conversation api

javier-aliaga:
# Description

Add new fields for conversation api

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/1639

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/durabletask-client/pom.xml
+++ b/durabletask-client/pom.xml
@@ -60,6 +60,10 @@
       <artifactId>jackson-datatype-jsr310</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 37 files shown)


salaboy:
/LGTM .. all green so I think we can merge this

javier-aliaga:
@siri-varma the problem is these new fields are only supported by some components but not the echo service. 

I modified the examples to print the usage in case the model is present. 

Let me work on the ResponseFormat, I would like to do a real test

javier-aliaga:
@siri-varma I modified the examples to add the new fields, bear in mind this fields only works with a real llm, no changes expected using the echo service

PR #1641, opened 2026-02-05
[1.17] make bulk pubsub stable

cicoyle:
update all tests/examples/code to the stable API and fallback to alpha API if UNIMPLEMENTED


cicoyle:
@javier-aliaga @salaboy @siri-varma @artur-ciocanu 

please have a look :)

Issue #1625, opened 2026-01-22
Support actor reminder failure policy


JoshVanL:
Support actor reminder failure policy as like the go-sdk

https://github.com/dapr/go-sdk/pull/793

PR #1643, opened 2026-02-06
feat: Add failure policy to actor reminders

javier-aliaga:
# Description

Support actor reminder failure policy

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/1625

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk-actors/src/main/java/io/dapr/actors/runtime/ActorObjectSerializer.java
+++ b/sdk-actors/src/main/java/io/dapr/actors/runtime/ActorObjectSerializer.java
@@ -16,9 +16,8 @@ package io.dapr.actors.runtime;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import io.dapr.client.ObjectSerializer;
-import io.dapr.client.domain.ConstantFailurePolicy;
-import io.dapr.client.domain.DropFailurePolicy;
 import io.dapr.client.domain.FailurePolicy;
 import io.dapr.utils.DurationUtils;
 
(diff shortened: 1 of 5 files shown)


sdk-actors/src/main/java/io/dapr/actors/runtime/ActorObjectSerializer.java
```diff
@@ -243,7 +267,28 @@ private ActorReminderParams deserializeActorReminder(byte[] value) throws IOExce
     Duration period = extractDurationOrNull(node, "period");
     byte[] data = node.get("data") != null ? node.get("data").binaryValue() : null;
 
-    return new ActorReminderParams(data, dueTime, period);
+    // Handle failure policy if present
+    JsonNode failurePolicyNode = node.get("failurePolicy");
```
siri-varma, line 262: @javier-aliaga What are your thoughts on creating a class and deserialize this failurePolicy into that ??

javier-aliaga: I added the  serialization and de-serialization using Jackson. I am not fully convinced on this last commit, 

Pros of Jackson:
 - No manual conversion
 - Clean code

Cons:
 - New dependency for Duration
 - Jackson annotations in the parent domain package
 - Mix of serialization modes in the serializer 

Tbh, I did consider this in my first implementation and I decided to go with the manual conversion, I give it a try a Jackson but it stills feel I like more the first approach. wdyt?

siri-varma: Thank you for putting out the Pros and Cons.

Given the amount of code required without the library, I think these trade-offs are acceptable. Are you concerned about any performance issues because of mixing serialization/annotations ?

javier-aliaga: Not really, just worry about side effects of the annotations in a different module, but should be fine. Lets keep with the new code. Thanks for your input here 👍

PR #1666, opened 2026-02-19
chore: Deprecate invoke methods

javier-aliaga:
# Description

The guidance since 2023 has been to not use SDK invoke* methods for Service Invocation. However this has never been formalised by marking the methods as deprecated.

Instead it is recommended to use language native http clients and grpc clients for Service Invoke.

Factory methods in the SDK which return a native http/grpc client is fine too, and are not deprecated.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/1642

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


siri-varma:
@javier-aliaga quick question, do we need to update the javadoc on the pr suggesting the new approach ? May be it would good to update the examples as well ?

javier-aliaga:
@olitomlinson this will not be part of 1.17, as Siri pointed we have tests and examples to update and it will take some time to replace them. I will plan this for 1.18

cicoyle:
This was called out on discord, but wanting to mention it here as well - we are holding off on this until examples and docs are updated.

PR #1689, opened 2026-03-09
Create the release + release notes automatically

cicoyle:
Before this PR, this was a manual step after releasing. I am adding it to be created automatically for us to reduce the manual steps taken while releasing. I added the backport label so if we need to cut a patch for 1.17 it should auto generate the release and release notes for us.


javier-aliaga:
@cicoyle can you review copilots comments?

cicoyle:
> @cicoyle can you review copilots comments?

done

PR #1695, opened 2026-03-11
feat: add application-level keepalive to prevent ALB idle connection timeouts

JoshVanL:
AWS ALBs do not forward HTTP/2 PING frames, causing idle gRPC connections to be closed. This adds a background loop that periodically calls the existing Hello RPC as application-level traffic to keep the connection alive through L7 load balancers.

### DIFF
--- a/durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
+++ b/durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
@@ -13,6 +13,7 @@ limitations under the License.
 
 package io.dapr.durabletask;
 
+import com.google.protobuf.Empty;
 import io.dapr.durabletask.implementation.protobuf.OrchestratorService;
 import io.dapr.durabletask.implementation.protobuf.TaskHubSidecarServiceGrpc;
 import io.dapr.durabletask.orchestration.TaskOrchestrationFactories;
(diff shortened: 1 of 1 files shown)


javier-aliaga:
ey @JoshVanL why is this inside the durable worker? I think it does not belong there but on the class we build the channel so all component will benefit from it not just the worflow

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -175,33 +180,39 @@ public void startAndBlock() {
...
-                    activityRequest.getOrchestrationInstance().getInstanceId(),
-                    Context.current()));
-
-            this.workerPool.submit(new ActivityRunner(workItem, taskActivityExecutor, sidecarClient, tracer));
-
-          } else {
-            logger.log(Level.WARNING,
-                "Received and dropped an unknown '{0}' work-item from the sidecar.",
-                requestType);
+        startKeepaliveLoop();
```
javier-aliaga, line 183: Can you move this to where the channel is created instead?

javier-aliaga:
@JoshVanL I take over this PR, I see why you add it to the durabletask, but I will refactor it a bit, add tests and also an opt-out config property

javier-aliaga:
Superseding PR https://github.com/dapr/java-sdk/pull/1795

PR #1727, opened 2026-04-14
fix(durabletask): detect shutdown and break gRPC worker loop

### LINKED ISSUE #1728: DurableTaskGrpcWorker loop does not exit on close() or stop()
## Expected Behavior

Calling `close()` or `stop()` on `DurableTaskGrpcWorker` should promptly terminate the `startAndBlock()` worker loop, regardless of whether the thread is sleeping during a retry or blocked on the gRPC `getWorkItems()` streaming call.

## Actual Behavior

The `startAndBlock()` method uses a `while (true)` loop that never checks for shutdown signals. The loop only exits if an `InterruptedException` is caught during the 5-second `Thread.sleep()` retry window.

When `close()` is called while the thread is blocked on the gRPC stream, `managedSidecarChannel.shutdownNow()` produces a `StatusRuntimeException` with `Status.Code.CANCELLED`. The catch block logs the disconnection but falls through to the retry sleep, causing the loop to continue indefinitely — reconnecting to a channel that has already been shut down.

Additionally, the `InterruptedException` catch swallows the thread's interrupt status (does not re-set it), breaking the interrupt contract for upstream callers.

## Steps to Reproduce the Problem

1. Create a `DurableTaskGrpcWorker` and call `start()` (which runs `startAndBlock()` in a background thread)
2. Wait for the worker to connect to the sidecar and begin processing work items via the blocking `getWorkItems()` stream
3. Call `close()` or `stop()` on the worker
4. Observe that the worker thread does not terminate — it logs `"Durable Task worker has disconnected"`, sleeps 5 seconds, and retries the connection in a loop

## Release Note

RELEASE NOTE: **FIX** DurableTaskGrpcWorker loop now exits promptly on close()/stop().

javier-aliaga:
# Description

## Summary
- Fix `DurableTaskGrpcWorker.startAndBlock()` loop not exiting when `close()`/`stop()` is called while the gRPC stream is blocking
- The `while (true)` loop only broke on `InterruptedException` during the 5s retry sleep — a `CANCELLED` from channel shutdown was logged but retried indefinitely

### durabletask-client — `DurableTaskGrpcWorker`
- Replace `while (true)` with `while (!isNormalShutdown && !Thread.currentThread().isInterrupted())` to check shutdown signals each iteration
- Break out of the retry path when `CANCELLED` is received during a normal shutdown, avoiding a stale 5s sleep cycle
- Re-set the thread interrupt flag before breaking on `InterruptedException` to preserve the interrupt contract for upstream callers

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/java-sdk/issues/1728

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


durabletask-client/src/test/java/io/dapr/durabletask/DurableTaskGrpcWorkerShutdownTest.java
```diff
@@ -0,0 +1,128 @@
...
+   * rather than hanging in the retry loop.
+   */
+  @Test
+  void workerThreadTerminatesPromptlyOnClose() throws Exception {
+    // Use an arbitrary port where no sidecar is running — the worker will
+    // enter the retry loop (UNAVAILABLE → sleep 5s → retry).
+    DurableTaskGrpcWorker worker = new DurableTaskGrpcWorkerBuilder()
+        .port(19876)
+        .build();
+
```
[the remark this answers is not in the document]
javier-aliaga: ports 19876-19878 are unlikely to collide, and the worker handles connection failure gracefully. Ephemeral ports add complexity for marginal gain

durabletask-client/src/test/java/io/dapr/durabletask/DurableTaskGrpcWorkerShutdownTest.java
```diff
@@ -0,0 +1,128 @@
...
+    Thread workerThread = getWorkerThread(worker);
+    if (workerThread != null) {
+      workerThread.join(Duration.ofSeconds(3).toMillis());
+      assertFalse(workerThread.isAlive(),
+          "Worker thread should have terminated after close()");
+    }
+
+    Duration elapsed = Duration.between(before, Instant.now());
+    assertTrue(elapsed.toMillis() < 3000,
+        "close() should return promptly, but took " + elapsed.toMillis() + "ms");
```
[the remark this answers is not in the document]
javier-aliaga: we do not do it

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
siri-varma, line 180: Should we have a check here as well ?

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -210,15 +211,23 @@ public void startAndBlock() {
               this.getSidecarAddress());
         } else if (e.getStatus().getCode() == Status.Code.CANCELLED) {
           logger.log(Level.INFO, "Durable Task worker has disconnected from {0}.", this.getSidecarAddress());
+          if (this.isNormalShutdown) {
```
siri-varma: We already added the check to the while loop. Just curoius what is the benefit of adding the if check on this line and 222 ?

javier-aliaga: both of them are to not wait the 5 seconds on line 228, but we do not need this one as it will be checked on line 222

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
javier-aliaga: yes!

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
```diff
@@ -171,12 +172,15 @@ public void startAndBlock() {
         this.dataConverter,
         logger);
 
-    while (true) {
+    while (!this.isNormalShutdown && !Thread.currentThread().isInterrupted()) {
       try {
         OrchestratorService.GetWorkItemsRequest getWorkItemsRequest = OrchestratorService.GetWorkItemsRequest
             .newBuilder().build();
         Iterator<OrchestratorService.WorkItem> workItemStream = this.sidecarClient.getWorkItems(getWorkItemsRequest);
         while (workItemStream.hasNext()) {
```
siri-varma, line 180:
```suggestion
        while (workItemStream.hasNext() && (this.isNormalShutdown || Thread.currentThread().isInterrupted())) {
```

siri-varma: We can remove the separate if check ? What do you think ?

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
siri-varma: Do you think there would be any issue if we wait for 5 seconds ?

siri-varma (approved):

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
javier-aliaga: It should not, but in deployments with several instances you have this constraint if you need to wait the pod to stop gracefully taking more time than you need

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcWorker.java
javier-aliaga: uff, honestly I prefer keeping them separate, mixing &&, || and () in the same condition makes them hard to read.

Issue #1732, opened 2026-04-20
Consolidating all Flaky Test issues into one single issue


siri-varma:
## Tracking: Flaky Tests and CI Reliability

This is a consolidated tracking issue for all known flaky test and CI reliability problems in the Java SDK. Flaky tests erode confidence in CI, slow down PR merges, and waste maintainer time on re-runs.

### Related Issues

| Issue | Area | Status |
|-------|------|--------|
| #725  | Integration tests (general) | Open |
| #1258 | CI build step — `dapr stop` failure | Open |
| #1321 | Actor tests | Open |
| #1367 | Jobs tests (`DaprJobsIT`) | Open |
| #1603 | PubSub Outbox tests (`DaprPubSubOutboxIT`) | Open |

#### 1. Flaky Integration Tests (#725)
The original umbrella issue. Multiple integration tests have historically been unreliable, causing spurious CI failures across different test suites.

#### 2. CI Build Step Failure — App Stop Race Condition (#1258)
The CI build sometimes fails because `dapr stop` is called on an app that has already exited. The error `couldn't find app id` causes the build step to fail even though the test itself passed. Failing to stop an already-exited app should not fail the build.

#### 3. Flaky Actor Tests (#1321)
Actor integration tests intermittently fail in CI. Linked CI run: https://github.com/dapr/java-sdk/actions/runs/14605088504/job/40972206420?pr=1235

#### 4. Flaky Jobs Test — `DaprJobsIT.testJobScheduleCreationWithAllParameters` (#1367)
This test fails intermittently with:

https://github.com/dapr/java-sdk/issues/1603
https://github.com/dapr/java-sdk/issues/1367
https://github.com/dapr/java-sdk/issues/1321
https://github.com/dapr/java-sdk/issues/1258
https://github.com/dapr/java-sdk/issues/725

siri-varma:
Fixed all the flaky tests for now. We can reopen if we see the issues again.

PR #1741, opened 2026-05-10
[Phase 1] Fix Flaky Integration Tests + Migrate some tests to Test Containers

siri-varma:
# Description

## Issue reference

Please reference the issue this PR will close: 
https://github.com/dapr/java-sdk/issues/1732

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


siri-varma:
@javier-aliaga Would you know why this is failing ? did not touch any protos or xml files

javier-aliaga:
@siri-varma , the proto files are generated from master dapr/dapr, every time they change something there it break our master. I already fixed that and it is merged to master

siri-varma:
@javier-aliaga need some reviews on this pr please. Mostly test changes.

* Reduce flakiness
* Move to test containers

sdk/src/main/java/io/dapr/client/Subscription.java
```diff
@@ -88,6 +88,7 @@ public class Subscription<T> implements Closeable {
     });
 
     this.receiver = new Thread(() -> {
+      int reconnectAttempts = 0;
```
javier-aliaga, line 91: When do we reset this counter?

sdk/src/main/java/io/dapr/client/Subscription.java
```diff
@@ -124,6 +125,7 @@ public void onNext(DaprPubsubProtos.SubscribeTopicEventsResponseAlpha1 topicEven
           @Override
           public void onError(Throwable throwable) {
             listener.onError(DaprException.propagate(throwable));
+            receiverStateChange.release();
```
javier-aliaga, line 128: This is the right fix but it introduces a problem in line 121.

sdk/src/main/java/io/dapr/client/Subscription.java
javier-aliaga, line 122: This call is a sync call, no grpc stream call, this releases the lock but is not telling the stream to be closed. 

2 possibilities here;

Do we want to kill the stream because one event fails? or we just notifiy the listener and we keep going?

```
} catch (Exception e) {
    listener.onError(DaprException.propagate(e));
    // optionally ack as RETRY so the broker can redeliver
    // do NOT call this.onError and do NOT release the semaphore
  }
```

If we want to close the stream then
```
  } catch (Exception e) {
    var s = streamRef.get();
    if (s != null) {
      try {
        s.onError(DaprException.propagate(e));  // client-side cancel of the gRPC stream
      } catch (Throwable ignore) {
        // already terminated
      }
    }
    // do NOT call this.onError(...) or release the semaphore here —
    // gRPC will invoke the server→client onError callback on cancellation,
    // which is the path at line 126, and that one releases the semaphore.
  }
```

I prefer the listener option, wdyt?

sdk/src/main/java/io/dapr/client/Subscription.java
```diff
@@ -142,6 +144,17 @@ public void onCompleted() {
           Thread.currentThread().interrupt();
           running.set(false);
         }
+
+        if (running.get()) {
+          long backoffMs = Math.min(1000L * (1L << reconnectAttempts), 30000L);
```
javier-aliaga, line 149: if the server is broken for a long period and you get the attemp 63 then it becomes negative causing this to throw an kill the receiver thread, so we may need to cap it to 5

```
  int safeAttempts = Math.min(reconnectAttempts, 5);
  long backoffMs = Math.min(1000L * (1L << safeAttempts), 30000L);
```

5 attempts = 32000 ms

siri-varma:
https://github.com/dapr/java-sdk/pull/1754

Closing this PR in favor or above

PR #1754, opened 2026-05-25
Fix Flaky Integration Tests + Migrate some tests to Test Containers

siri-varma:
### What this PR does

1. **Fixes flakes** in the existing sdk-tests harness (actor registration race, tracing span-name mismatch, streaming subscription reconnect).
2. **Migrates 13 ITs** off the legacy `BaseIT`/`DaprRun` harness (which requires a host-installed Dapr CLI + `dapr init`) onto Testcontainers via `DaprContainer`, matching the spring-boot-4-sdk-tests module.

### Notable behavior fixes

- **Actor IT registration timeout** — `BaseContainerIT.startAppAndAttach` now starts `AppRun` before the daprd container, with daprd's HTTP/gRPC host ports pinned (via `setPortBindings`) to the same `DaprPorts` values the app's env already received. Previously daprd would probe the app channel before the JVM had bound the host port, fetch `/dapr/config` against the SSH bridge, get nothing, report actor types `[]` to placement, and never re-query.
- **Tracing span name** — Zipkin records `calllocal/<APP_NAME>/<method>`; the assertion expected the bogus `<APP_NAME>-service` suffix. Fixed in both grpc and http TracingITs.
- **Streaming subscription reconnect** (`sdk/src/main/java/io/dapr/client/Subscription.java`) — receiver thread now releases the state-change semaphore on `onError`, sleeps with a doubled-long backoff (1s → 30s cap, no overflow path), and is interrupted on `close()` so shutdown isn't blocked by a backoff sleep.

### Files deleted in this PR — and why

| File | Lines | Replacement | Notes |
|---|---|---|---|
| `pubsub/http/PubSubIT.java` | 723 | `testcontainers/pubsub/http/DaprPubSubIT.java` | Legacy `BaseIT` harness. Method mapping: `publishPubSubNotFound`→`shouldReceiveInvalidArgument`, `testBulkPublishPubSubNotFound`→`shouldReceiveInvalidArgumentWithBulkPublish`, `testBulkPublish`→`shouldPublishSomePayloadTypesWithNoError`, `testPubSub`→`testPubSub`, `testPubSubBinary`→`shouldPublishBinary`, `testPubSubTTLMetadata`→`testPubSubTTLMetadata`, `testLongValues`→`testLongValues`. **`testPubSubBulkSubscribe` was missing — added on the migrated `DaprPubSubIT` in this PR.** |
| `pubsub/http/SubscriberController.java` | 265 | (n/a) | Companion subprocess for legacy `PubSubIT`. The Testcontainers IT runs the Spring Boot subscriber inside the test JVM, so a separate process file isn't needed; a different `SubscriberController` already lives under `testcontainers/pubsub/http/`. |
| `pubsub/http/SubscriberService.java` | 46 | (n/a) | `main()` entry-point that spawned the legacy subscriber. No replacement needed. |
| `pubsub/stream/PubSubStreamIT.java` | 339 | `testcontainers/pubsub/stream/DaprPubSubStreamIT.java` | **`testPubSubDeadLetterTopic` was missing — added on the migrated `DaprPubSubStreamIT` in this PR.** |
| `state/HelloWorldClientIT.java` | 74 | `state/GRPCStateClientIT.rawGrpcStubGetAndDeleteState` | Single test of `DaprClient.newGrpcStub(...)` (raw `DaprGrpc.DaprBlockingStub` save/get/delete). **Restored as a method on the migrated `GRPCStateClientIT`** — no companion subprocess needed. |
| `state/HelloWorldGrpcStateService.java` | 68 | (n/a) | Companion subprocess for `HelloWorldClientIT`. |

### Mongo state store

`AbstractStateClientIT.saveAndQueryAndDeleteState` (Dapr preview Query State API, requires Mongo) was briefly `@Disabled` during the migration since Redis doesn't support that API and we'd dropped the docker-compose Mongo service. Now restored: `SharedTestInfra.mongo()` runs a pinned `mongo:7` Testcontainer, and `BaseContainerIT.mongoStateStore(...)` registers a `state.mongodb` component on `GRPCStateClientIT`.

### Container image pinning

`SharedTestInfra` now pins all three shared containers (`redis:7-alpine`, `mongo:7`, `openzipkin/zipkin:3.4`) instead of letting any of them float on `:latest`.

### ITs intentionally NOT migrated (per spec Non-Goals)

`BindingIT`, `ActorReminderFailoverIT`, `ActorReminderRecoveryIT`, `ActorTimerRecoveryIT`, `ActorStateIT`, `WaitForSidecarIT`, `ActorSdkResiliencyIT`, plus 2 ITs in `durabletask-client/`. These remain on the legacy `BaseIT`/`DaprRun` harness.

### Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


siri-varma:
@javier-aliaga / @dapr/maintainers-java-sdk / @cicoyle This PR reduce some flakiness when running integration tests. I ran 5-6 times and all of them succeeded.

Could you folks please review ? Thank you.

siri-varma:
Looks like last commit is having some issues

[Restore parity for two legacy ITs deleted earlier](https://github.com/dapr/java-sdk/pull/1754/commits/7204e81c68473ad693ffc1dcfde0de4265b703b6)

Let me look into that

siri-varma:
@javier-aliaga / @cicoyle / @dapr/maintainers-java-sdk This PR is ready to be reveiwed. Ran the tests 5-6 times and all the instances succeeded.

So, we should definitely see a reduction in number of failures due to flakiness

siri-varma:
I think next task would be to see how we can reduce the tests runtime to lower than 20 mins

siri-varma:
@javier-aliaga / @cicoyle Just a bump on this PR. Would like to merge and get as many runs as possible to get data points

sdk/src/main/java/io/dapr/client/Subscription.java
```diff
@@ -142,6 +144,17 @@ public void onCompleted() {
...
+
+        if (running.get()) {
+          try {
+            Thread.sleep(backoffMs);
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            running.set(false);
+          }
+          // Double the backoff for the next reconnect, capped at 30s.
+          backoffMs = Math.min(backoffMs * 2, 30_000L);
```
javier-aliaga, line 161: Where do we reset the backoff?

sdk/src/main/java/io/dapr/client/Subscription.java
```diff
@@ -124,6 +125,7 @@ public void onNext(DaprPubsubProtos.SubscribeTopicEventsResponseAlpha1 topicEven
           @Override
           public void onError(Throwable throwable) {
             listener.onError(DaprException.propagate(throwable));
+            receiverStateChange.release();
```
javier-aliaga: this still applies https://github.com/dapr/java-sdk/pull/1741/changes#r3265853071

siri-varma: Repasting the comment for easy reference. Apologies for missing it

```
This call is a sync call, no grpc stream call, this releases the lock but is not telling the stream to be closed.

2 possibilities here;

Do we want to kill the stream because one event fails? or we just notifiy the listener and we keep going?

} catch (Exception e) {
    listener.onError(DaprException.propagate(e));
    // optionally ack as RETRY so the broker can redeliver
    // do NOT call this.onError and do NOT release the semaphore
  }
If we want to close the stream then

  } catch (Exception e) {
    var s = streamRef.get();
    if (s != null) {
      try {
        s.onError(DaprException.propagate(e));  // client-side cancel of the gRPC stream
      } catch (Throwable ignore) {
        // already terminated
      }
    }
    // do NOT call this.onError(...) or release the semaphore here —
    // gRPC will invoke the server→client onError callback on cancellation,
    // which is the path at line 126, and that one releases the semaphore.
  }
I prefer the listener option, wdyt?
```

siri-varma: Addressed

siri-varma:
@javier-aliaga Addressed the comments on this

sdk/src/main/java/io/dapr/client/Subscription.java
siri-varma: Fixed it now

sdk/src/main/java/io/dapr/client/Subscription.java
```diff
@@ -117,13 +120,14 @@ public void onNext(DaprPubsubProtos.SubscribeTopicEventsResponseAlpha1 topicEven
                 }
               });
             } catch (Exception e) {
-              this.onError(DaprException.propagate(e));
+              listener.onError(DaprException.propagate(e));
             }
```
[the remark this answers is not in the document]
javier-aliaga: this is already present in master, we can fix it in a following pr

javier-aliaga (approved):
I will approve, copilot comments are in test file, I will let up to you to fix it or leave as is, I have removed the auto-merge so you can take a look and click the button if you want

siri-varma:
I will address them in the subsequent pr. Thank you for the quick review Javier.

PR #1758, opened 2026-05-28
feat: attach SDK User-Agent

javier-aliaga:
# Description

newRequestBuilder now sets the User-Agent header to the SDK version, matching what the deprecated invokeMethod path emitted via DaprHttp. This is the only default the SDK can supply that callers cannot, and it is useful for runtime side triage. Other implicit invokeMethod behaviors (Content-Type default, request-id, error mapping, trace propagation) are intentionally left to the caller, since exposing the native HttpClient is the point of the migration.

Also document on DaprBodyPublishers.json that the helper uses the SDK's default object serializer and is not a wrapper around a configured custom DaprObjectSerializer. Its only contribution over a plain ofByteArray call is guaranteeing a length-known BodyPublisher, which keeps the JDK on Content-Length framing and sidesteps the Transfer-Encoding: chunked race reported under load.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/sdk/src/main/java/io/dapr/client/DaprBodyPublishers.java
+++ b/sdk/src/main/java/io/dapr/client/DaprBodyPublishers.java
@@ -59,6 +59,18 @@ public final class DaprBodyPublishers {
    * <p>Callers are still responsible for setting an appropriate
    * {@code Content-Type} header (typically {@code application/json}).
    *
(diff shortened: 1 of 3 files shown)


siri-varma (approved):
LGTM

PR #1766, opened 2026-06-10
use v1.18.0

cicoyle:
use v1.18.0


cicoyle:
this will fail bc the cli is not released yet

javier-aliaga (dismissed):

javier-aliaga:
I think you are missing test container config

javier-aliaga (approved):

PR #1772, opened 2026-06-19
chore: migrate to Spring Boot 4.0

javier-aliaga:
# Description

Spring boot 3.x is about to reach EOL, moving default spring boot version to 4.0 for now, later we will introduce 4.1

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


javier-aliaga:
@dapr/maintainers-java-sdk would be nice to get feedback. I still need to decide what to do with the current 1.18 

We deprecate SB 3.5 and we do updates on deps best effort and suggest people to move to SB 4. Shall we backport this to 1.18.x ?

spring-boot-examples/producer-app/src/test/java/io/dapr/springboot/examples/producer/ProducerAppIT.java
```diff
@@ -74,89 +84,124 @@ void cleanUp() {
...
-
-    given().contentType(ContentType.JSON)
-            .when()
-            .queryParam("amount", 2)
-            .get("/orders/byAmount/")
-            .then()
-            .statusCode(200).body("size()", is(0));
-
-  }
+//  @Test
```
cicoyle: Did this start to fail and that is why you commented it out? I think we should keep the test to ensure no regressions.

cicoyle (changes requested):

spring-boot-examples/producer-app/src/test/java/io/dapr/springboot/examples/producer/ProducerAppIT.java
javier-aliaga: I reverted, it should work fine

cicoyle (dismissed):
Once all copilots feedback is addressed this is good to merge. @javier-aliaga can you confirm its addressed?

siri-varma:
> We deprecate SB 3.5 and we do updates on deps best effort and suggest people to move to SB 4. Shall we backport this to 1.18.x ?

@javier-aliaga should we have a major version upgrade for this ? I believe moving from SB 3.5 to 4 is a big change

javier-aliaga:
@siri-varma do you mean jumping to 2.0.0? or just 1.19.0?

siri-varma:
Yes. Jumping to 2.0.0

siri-varma:
With Spring opensource support ending every year I am rethinking if a major version bump is a good idea

javier-aliaga:
@siri-varma I am not really sure, I see you point but also the main artifact here is the sdk, why we jump to 2.0 if the main component remains the same? What could make more sense is splitting the release or the versioning, so we stay sdk into 1.19 what we start releasing spring with its own versioning, if we do that we do not need to sync releases with dapr and spring

salaboy:
👀👀👀👀👀👀👀👀👀👀 
This is a major change (220 files already), consider the scenario when a Spring Boot 3.x user comes and requires compatibility with that version. Along with this change, the project should provide clear docs on how to deal with that. Failing to do so, you can run into major unexpected combinations of versions. You can be clear in supporting only the 3.5.x branch (https://spring.io/projects/spring-boot#support). I am worried about removing the base support for Spring 5.x as a single customer might push you to go back an all these changes.

javier-aliaga:
> 👀👀👀👀👀👀👀👀👀👀 This is a major change (220 files already), consider the scenario when a Spring Boot 3.x user comes and requires compatibility with that version. Along with this change, the project should provide clear docs on how to deal with that. Failing to do so, you can run into major unexpected combinations of versions. You can be clear in supporting only the 3.5.x branch (https://spring.io/projects/spring-boot#support). I am worried about removing the base support for Spring 5.x as a single customer might push you to go back an all these changes.

@salaboy I see your point, but I do not see an easy path going forward other than documenting it https://github.com/dapr/java-sdk/pull/1772/changes/f3bcbd3f0fe2b92811048c5ed2b9b1ab1869fc0b

If SB 3.5.x is EOL and the do not provide vulnerabilities fixes we can do as a best effort, but eventually it will pop one that we will not able to fix, and I am sure it will appear sooner than later. In case someone wants to stay in 3.5 they can stay in 1.18.x, we are continuing providing updates.

siri-varma:
Yep. Agree with this approach. They could still remain on 1.18 if 3.5 is needed and may be we could support with patches.

siri-varma (approved):
LGTM. Do we have to remove the 3.5 build ?

PR #1777, opened 2026-07-01
Add stateful-history delta work items to the workflow worker

JoshVanL:
The sidecar re-sends a workflow instance's entire committed history to the worker on every turn. This adds the worker half of the "stateful history" optimization so that, once a worker is warm for an instance on a work-item stream, the sidecar sends only the new committed events (the delta) and the worker reconstructs the full history from its own cache. It mirrors the Go (durabletask-go), Python, and .NET SDK implementations and is on by default.

Worker (durabletask-client):
- WorkflowHistoryCache: a per-stream cache of each instance's committed history, bounded by a sliding TTL, an instance-count cap, and a byte budget with LRU eviction. Injectable clock for deterministic tests.
- DurableTaskGrpcWorker: advertise WORKER_CAPABILITY_STATEFUL_HISTORY in GetWorkItemsRequest, reset the cache on every reconnect (the sidecar drops the old stream's warm set), and reclaim idle entries with a daemon janitor stopped on close.
- OrchestratorRunner: before replay, resolve the full committed history (cached prefix + delta on a hit, or a GetInstanceHistory fetch on a miss) instead of using the request's pastEvents directly; after replay, cache the committed history, or drop it once the instance ends (a CompleteWorkflow action, covering completed/failed/terminated/continued-as-new). A TerminateWorkflow action targets a different instance and is deliberately not treated as a reset.

Correctness never depends on the cache: any miss (cold stream, eviction, desync) self-heals via the GetInstanceHistory fallback, so this only changes per-turn bandwidth, not results. A fallback fetch that fails abandons the work item for backend redelivery rather than completing with a partial history.

Configuration (DurableTaskGrpcWorkerBuilder):
- disableStatefulHistory to opt out, plus historyCacheTtl, historyCacheMaxInstances, and historyCacheMaxBytes to tune the bounds.

### DIFF
--- a/durabletask-client/src/main/java/io/dapr/durabletask/WorkflowHistoryCache.java
+++ b/durabletask-client/src/main/java/io/dapr/durabletask/WorkflowHistoryCache.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright 2026 The Dapr Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package io.dapr.durabletask;
+
+import io.dapr.durabletask.implementation.protobuf.HistoryEvents;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.LongSupplier;
+
+/**
+ * Per-stream cache of each workflow instance's committed history, enabling "stateful history"
+ * delta work items: a worker that advertises {@code WORKER_CAPABILITY_STATEFUL_HISTORY} retains
+ * the history it has already replayed so the sidecar can send only the new events (the delta).
+ * Entries are reclaimed by a sliding TTL, an instance-count cap, and an optional byte budget
+ * (LRU eviction). Eviction is always safe: a miss is recovered via the GetInstanceHistory RPC.
+ *
+ * <p>Thread-safe; work items are processed concurrently on the worker pool.</p>
+ */
+public final class WorkflowHistoryCache {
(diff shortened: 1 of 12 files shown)


durabletask-client/src/main/java/io/dapr/durabletask/runner/OrchestratorRunner.java
```diff
@@ -88,4 +146,55 @@ public void run() {
...
+    }
+
+    boolean ended = result.getActions().stream()
+        .anyMatch(OrchestratorActions.WorkflowAction::hasCompleteWorkflow);
+    if (ended) {
+      this.historyCache.remove(instanceId);
+    } else {
+      this.historyCache.put(instanceId, pastEvents);
+    }
+  }
```
javier-aliaga, line 208: It seems we're not porting one check from the Go reference. We write to the cache
unconditionally, while Go first checks that the turn doing the writing is still the
current one (`worker_grpc.go:245`).

Without it, a runner that stalls across a reconnect can overwrite a newer entry.
Most of the time the length check catches it and we just refetch, but after a
continue-as-new the stale prefix can have the same length as the good one, and
then it goes through silently.

Possible fix: port `latestTokens` / `noteDispatch` / `isLatestDispatch` /
`forgetDispatch` from `worker_history.go:120-186`, call `noteDispatch` where the
work item is dispatched, and gate the `updateHistoryCache` body on it. The one
thing not to get wrong: the token map must not be cleared in `reset()`.

durabletask-client/src/main/java/io/dapr/durabletask/WorkflowHistoryCache.java
```diff
@@ -0,0 +1,224 @@
...
+    synchronized (this.lock) {
+      Entry existing = this.entries.get(instanceId);
+      if (existing != null) {
+        this.totalBytes -= existing.bytes;
+      }
+      this.entries.put(instanceId, new Entry(snapshot, bytes, this.clockNanos.getAsLong()));
+      this.totalBytes += bytes;
+      this.evictToFit(instanceId);
+    }
+  }
```
javier-aliaga, line 154: Should be check if bytes are 0 so we do not put it on the entries?

JoshVanL:
Thanks @javier-aliaga, please take another look!

javier-aliaga (approved):

marcduiker:
@holopin-bot @JoshVanL sdk-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #1783, opened 2026-07-07
feat: propagate caller trace context when scheduling workflows

javier-aliaga:
Upstream half of workflow trace propagation (follow-up to #1619).

DaprWorkflowClient gains a DaprWorkflowClient(Properties, Tracer) constructor that passes the tracer to the internal DurableTaskGrpcClientBuilder. DurableTaskGrpcClient now uses the configured tracer when scheduling: it emits a create_orchestration CLIENT span as a child of the caller's current OpenTelemetry context and stamps its W3C trace context into
CreateInstanceRequest.parentTraceContext, so the workflow execution nests under the caller's trace instead of starting a separate one.

Tracing stays opt-in: without a tracer the request is unchanged and no spans are emitted. The unused GlobalOpenTelemetry fallback is removed so the global instance is never pulled implicitly.

# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/durabletask-client/pom.xml
+++ b/durabletask-client/pom.xml
@@ -90,6 +90,11 @@
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-context</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcClient.java
+++ b/durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcClient.java
@@ -35,6 +35,7 @@ import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
 import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
 
 import javax.annotation.Nullable;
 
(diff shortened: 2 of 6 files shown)


sdk-workflows/src/main/java/io/dapr/workflows/client/DaprWorkflowClient.java
```diff
@@ -67,16 +84,17 @@ public DaprWorkflowClient(Properties properties) {
    * @param additionalInterceptors extra interceptors appended after the API-token interceptor.
    */
   protected DaprWorkflowClient(Properties properties, ClientInterceptor... additionalInterceptors) {
-    this(buildChannelWithAdditional(properties, additionalInterceptors));
+    this(buildChannelWithAdditional(properties, additionalInterceptors), null);
```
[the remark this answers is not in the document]
javier-aliaga: No compilation error, false finding

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcClient.java
```diff
@@ -198,14 +204,68 @@ public String scheduleNewOrchestrationInstance(
       builder.setScheduledStartTimestamp(ts);
     }
 
+    Span span = null;
+    if (this.tracer != null) {
+      span = this.tracer.spanBuilder("create_orchestration:" + orchestratorName)
+          .setSpanKind(SpanKind.CLIENT)
+          .setAttribute("durabletask.type", "orchestration")
```
salaboy, line 211: @javier-aliaga this is confusing already.. why DurableTask here? can you not make it "workflow."

durabletask-client/pom.xml
```diff
@@ -90,6 +90,11 @@
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-context</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk</artifactId>
```
salaboy, line 95: @javier-aliaga I think you can use only opentelemetry-api, so you don't bring the entire SDK dependency

sdk-workflows/pom.xml
```diff
@@ -42,6 +42,10 @@
       <artifactId>durabletask-client</artifactId>
       <version>${project.parent.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
```
salaboy, line 47: @javier-aliaga this one better.

durabletask-client/pom.xml
javier-aliaga: This is a test only dep, we need the full sdk for testing

durabletask-client/src/main/java/io/dapr/durabletask/DurableTaskGrpcClient.java
javier-aliaga: we still have durabletask.* all around the place

salaboy (dismissed):
/lgmt besides the durabletask.* properties that I would love to get rid of.. this is a good step forward.

cicoyle (approved):

PR #1784, opened 2026-07-09
Fix durable task flaky integration test

siri-varma:
# Description

Fixes flaky Durable Task integration tests and adds a CI retry safety net so transient infrastructure hiccups don't fail otherwise-green runs.

## Test fixes (`DurableTaskClientIT`)

**`longTimeStampTimer`** — The timer's fire time was computed from the test's wall clock at method start, and the test then asserted an exact internal sub-timer count (`counter >= 4`). The number of sub-timers a long timer is split into depends on the span from the orchestration's *creation* time to the deadline; worker startup + scheduling latency shrank that span under CI load, occasionally dropping a sub-timer and failing `expected <true> but was <false>`. Fix: anchor the timer to the orchestration's own replay-safe clock via `ctx.getCurrentInstant()` (what `createTimer(Duration)` does internally) so the span is always exactly `delay`, and assert the *splitting invariant* (`counter >= 3`) instead of a latency-sensitive exact count.

**`waitForInstanceStartThrowsException`** — Scheduling the instance on a background thread raced `waitForInstanceStart`: if the wait reached the sidecar first it failed with "no such instance exists"; if the schedule won, the worker started the instance within 2s and nothing was thrown. Fix: schedule with a future start time so the instance deterministically exists but stays `Pending`, guaranteeing `waitForInstanceStart` blocks to its 2s deadline and throws `TimeoutException`.

## CI retries (`build.yml`, `validate.yml`)

Wrap the flaky integration/validation commands in `nick-fields/retry@v3` (`max_attempts: 2`, `timeout_minutes: 30`) at job granularity:

- **Durable Task** — retry the `mvnw ... verify` step; `continue_on_error` keeps the sidecar-log upload and the outcome-gated fail step intact.
- **Spring Boot 3.5 / 4.0 build legs** — retry the integration-test command; `continue_on_error` is left off so an exhausted retry still fails the step, preserving the existing `failure()` artifact uploads and job-failure behavior.
- **Auto Validate** — a single retry around all example validations (reruns the batch once on failure) rather than per-example.

Retries are a safety net for transient failures (sidecar/gRPC connect races, example runs), not a substitute for the root-cause test fixes above.

## Issue reference

_[add issue number if this closes one]_

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -206,11 +206,21 @@ jobs:
     - name: Integration tests using spring boot 3.x version ${{ matrix.spring-boot-version }}
       id: integration_tests
       if: ${{ matrix.spring-boot-display-version == '3.5.x' }}
-      run: PRODUCT_SPRING_BOOT_VERSION=${{ matrix.spring-boot-version }} ./mvnw -B -pl !durabletask-client -Pintegration-tests dependency:copy-dependencies verify
+      uses: nick-fields/retry@v3
+      with:
+        max_attempts: 2
+        timeout_minutes: 30
+        retry_wait_seconds: 15
+        command: PRODUCT_SPRING_BOOT_VERSION=${{ matrix.spring-boot-version }} ./mvnw -B -pl !durabletask-client -Pintegration-tests dependency:copy-dependencies verify
     - name: Integration tests using spring boot 4.x version ${{ matrix.spring-boot-version }}
       id: integration_sb4_tests
       if: ${{ matrix.spring-boot-display-version == '4.0.x' }}
-      run: PRODUCT_SPRING_BOOT_VERSION=${{ matrix.spring-boot-version }} ./mvnw -B -pl !durabletask-client -Pintegration-sb4-tests dependency:copy-dependencies verify
+      uses: nick-fields/retry@v3
+      with:
+        max_attempts: 2
+        timeout_minutes: 30
+        retry_wait_seconds: 15
+        command: PRODUCT_SPRING_BOOT_VERSION=${{ matrix.spring-boot-version }} ./mvnw -B -pl !durabletask-client -Pintegration-sb4-tests dependency:copy-dependencies verify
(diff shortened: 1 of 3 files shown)


siri-varma:
@javier-aliaga durable task has a flaky test. Fixed it as part of this PR

siri-varma:
nvm. It failed again

siri-varma:
@javier-aliaga need one review on this pr please

.github/workflows/build.yml
```diff
@@ -200,11 +201,21 @@ jobs:
     - name: Integration tests using spring boot 3.x version ${{ matrix.spring-boot-version }}
       id: integration_tests
       if: ${{ matrix.spring-boot-display-version == '3.5.x' }}
-      run: PRODUCT_SPRING_BOOT_VERSION=${{ matrix.spring-boot-version }} ./mvnw -B -pl !durabletask-client -Pintegration-tests dependency:copy-dependencies verify
+      uses: nick-fields/retry@v3
+      with:
+        max_attempts: 2
+        timeout_minutes: 30
```
javier-aliaga, line 207: Do we need to increase the job timeout to 1hour to fit the retry time?

.github/workflows/build.yml
javier-aliaga, line 101: increase?

javier-aliaga, line 52: increase?

siri-varma: I thought about but currently all our tests finish  within 30 mins. So, I think we should be fine. I actually think we should bring it further down because I did around 20 runs and all of those runs completed within 30 mins. 

There are couple of optimizations I am pushing in another which should bring the timing further down. In this way if any subsequent PR increases the duration we will know automatically.

javier-aliaga (approved):

.github/workflows/build.yml
javier-aliaga: This timeout is for the complete setup.

PR #1785, opened 2026-07-10
Move most of the remaining tests to Test Containers as well

### LINKED ISSUE #1522: Optimize CI/CD build times - Integration tests timing
The CI/CD pipeline is experiencing timeouts during integration tests, specifically for Spring Boot 3.3.9. The build job `Build jdk:17 sb:3.3.x exp:false` is being cancelled after exceeding the 30-minute execution limit, as seen in [this failed run](https://github.com/dapr/java-sdk/actions/runs/17164312802/job/48848725515).

## Current State
- Integration tests are taking longer than 30 minutes to complete
- The workflow includes resource-intensive operations, for ex:
  - Docker container setup (MongoDB, Kafka)
  - Dapr CLI and runtime initialization
  - Complex workflow testing with external dependencies
- Tests run sequentially without parallelization

### Parallel Test Execution
- Enable Maven Surefire parallel execution for unit tests
- Configure Maven Failsafe for parallel integration tests
- Add parallel module building with `-T` flags

### Split Workflow Strategy
- Separate unit tests from integration tests into different jobs
- Run unit tests first and in parallel
- Only run integration tests (in parallel) after unit tests pass
- Use focused timeouts for each job type

### Test Optimization
- Review and optimize slow integration tests

## Expected Benefits
- Reduce overall build time by ~30-50%
- Eliminate timeout issues
- Improve developer experience with faster feedback at PR time
- Reduce CI/CD resource usage

## Acceptance Criteria
- [ ] Build completes within reasonable time (< 20 minutes)
- [ ] All existing tests continue to pass
- [ ] No regression in test coverage
- [ ] Maintainable and clear CI/CD configuration

siri-varma:
# Description

Finishes migrating the `sdk-tests` integration tests off the legacy `dapr run` CLI harness and onto **Testcontainers** (`BaseContainerIT` / `DaprContainer`), and strips the now-unused legacy setup from the `build` CI job. This removes the slow, flaky host-`daprd` path that was driving the `build sb:3.5.x` job toward its timeout.

## What changed

**Test migration (last legacy ITs → Testcontainers)**
- Migrated the remaining CLI-based ITs: `ActorReminderRecoveryIT`, `ActorReminderFailoverIT`, `WaitForSidecarIT`, `BindingIT` (now Testcontainers Kafka + in-code bindings), and `ActorSdkResiliencyIT` (moved off `ToxiProxyRun`; still `@Disabled`).
- Added reusable mechanics to `BaseContainerIT`: `startAppAndAttach` (app subprocess + pinned-port `daprd`), `restartApp` / `restartSidecar`, `waitForActorsReady`, and `newToxiproxy` (Testcontainers ToxiProxy for network-fault injection — replaces the host-installed binary; same toxics, now hermetic and version-pinned).
- Centralized shared backing services in `SharedTestInfra`: one Docker `Network` + JVM-singleton Redis / Zipkin / Mongo / Kafka.

**Actor IT flakiness fix — shared control plane (DNS-alias collision)**
- Root cause: every actor IT attaches `daprd` to the JVM-wide shared `Network`. The builder used `withReusablePlacement(true)`; with Testcontainers reuse disabled (as on CI), `DaprContainer.configure()` auto-creates a placement **and** scheduler per `daprd` instance — each claiming the `placement` / `scheduler` network alias and never stopped (no `stop()` override). Multiple containers answering one alias make Docker DNS round-robin `daprd` to an arbitrary/empty control plane, so multi-sidecar failover (`ActorReminderFailoverIT`) and post-restart reminder recovery (`ActorReminderRecoveryIT`) hit a placement that never saw the actor-host registrations → `did not find address for actor` / reminders not resumed.
- Fix: JVM-singleton `placement()` / `scheduler()` in `SharedTestInfra` on the shared network, wired into every `daprd` via `withPlacementContainer` / `withSchedulerContainer`. Exactly one container answers each alias, so DNS resolves deterministically; this also removes the per-`daprd` control-plane container leak.

**CI cleanup**
- Dropped Dapr CLI + `dapr init`, host Kafka, and host ToxiProxy setup from the `build` job.
- Removed the dead `dapr run` harness, binding YAMLs, and `local-test.yml`.

## Issue reference

Closes #1522

## Test results

- **Actor package: green** — `ActorReminderRecoveryIT` (4/4) and `ActorReminderFailoverIT` (1/1), which failed deterministically before the fix, now pass both in isolation and in the full suite.
- **`tracing.http.TracingIT`: fixed.** Its Spring `@Bean` app eagerly called the legacy `OpenTelemetry.createOpenTelemetry(SERVICE_NAME)` overload, which probes a fixed `127.0.0.1:9411` for Zipkin readiness. After this migration Zipkin runs as a Testcontainer on a random mapped host port, so the probe threw `ConnectException`, the app subprocess crashed before binding its port, and `startAppAndAttach` reported "connection refused". Switched the bean to the explicit-endpoint overload that skips the probe (the app exports no spans the test asserts on — the validated `calllocal/tracing-http-it/sleep` span is emitted by the sidecar over the container network). Passes in isolation.
- **Full `sdk-tests` suite (local): 7 → 1 error.** All four actor errors resolved by the control-plane fix; `TracingIT.http` resolved by the bean fix above.
- **1 remaining local failure** (`spring.messaging.DaprSpringMessagingIT`) is a topic-subscription readiness timeout. It reproduces even in isolation and is **independent of this PR** — it is a standalone `@SpringBootTest` that manages its own container and touches none of the code changed here. It appears environment-sensitive on a local macOS Docker host (uses `host.testcontainers.internal`); CI on this PR is the arbiter for whether it affects the target environment. Tracked as follow-up, not introduced here.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


pom.xml
```diff
@@ -689,6 +689,10 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>${maven-javadoc-plugin.version}</version>
+        <!-- Aggregate is a root-only report. Without inherited=false the site
+             lifecycle re-runs it at every reactor module, multiplying a full-source
+             Javadoc pass ~10x. Keep it pinned to the aggregator module. -->
+        <inherited>false</inherited>
```
siri-varma, line 695: @javier-aliaga / @cicoyle Please double check this change.

We are able to save 30 mins because of this change. It was taking 32 mins before now just takes 4 mins

siri-varma:
@javier-aliaga need one review on this pr please

javier-aliaga:
The PR looks good, the only thing is the README still mentions we need to run `docker compose -f ./sdk-tests/deploy/local-test.yml up -d` and I do not think this is true anymore

siri-varma:
@javier-aliaga updated the README.

javier-aliaga (approved):

PR #1543, opened 2025-09-07
Remove shaded deps

mcruzdev:
# Description

I am not sure if this one is the best way to solve some issues related to native compilation. For Spring Native and Quarkus the `grpc-netty-shaded` is not recommended due to recents CVEs and a lot of GraalVM configurations and substitutions.

> We strongly recommend NOT using a shaded Netty - the recent cves should tell you why

> Not to mention that it increase drammatically the memory usage too: the most of the assumption from Netty related pooling is to be the only one (like last Highlander...)
if you shade it...it end up pretty bad.

I am not an expertise in the netty subject, I appreciate any suggestions and comments about this change.

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Extended the documentation

PR #1793, opened 2026-07-20
fix: pin Netty to the gRPC-supported version in Spring modules

javier-aliaga:
## Description

The Spring Boot 4 migration (#1772) has the Dapr Spring modules import the Spring Boot 4 BOM, which manages **Netty 4.2.x**. The SDK uses the non-shaded `grpc-netty` (1.79.0), and per gRPC's [compatibility matrix](https://github.com/grpc/grpc-java/blob/master/SECURITY.md) gRPC 1.79 targets **Netty 4.1.x** — no gRPC release supports Netty 4.2.x. Inside these modules `grpc-netty` is a `compile`-scope transitive (via `dapr-sdk`), so it was resolving onto an unsupported Netty line.

The concrete impact is on the SDK's **own** reactor: it was building and — more importantly — running its Spring integration tests and examples against a Netty version gRPC does not support (`sdk-tests` / `spring-boot-sdk-tests` exercise real gRPC over a Testcontainers `daprd`). In other words, "CI green" was validating gRPC on Netty 4.2.x, an untested combination that can fail with `NoSuchMethodError` / `NoClassDefFoundError`.

**Scope note:** downstream apps that import `dapr-spring-bom` are already protected — it pins Netty `4.1.132` transitively via `dapr-sdk-bom`. This change does not alter that; it aligns the SDK's own builds, tests, and examples with the same supported Netty version the core modules already use. The root `pom.xml` pins `netty-bom` `4.1.132`, but the Spring modules override it by importing the Spring Boot BOM in their own `dependencyManagement`, so the pin never reached them.

Switching to `grpc-netty-shaded` is not an option — it was deliberately removed in #1543 because shaded Netty breaks native compilation (Quarkus / Spring Native), prevents patching Netty CVEs, and increases memory usage.

## Fix

Import `netty-bom` (`${netty.version}` = `4.1.132`) **before** `spring-boot-dependencies` in the gRPC-reachable Spring modules, so Maven's first-declared-wins rule keeps Netty on the gRPC-supported 4.1.x line:

- `dapr-spring-boot-autoconfigure`
- `dapr-spring-data`
- `dapr-spring-boot-starter`
- `sdk-tests`
- `spring-boot-examples`
- `spring-boot-sdk-tests`

`dapr-spring-boot-starter-test` pulls no Netty transitively, so it is intentionally left unpinned.

## Verification

Resolved `io.netty:netty-common` via `dependency:tree` in each affected module — all now resolve **`4.1.132.Final`** (previously `4.2.12.Final`), matching the version already used by the core SDK.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Extended the documentation (inline comment explaining the pin)

### DIFF
--- a/sdk-tests/pom.xml
+++ b/sdk-tests/pom.xml
@@ -35,6 +35,15 @@
     <dependencies>
       <!-- Import the Spring Boot 4 BOM so JUnit (jupiter + platform) resolves consistently to 6.x,
            matching the spring-boot-sdk-tests module and the Spring Boot 4 starter pulled in here. -->
+      <!-- Pin Netty to the gRPC-supported 4.1.x line (root netty.version); MUST precede the Spring Boot
+           BOM, which manages Netty 4.2.x - unsupported by the non-shaded grpc-netty this SDK uses. -->
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-bom</artifactId>
+        <version>${netty.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
       <dependency>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-dependencies</artifactId>
--- a/dapr-spring/dapr-spring-boot-autoconfigure/pom.xml
+++ b/dapr-spring/dapr-spring-boot-autoconfigure/pom.xml
@@ -31,6 +31,15 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <!-- Pin Netty to the gRPC-supported 4.1.x line (root netty.version); MUST precede the Spring Boot
+           BOM, which manages Netty 4.2.x - unsupported by the non-shaded grpc-netty this SDK uses. -->
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-bom</artifactId>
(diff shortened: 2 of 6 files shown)


sdk-tests/pom.xml
```diff
@@ -35,6 +35,15 @@
     <dependencies>
       <!-- Import the Spring Boot 4 BOM so JUnit (jupiter + platform) resolves consistently to 6.x,
            matching the spring-boot-sdk-tests module and the Spring Boot 4 starter pulled in here. -->
+      <!-- Pin Netty to the gRPC-supported 4.1.x line (root netty.version); MUST precede the Spring Boot
+           BOM, which manages Netty 4.2.x - unsupported by the non-shaded grpc-netty this SDK uses. -->
+      <dependency>
```
siri-varma, line 40: @javier-aliaga I think we should do this in the parent pom. And just import it here. That way all the versions are in one single place.

Applies to changes in all other poms as well

javier-aliaga: I tried moving it to parent but then it resolves to 4.2.x instead of 4.1.x, I did a research and it seems as we import the spring-boot-dependencies it wins over the parent import, that is why we need it in every pom. To make it to the parent it would require a bigger refactor...

siri-varma (approved):

